PR #11, opened 2022-05-24
feat: mdns & dht discovery

sethvincent:
This pr implements a wrapper module around [mdns-sd-discovery](https://github.com/digidem/mdns-sd-discovery) and [hyperswarm](https://github.com/hyperswarm/hyperswarm) to provide peer discovery through mdns and dht with one api.


lib/discovery.js
```diff
@@ -0,0 +1,198 @@
...
+        });
+
+        if (this.dhtOptions.server) {
+          await discover.dht.flushed();
+        } else {
+          await this.#dht.flush();
+        }
+      }
+
+      this.#discover.set(topicString, discover);
```
gmaclennan, line 124: I think this needs to be above the two `await`s above, to avoid errors when calling join() on the same topic more than once in quick succession.

lib/discovery.js
```diff
@@ -0,0 +1,198 @@
...
+      });
+
+      this.#peers.set(publicKey, {
+        connection,
+        peerInfo,
+      });
+
+      this.emit("peer", connection, peerInfo);
+    } else {
+      this.#dht.leavePeer(publicKey);
```
gmaclennan, line 76: Do I understand this correctly: Is this to de-dupe connections made on both mdns and dht? Prioritizing mdns (if it is first)? Maybe we could do an additional check on the IP range so that we could also prioritize mdns if it is not first?

lib/localpeers.js
```diff
@@ -0,0 +1,315 @@
...
+}
+
+/**
+ * @typdef {object}
+ * @property {string} host
+ * @property {number} number
+ * @property {string} topic
+ * @property {Buffer} identityPublicKey
+ */
+class MdnsPeerInfo {
```
gmaclennan, line 299: Why a class rather than a plain object? (a question, not a criticism)

lib/discovery.js
```diff
@@ -0,0 +1,198 @@
...
+      });
+
+      this.#peers.set(info.identityPublicKey, {
+        connection,
+        peerInfo,
+      });
+
+      this.emit("peer", connection, peerInfo);
+    } else {
+      peer.addTopic(info.topic);
```
gmaclennan, line 96: Will this work ok across mixed dht & mdns connections to peers? Wondering what happens if a peer connects first via dht, then adds a topic which is picked up on mdns first?

gmaclennan (commented):
Great work on this. I'd love to hop on a call so you can talk me through some of this - there are parts that I don't quite understand. My questions (likely due to not fully understanding the code):
- Are we broadcasting a new mdns service for each topic? All on the same port? And a single connection for all topics? I wasn't clear what is happening with the peer disconnects and leaving some topics (but not all topics).
- Where are we deduping mdns connections (e.g. there will be an incoming and outgoing connection - Hyperswarm seems to do that here: https://github.com/hyperswarm/hyperswarm/blob/master/index.js#L224-L235
- Race conditions in async methods like join and leave can be common (based on prior experience with the app) due to users being able to switch rapidly between screens and in and out of the app (which causes join and leave actions). I haven't thoroughly reviewed the tests, but I want to make sure we correctly handle the "joining", "leaving", "destroying" states etc. if they do exist - think about the whole thing in terms of a state machine and what possible states exist and what might happen when methods are called in some of these "intermediary" states that happen during an async function.

lib/discovery.js
```diff
@@ -0,0 +1,198 @@
...
+ */
+export class Discovery extends TypedEmitter {
+  #peers = new Map();
+  #discover = new Map();
+  #dht;
+  #mdns;
+
+  /**
+   * @param {object} options
+   * @param {number} options.port
```
gmaclennan, line 23: This is setting the port for the mdns discovery, but I hyperswarm is choosing its own port ([49737 by default](https://github.com/hyperswarm/dht/blob/master/index.js#L22)). I wonder if we just let the localpeers class handle the port? Do we risk just choosing a very random port (like hyperswarm) and assuming no other service will be on it? One thing to bear in mind is different instances of Mapeo running on the same device (QA build and production) which can cause issues with fixed port numbers.

tests/discovery.js
```diff
@@ -0,0 +1,72 @@
...
+      await discover1.destroy();
+      await discover2.destroy();
+      await testnet.destroy();
+    }
+  }
+
+  // TODO: use discoveryKey
+  await discover1.join(keyPair.publicKey);
+  await discover2.join(keyPair.publicKey);
+});
```
gmaclennan, line 72: I'm not sure the step() and cleanup is doing what you think it is here (probably why tape tests were causing issues). The test function is going to resolve before the cleanup, and the `await` in the "peer" handlers seems unnecessary.

I think you need something like:

```js
function deferredPromise () {
  let resolve, reject
  const promise = new Promise((res, rej) => {
    resolve = res
    reject = rej
  })
  return { resolve, reject, promise }
}

const deferred = deferredPromise()
```

Then, at the end of the step() function, call `deferred.resolve()` and at the last line of the test call `await deferred.promise`

gmaclennan:
It would be good to think about the boundaries in mapeo-core between the different modules - thinking as if we would publish as a separate module. My understanding is that the "public" interface (i.e. the interface used by the rest of mapeo core) for discovery is the `./lib/discovery` module, and the `./lib/localpeers` is an implementation detail. If we clearly define that then we can clearly define the internal API the other pieces will use - the `Discovery` class - and it would be good to have that API thoroughly tested, so that we have the flexibility to change the implementation details of that module and have good separation between components of Mapeo Core. I think you've written it like this, it's just not currently that clear from how the code is organized and presented. Also good to have function methods and params documented since it really helps with future maintenance (and for code reviews).

lib/discovery.js
sethvincent: I realized we can reasonably let the net `server.listen` handle picking an available port.

lib/discovery.js
```diff
@@ -0,0 +1,759 @@
...
+ * @property {(error: Error) => void} error
+ */
+
+/**
+ * @extends {TypedEmitter<DiscoveryEvents>}
+ */
+export class Discovery extends TypedEmitter {
+  #identityKeyPair
+  #topics = new Map()
+  #peers = new Map()
```
gmaclennan, line 22: `/** @type {Map<string, PeerInfo>} */`

lib/discovery.js
```diff
@@ -0,0 +1,759 @@
...
+      this.mdns = mdns
+    }
+
+    if (dht) {
+      this.dhtOptions = /** @type {DhtOptions} */ (
+        dht && typeof dht === 'object' ? dht : {}
+      )
+
+      this.dhtOptions.keyPair = identityKeyPair
+      this.dht = new Hyperswarm(this.dhtOptions)
```
gmaclennan, line 49: Stricter typing of this.dht will help catch issues with this.dht being undefined - I think there is a risk of this in later code.

lib/discovery.js
```diff
@@ -0,0 +1,759 @@
...
+
+  /**
+   * @param {import('net').Socket} socket
+   */
+  _onMdnsConnection(socket) {
+    const socketAddress = /** @type {import('net').AddressInfo} */ (
+      socket.address()
+    )
+
+    const connection = new SecretStream(false, socket)
```
gmaclennan, line 87: Is this of type ConnectionStream?

lib/types.js
```diff
@@ -0,0 +1,70 @@
+/**
```
gmaclennan, line 1: If you prefer, it's possible to write these types-only files as typescript - it's a bit more concise, but also perfectly fine to leave as jsdoc.

lib/discovery.js
```diff
@@ -0,0 +1,759 @@
...
+        // TODO: it seems like this ECONNRESET error is expected when a stream closes, but that also doesn't seem right
+        return
+      }
+
+      console.error(error)
+    })
+
+    this.emit('connection', connection, peer)
+    for (const topic of this.#topics.values()) {
+      await this.dht.flush()
```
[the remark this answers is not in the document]
sethvincent: 👍

lib/discovery.js
sethvincent: :+1:

lib/discovery.js
```diff
@@ -0,0 +1,759 @@
...
+    }
+  }
+
+  /**
+   * @param {import('net').Socket} socket
+   */
+  _onMdnsConnection(socket) {
+    const socketAddress = /** @type {import('net').AddressInfo} */ (
+      socket.address()
+    )
```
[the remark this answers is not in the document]
sethvincent: This is happening after the connection completes so it looks like the socket.connecting line isn't needed, but added the other things.

sethvincent:
@gmaclennan I think this is ready now unless there's anything else we should get in to this pr.

I made a few issues that seem best handled in follow-up prs:
- #13 
- #12
- #14

lib/discovery.js
sethvincent: made an issue for this https://github.com/digidem/mapeo-core-next/issues/14

Issue #20, opened 2022-08-24
add examples & descriptions to jsdoc in discovery.js


sethvincent:
to produce better api docs we'll need better comments in lib/discovery.js.

currently the jsdoc needs:

- descriptions of classes and methods
- descriptions of parameters
- make sure all return values are documented
- `@example` tags for classes and methods https://tsdoc.org/pages/tags/example/

sethvincent:
closed by #30

PR #30, opened 2022-09-19
chore: add additional jsdoc annotations for discovery module

achou11:
Addresses #20

So far:
- descriptions for methods of classes

What's left:
- descriptions about classes
- example annotations


achou11:
@sethvincent working through adding examples and a couple of questions come up:

- Is it safe to assume that for methods, we only add an example for the public ones?
- How detailed should the examples be? Trying to balance between "detailed enough to basically copy-paste and have full-context" and "reduce noise caused by boilerplate and highlight the important bits". Any thoughts on what approach we should go with, or will it depend on what's being documented?

lib/discovery/index.js
```diff
@@ -57,27 +72,109 @@ export class Discovery extends TypedEmitter {
...
   /**
+   * Returns the list of peers that have been connected to
+   *
    * @returns {PeerInfo[]}
+   *
+   * @example
+   *
+   * ```js
+   *
+   * // Create identities...
```
sethvincent, line 136: It would be useful to list relevant methods for creating identities here and maybe link to their spot in the docs (not sure off the top of my head if that's possible from within example code using typedoc, so maybe this note would be outside the code block).

sethvincent:
Yeah only public methods.

I think we should lean toward minimal examples as much as possible. The examples you've added so far look pretty good to me. I just added one comment about referencing/linking to related methods that would add too much boilerplate in a method's example.

lib/discovery/index.js
achou11: Yeah makes sense! Another idea: not sure how it would look in the generated output but could put a (collapsible?) snippet for an example of creating the identity, as a footnote type of thing that can be referenced from here. Maybe too complicated so probably stick with your idea

achou11:
still missing class descriptions but not entirely sure what to put there just yet. otherwise think this is ready for review

lib/discovery/index.js
```diff
@@ -57,28 +72,138 @@ export class Discovery extends TypedEmitter {
...
+   * const discover = new Discover({
+   *  identityKeyPair: new KeyManager(
+   *    KeyManager.generateIdentityKey()
+   *  ).getIdentityKeypair(),
+   * })
+   *
+   * await discover.ready()
+   *
+   * // Create topic name (as a buffer) to be joined (see footnote 1)
+   * const topic = Buffer.from('topics-example')
```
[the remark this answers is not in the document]
achou11: alright addressed via https://github.com/digidem/mapeo-core-next/pull/30/commits/d4bf4428dd1d1673a4bf967d28e4e4e13eadef23

lib/discovery/index.js
```diff
@@ -57,28 +80,138 @@ export class Discovery extends TypedEmitter {
...
+   * const topic = Buffer.alloc(32).fill('topics-example')
+   *
+   * discover.join(topic)
+   *
+   * console.log(discover.topics) // prints out `['746f706963732d6578616d706c65746f706963732d6578616d706c65746f7069']`
+   * ```
+   *
+   * Footnotes:
+   *
+   * 1. To create a secure topic name, refer to `KeyManager`, `createIdentityKeys`, and `getHypercoreKeypair` in the [`@mapeo/crypto`](https://github.codigidem/mapeo-crypto) docs.
```
sethvincent, line 134:
```suggestion
   * 1. To create a secure topic name, refer to `KeyManager`, `createIdentityKeys`, and `getHypercoreKeypair` in the [`@mapeo/crypto`](https://github.com/digidem/mapeo-crypto) docs.
```

achou11: lol whoops, thanks!

Issue #14, opened 2022-07-14
add types for dependencies


sethvincent:
In particular:

- [ ] `brittle`
- [x] `multi-core-indexer`
- [x] `@mapeo/sqlite-indexer`
- [x] `@mapeo/crypto`
- [ ] `sodium-universal`
- [ ] `base32.js`

and anything else in https://github.com/digidem/mapeo-core-next/blob/main/types/modules.d.ts

achou11:
a good reference in case i start losing steam...

https://github.com/digidem/multi-core-indexer/tree/main/types

achou11:
note to self, `base32.js` already has ts defs but they're not published with the module right now (currently available on master branch)

sethvincent:
updated the list to reflect newly added types 🎉

achou11:
relevant thing that Gregor started work on: https://github.com/digidem/digidem-types

gmaclennan:
done, enough for now

PR #34, opened 2022-10-04
chore: add types for @hyperswarm/secret-stream

achou11:
Towards #14 

wow i thought this would be easier...

Notes:
- A few todo comments but will point them out and see if they're blocking (most aren't)

### DIFF
--- a/lib/discovery/index.js
+++ b/lib/discovery/index.js
@@ -8,7 +8,7 @@ import base32 from 'base32.js'
 
 /**
  * @typedef {Object} DiscoveryEvents
- * @property {(connection: NoiseSecretStream, info: PeerInfo) => void} connection
+ * @property {(connection: SecretStream<RawConnectionStream>, info: PeerInfo) => void} connection
  * @property {(topicStatus: { topic: TopicString, mdns: TopicServiceStatus, dht: TopicServiceStatus }) => void} topicStatus
  * @property {(peerStatus: { status: String, peer: PeerInfo }) => void} peerStatus
  * @property {(error: Error) => void} error
@@ -247,7 +247,6 @@ export class Discovery extends TypedEmitter {
       throw new Error('Socket not connected')
     }
 
-    /** @type {ConnectionStream} */
     const connection = new SecretStream(false, socket)
 
     connection.on('error', (error) => {
@@ -298,7 +297,7 @@ export class Discovery extends TypedEmitter {
    * Handle connection that occur via the Hyperswarm DHT
    *
    * @private
-   * @param {ConnectionStream} connection
+   * @param {SecretStream<RawDHTConnectionStream>} connection
    * @param {HyperswarmPeerInfo} info
    * @returns {Promise<void>}
    */
@@ -594,7 +593,7 @@ export class Discovery extends TypedEmitter {
    *
    * @private
    * @param {Address} options
-   * @returns {ConnectionStream}
+   * @returns {SecretStream<net.Socket>}
    */
   _connect(options) {
     const { host, port } = options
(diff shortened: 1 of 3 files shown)


lib/types.js
```diff
@@ -52,24 +52,13 @@
...
 
+// TODO: Figure out where those extra fields come from and find more elegant way to represent this
 /**
- * @typedef {Object} NoiseSecretStream
- * @property {Buffer} remotePublicKey
- * @property {number} remotePort
- * @property {string} remoteHost
- * @property {import('net').Socket} rawStream
- * @property {function} end
+ * @typedef {import('streamx').Duplex & {remoteAddress: string, remotePort:number}} RawDHTConnectionStream
```
achou11, line 57: Seth may have an idea of where these come from, but my guess is that it would require even more work to get the proper typings for that to update this

types/modules.d.ts
```diff
@@ -1,6 +1,64 @@
...
 declare module 'hyperswarm'
-declare module '@hyperswarm/secret-stream'
+declare module '@hyperswarm/secret-stream' {
+  import { Duplex as NodeDuplex } from 'stream'
+  import { Duplex, DuplexEvents } from 'streamx'
+
+  interface Opts {
+    autostart?: boolean
+    // TODO: Use https://github.com/chm-diederichs/noise-handshake/blob/main/noise.js for specific patterns
+    pattern?: string
```
achou11, line 10: tbh i had no idea how to read those pattern variables in the link provided, so went with string for now

types/modules.d.ts
```diff
@@ -1,6 +1,64 @@
...
+    RawStream extends NodeDuplex | Duplex = Duplex
+  > extends Duplex<
+    any,
+    any,
+    any,
+    any,
+    true,
+    true,
+    DuplexEvents<any, any> & NoiseStreamEvents
+  > {
```
achou11, line 36: is there a more concise way to write this??

lib/discovery/index.js
```diff
@@ -889,7 +888,7 @@ export class PeerInfo extends TypedEmitter {
    * @param {Object} options
    * @param {IdentityPublicKeyString} [options.identityPublicKey]
    * @param {('dht'|'mdns')} [options.discoveryType]
-   * @param {ConnectionStream} [options.connection]
+   * @param {SecretStream<RawConnectionStream>} [options.connection]
```
achou11, line 891: ideally would do some conditional typing to narrow down the RawConnectionStream type based on the `discoveryType`. not sure how to go about that in TS + JSDoc

lib/discovery/index.js
```diff
@@ -832,7 +831,7 @@ export class PeerInfo extends TypedEmitter {
    * @typedef {Object} PeerInfoOptions
    * @property {IdentityPublicKeyString} identityPublicKey
    * @property {('dht'|'mdns')} discoveryType
-   * @property {ConnectionStream} connection
+   * @property {SecretStream<RawConnectionStream>} connection
```
achou11, line 834: ideally would do some conditional typing to narrow down the RawConnectionStream type based on the `discoveryType`. not sure how to go about that in TS + JSDoc

lib/types.js
sethvincent: i think the approach used here makes sense.

types/modules.d.ts
sethvincent: yeah i think we're unlikely to set the pattern option too, so seems lower priority.

lib/discovery/index.js
sethvincent: ah, yeah that would be nice. i think we're good for now with `SecretStream<RawConnectionStream>` in part because ideally at this stage the streams from the two discovery types should be pretty much equivalent.

sethvincent (approved):
this looks good to me! i think for any remaining todos here we could make some low-priority issues for improvements

PR #35, opened 2022-10-05
chore: add types for random-access-memory

achou11:
Towards #14 

Notes:

- This also adds the definitions for [`random-access-storage`](https://github.com/random-access-storage/random-access-storage/) since `random-access-memory` extends it

### DIFF
--- a/types/modules.d.ts
+++ b/types/modules.d.ts
@@ -68,11 +68,11 @@ declare module 'random-access-storage' {
   type Cb<T> = (err: any, val?: T) => void
 
   class Request {
-    public type: number
-    public offset: number
-    public size: number
-    public data: Buffer
-    public storage: RandomAccessStorage
+    readonly type: number
+    readonly offset: number
+    readonly size: number
+    readonly data: Buffer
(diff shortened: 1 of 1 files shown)


types/modules.d.ts
```diff
@@ -6,7 +6,102 @@ declare module 'base32.js'
...
+    run(req: Request, writing?: boolean): void
+  }
+
+  export = RandomAccessStorage
+}
+declare module 'random-access-memory' {
+  import RandomAccessStorage from 'random-access-storage'
+
+  // TODO: Seems like many of the hypercore modules can also use Uint8Array
+  type HypercoreBuffer = Buffer | Uint8Array
```
achou11, line 84: this will affect other types additions for other hyper modules, but seems like a buffer by their definition can also be a Uint8Array in some cases

achou11: decided to just stick with Buffer for simplicity for now. may want to do an update for everything to account for Uint8Array later instead

sethvincent (approved):
🎉

PR #36, opened 2022-10-05
WIP: chore: add types for hyperswarm testnet

achou11:
Probably overdid it and typed a whole chain of hyper deps. Probably way more than was necessary but a good exercise in understanding the ecosystem a little better.

Lots of uncertainties all over the place, and there's some stuff in `@hyperswarm/dht` that I haven't fully figured out yet. This may be a pain to review so open to making this easier to review somehow.


sethvincent:
Wooo that's a lot of types! I did a quick read through and checked with tsc and I think this is looking good.

sethvincent (approved):

PR #38, opened 2022-10-11
authstore

sethvincent:
see the authstore readme in this pr to get an overview of functionality: https://github.com/digidem/mapeo-core-next/pull/38/files#diff-527a78f1ad3b8babf814bceb880b5b558e4d64cd431091e2c28850b56488b4fd


lib/authstore/index.js
```diff
@@ -0,0 +1,73 @@
...
+      corestore: options.corestore,
+      indexer: options.indexer,
+    })
+  }
+
+  async ready() {
+    await this.capabilities.ready()
+  }
+
+  async create({ identityPublicKey, creator, capability }) {
```
gmaclennan, line 41: The `creator` would not be user-settable right? Neither do we need to store it maybe? since we know it from the key of the hypercore it is stored in.

package.json
```diff
@@ -55,7 +55,9 @@
     "hyperswarm": "^4.3.3",
     "multi-core-indexer": "github:digidem/multi-core-indexer",
     "multicast-service-discovery": "^4.0.3",
+    "mutexify": "^1.4.0",
```
gmaclennan, line 58: Is this being used?

lib/authstore/index.js
```diff
@@ -0,0 +1,73 @@
...
+    // TODO: validate that user can write new capabilities based on current capabilities
+    return this.capabilities.create({
```
gmaclennan, line 43: I think this should also have a "fromIndex" property, which is:

```
const capabilitySubjectCore = this.corestore.get(identityPublicKey)
const fromIndex = capabilitySubjectCore.length
```

E.g. it should be the current length of the capability core of the device you are making the capability statement about.

lib/authstore/index.js
```diff
@@ -0,0 +1,73 @@
...
+      id: identityPublicKey,
+      creator,
+      capability,
+    })
+  }
+
+  validateCapability({ identityPublicKey, capability }) {}
+
+  getCapabilities(identityPublicKey) {
+    return this.capabilities.query(`identifyPublicKey == ${identityPublicKey}`)
```
gmaclennan, line 62: We need to start returning the block index from data store queries, so that we can validate the statement (because capabilities need to be "valid from" a particular point in a device's capability core history).

lib/authstore/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+The approach used so far makes it possible to define new types of statements about devices and their relationships. We could in the future, for example, add device statements that links two devices as being owned by the same user, or add a capability statement that makes it possible to add specific capabilities in addition to those specified by a role.
+
+### Project creator
+As the root of the tree of relationships between devices, the public key of the `AuthStore` keyPair of the project creator is also the public key of the project. A project creator can't be removed from a project, however a group of coordinators could decide to fork a project with a new project creator and retain the data from the existing project.
+
+### Capability data models
+Each type of capability is a `DataType`, each with its own schema, validation, and indexing. The data types are stored in a `DataStore` instance internal to the `AuthStore` class.
+
+See the available data schemas in [authtypes.js](authtypes.js).
```
sethvincent, line 52: to be updated when we place schemas in their forever home

lib/authstore/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+The device statements are used to add, remove, and restore device access. The role statements are used to assign the role and its associated capabilities to a device. The core ownership statements are used to prove that a device owns a specific hypercore.
+
+Each device in a project has a role. Roles define the capabilities a device has in the project. Roles are assigned by other devices. In assigning roles there is a tree of relationships between devices that can be used to validate a device's role & capabilities.
+
+There is a limited set of capabilities that can be assigned to each role:
+- `read` - read data & media in a project
+- `write` - create new data & media in a project
+- `edit` - edit data & media in a project
+- `manage:devices` - assign roles and add, remove, and restore devices in a project
```
sethvincent, line 26: `manage:devices` could be broken down to be more granular but i'm not sure that it is needed

lib/authstore/index.js
```diff
@@ -0,0 +1,1263 @@
...
+  /**
+   * @param {Object} options
+   * @param {IdentityKeyPair} options.identityKeyPair the key pair of the identity
+   * @param {KeyPair} options.keyPair the key pair used for the local writer hypercore
+   * @param {Corestore} options.corestore
+   * @param {import('../sqlite.js').Sqlite} options.sqlite an instance of the internal Sqlite class
+   * @param {AvailableRoles} [options.roles] an array of role names and capabilities
+   */
+  constructor(options) {
+    this.#identityKeyPair = options.identityKeyPair
```
sethvincent, line 26: todo: update so that identityKeyPair and keyPair are actually the same thing

lib/authstore/index.js
```diff
@@ -0,0 +1,1263 @@
...
+
+    const existing = await this.getRole({ identityId })
+    const authorRole = await this.getRole({ identityId: this.identityId })
+
+    if (existing && existing.role === role) {
+      throw new Error(`Role ${role} already set`)
+    }
+
+    if (existing && existing.role === 'project-creator') {
+      // TODO: we may want to allow this in the future
```
sethvincent, line 653: should probably remove this todo? for now things rely on a single project creator and removing them or changing their role means forking to a new project that retains existing data.

lib/datastore/index.js
```diff
@@ -1,132 +1,260 @@
...
-    const indexing = new Promise((resolve) => {
-      this.#indexer.onceWriteDoc(doc.version, (doc) => {
-        resolve(doc)
-      })
-    })
+    if (keyToId(key) === keyToId(this.#writer.key)) {
+      return this.#writer
+    }
 
-    this.validate(doc)
```
sethvincent, line 117: todo: bring back the validate method using mapeo-schema

lib/datatype/index.js
```diff
@@ -72,25 +100,44 @@ export class DataType {
     if (decode) {
       this.#decode = decode
     }
+
+    const secretKey = this.#keyPair.secretKey
+    const publicKey = this.#keyPair.publicKey
+
+    this.#writer = this.#corestore.get({
```
sethvincent, line 107: this is actually the same writer hypercore defined in datastore, so it may be that we can just pass that core in this constructor. the only difference really is just that we wouldn't be repeating this `corestore.get` boilerplate.

lib/datatype/index.js
```diff
@@ -108,4 +155,164 @@ export class DataType {
...
+    await this.#writer.ready()
+
+    const created = data.created || new Date().getTime()
+    const doc = Object.assign({}, data, {
+      id: data.id || randomBytes(8).toString('hex'),
+      authorId: this.identityId,
+      version: `${this.#writer.key.toString('hex')}@${this.#writer.length}`,
+      created,
+      timestamp: created,
+    })
```
sethvincent, line 213: todo: update properties with finalized schema

lib/discovery/index.js
```diff
@@ -9,7 +9,7 @@ import base32 from 'base32.js'
 /**
  * @typedef {Object} DiscoveryEvents
  * @property {(connection: NoiseSecretStream<RawConnectionStream>, info: PeerInfo) => void} connection
- * @property {(topicStatus: { topic: TopicString, mdns: TopicServiceStatus, dht: TopicServiceStatus }) => void} topicStatus
+ * @property {(topicStatus: { topic: TopicId, mdns: TopicServiceStatus, dht: TopicServiceStatus }) => void} topicStatus
```
sethvincent, line 12: Did a handful of renaming things in here to match the pattern that I ended up with in authstore:

- `<entityname>Key` or `<entityname>PublicKey` (if part of a keypair) for buffers
- `<entityname>Id` for hex string encoding of public key buffer

tests/authstore.js
```diff
@@ -0,0 +1,130 @@
+import test from 'brittle'
```
sethvincent, line 1: so far just writing out tests rather than using a strategy runner that takes json as input. the api was changing too much for that. i think it would be best to create an issue for adding a strategy runner later.

tests/scenarios/authstore.js
```diff
@@ -0,0 +1,158 @@
+export const scenarios = [
```
sethvincent, line 1: this isn't currently in use. it can either be deleted or we can just keep it around for later when we finish up the strategy runner approach to authstore tests.

lib/datatype/index.js
gmaclennan: I think this would be a good change to make, because if someone were to change the sign or verify functions in one place but not another I think that would create problems?

lib/datatype/index.js
```diff
@@ -108,4 +155,164 @@ export class DataType {
...
+  getById(id) {
+    return this.#sqlite.get(`select * from ${this.name} where id = '${id}'`)
+  }
+
+  /**
+   * Put a doc
+   * @param {Doc} data
+   * @returns {Promise<Doc>}
+   */
+  async put(data) {
```
gmaclennan, line 203: Maybe keep this as a private method?

lib/datatype/index.js
```diff
@@ -108,4 +155,164 @@ export class DataType {
...
+    await this.#writer.ready()
+
+    const created = data.created || new Date().getTime()
+    const doc = Object.assign({}, data, {
+      id: data.id || randomBytes(8).toString('hex'),
+      authorId: this.identityId,
+      version: `${this.#writer.key.toString('hex')}@${this.#writer.length}`,
```
gmaclennan, line 210: We would not actually store this on the doc written to the database, since it can be calculated on read (and putting it in the doc risks `this.#write.length` not being equal to the actual index when stored because of a race condition).

lib/datatype/index.js
```diff
@@ -108,4 +155,164 @@ export class DataType {
...
+  }
+
+  /**
+   * Create a read stream of blocks from the local writer hypercore
+   * @return {Promise<void>}
+   */
+  async close() {
+    await this.#writer.close()
+    await this.#corestore.close()
+    await this.#sqlite.close()
```
gmaclennan, line 316: Promise.all()? or is order important?

gmaclennan: Also, I think a class should only be responsible for closing instances that it "owns". E.g. if this class did not "open" these instances, then it should not close them.

lib/datatype/indexer/index.js
```diff
@@ -82,6 +76,7 @@ export class Indexer {
           typeof value === 'object' &&
           ['links', 'forks'].includes(key) === false
         ) {
+          /* @ts-ignore */
```
gmaclennan, line 79: I realize this change is not in this PR, but just spotting this now and wondering how we know to parse these values on read?

lib/datastore/index.js
```diff
@@ -1,132 +1,260 @@
...
 
-    const indexing = new Promise((resolve) => {
-      this.#indexer.onceWriteDoc(doc.version, (doc) => {
-        resolve(doc)
-      })
+  /**
+   * Wait for the indexer to finish available blocks
+   * @returns {Promise<void>}
+   */
+  async indexing() {
```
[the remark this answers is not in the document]
gmaclennan: We need to be very careful about how this might be used outside a testing context (ok to use it as a helper in tests).
In real usage a device is going to be switching between "indexing" and "idle" multiple times — imagine sync over a slow network where new blocks are arriving more slowly than blocks are indexed. You might want to consider "indexing is done when nothing new is arriving for XX seconds". 
For now this is used only in the test helper, but I want to make sure we don't leave it exposed and start using it in other code without considering the limitations/fagility of it.

tests/helpers/index.js
```diff
@@ -1,18 +1,50 @@
...
+export function replicate(peers) {
+  for (const peer1 of peers) {
+    for (const peer2 of peers) {
+      if (peer1 === peer2) continue
+      const stream1 = peer1.authstore.replicate(true, { live: true })
+      const stream2 = peer2.authstore.replicate(false, { live: true })
+      stream1.pipe(stream2).pipe(stream1)
+    }
+  }
+}
```
gmaclennan, line 31: Won't this duplicate replication?
e.g. two peers A, B then this will do `A.pipe(B).pipe(A)` _and_ `B.pipe(A).pipe(B)`

sethvincent: ah that would! fixed in recent commit.

lib/datastore/index.js
```diff
@@ -1,132 +1,260 @@
...
+   * @param {Object} options
+   * @param {String} options.name
+   * @param {Object} options.schema
+   * @param {BlockPrefix} options.blockPrefix
+   * @param {String} options.extraColumns
+   * @returns
    */
-  encode(doc) {
-    return this.#dataType.encode(doc)
+  dataType(options) {
```
sethvincent, line 116: from @gmaclennan:

> Also wondering about the way that datatypes can be added after creating the datastore instance. We don't restart indexing when adding a new datatype, so as I understand it this can result in data not being indexed, e.g. create DataStore instance on cores with existing data -> starts indexing -> a few ticks later call dataStore.dataType() to add a data type -> items indexed by multi-core indexer will not be included in the index.

> I'm wondering if the datatypes should just be in the constructor to simplify things? Also it seems at first glance that cores would be added multiple times to the multicore indexer, because they are added on initialization of the datastore, then from what I understand the same cores are re-added for each datatype that is added, since we basically one core for each device in the project.

tests/datatype.js
```diff
@@ -23,7 +23,7 @@ test('datatype - create, encode, decode', async (t) => {
   })
 
   const updated = await dataType.update(
-    Object.assign({}, created, { title: 'hi' })
+    Object.assign({}, created, { title: 'hi', links: [created.id] })
```
gmaclennan, line 26: This should be `created.version`. I wonder if we should validate if a given version actually exists? It wouldn't actually require a read, we could just parse the hypercore key and index from the version and check that hypercore has length > index. Although perhaps maybe also worth validating the record that is being linked to check it is of the same data type? To avoid really weird data.

gmaclennan:
Tangential naming comment: I see we're using `projectId` to refer to the project key. I think we will want to expose the project ID in the user interface, and maybe it data shared externally, so I think it needs to be different from the project key (we could use the discovery key, or we could use a different hash of the key), so maybe we call this `projectKey` and reserve `projectId` for the hashed key for future use.

sethvincent:
I think the name for the hashed discovery key should be `projectDiscoveryKey` (`projectDiscoveryId` for the base32 encoded version) to keep with the hypercore pattern.

gmaclennan:
Makes sense to me. Just to bear in mind that the likely user-facing name for `projectDiscoveryId` will be "Project ID". We might also choose to only use the first 64-bits as the user-facing ID (see this discussion: https://github.com/digidem/mapeo-core/issues/104)
But I think the naming for code purposes that you have makes sense.

lib/datatype/indexer/index.js
sethvincent: Yeah that's currently in the sqlite wrapper: https://github.com/digidem/mapeo-core-next/pull/38/files/82fb39f5097565d9e9c8238c0f7d7be01b2b15f2#diff-f391aa92bfbc3da0b7157d69e715a17508f09dc1fbd520a49befb3b446ede72eR32

Ideally the mapeo-schema encode function would handle this part. But for parsing json from sqlite maybe we need to read the json schema and parse values based on that.

tests/datatype.js
sethvincent: :+1: fixed the test type and added validation that checks that links exist and are of the same datatype.

lib/authstore/index.js
```diff
@@ -0,0 +1,1323 @@
...
+  #availableRoles
+  #datastore
+  #sqlite
+
+  /**
+   * @param {Object} options
+   * @param {PublicKey} [options.projectPublicKey] the public key of the project. Only needed if user is not the project owner.
+   * @param {KeyPair} options.keyPair The key pair used for the local writer hypercore.
+   * @param {IdentityKeyPair} options.identityKeyPair The key pair of the identity.
+   * @param {KeyPair} options.keyPair The key pair used for the local writer hypercore.
```
tomasciccola, line 22: duplicated param??

sethvincent: ah, yeah. thanks!

gmaclennan:
I've been thinking more about this recently and the API code/structure of this "auth store". It is actually quite similar to the main data store, in that we have, or we plan to add, different record types to here (we already have the role records and ownership records, and I think we plan device name records and potentially project info records). The only real difference is that we have a validation step for these records. I wonder if it makes sense to namespace this store similarly to the data store, e.g. with `authStore.role` and `authStore.ownership` namespaces, with common CRUD methods? Could that be easier to reason with for the API consumer?

PR #39, opened 2022-10-13
chore: add more tests for discovery

### LINKED ISSUE #18: write additional tests for discovery.js module
module file: https://github.com/digidem/mapeo-core-next/blob/main/lib/discovery.js
existing tests: https://github.com/digidem/mapeo-core-next/blob/main/tests/discovery.js

Tasks:
- make list of types of tests needed (look at modules like hyperswarm for ideas)
- write tests

achou11:
Closes #18. May add more to this but can iterate later on too.

Adds tests for:

- Replication over mDNS
- Ensuring that connections are made over valid discovery mechanism

### DIFF
--- a/tests/discovery.js
+++ b/tests/discovery.js
@@ -186,6 +186,79 @@ test('discovery - mdns', async (t) => {
   await discover2.join(keyPair.publicKey)
 })
 
+test('replication - mdns', async (t) => {
+  t.plan(3)
+
+  const testnet = await createTestnet(10)
+  const bootstrap = testnet.bootstrap
+
+  const core1 = new Hypercore(ram)
+  await core1.ready()
+
+  const core2 = new Hypercore(ram, core1.key)
+  await core2.ready()
+
+  core1.append(['a', 'b', 'c'])
+
+  const identity1 = createIdentityKeys()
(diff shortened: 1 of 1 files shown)


tests/discovery.js
```diff
@@ -257,3 +330,58 @@ test('discovery - mdns join, leave, join', async (t) => {
...
 })
+
+// Create a stress test with many peers to see if the connections made using the proper discovery mechanisms
+// i.e. mdns peers will only connect with mdns-enabled peers and dht peers will only connect with dht-enabled peers
+test('discovery - valid connection discovery types', async (t) => {
+  let connectionsCount = 0
+
+  const keyPair = createCoreKeyPair('stress test')
+
+  const instances = Array(20)
```
sethvincent, line 341: out of curiousity I tried increasing this number and after about 75 the tests will start to hang without showing errors. not a blocker at all just kinda interesting!

we might want to explore whether that's a network limitation or a number of mdns/dht servers on a device limitation, but that's outside the scope of this pr. I think 20 is a good number for this.

sethvincent (approved):
this is great! i added one note that we might decide to make an issue for but i think this is ready to merge. if we come up with other discovery tests i think we can make issues for those as we think of them.

PR #41, opened 2022-10-17
chore: add types for hyperswarm

achou11:
Towards #14 

Required some code changes that I'm not entirely sure about (although existing tests pass).

### DIFF
--- a/lib/discovery/index.js
+++ b/lib/discovery/index.js
@@ -298,7 +298,7 @@ export class Discovery extends TypedEmitter {
    *
    * @private
    * @param {NoiseSecretStream<RawDhtConnectionStream>} connection
-   * @param {HyperswarmPeerInfo} info
+   * @param {import('hyperswarm').PeerInfo} info
    * @returns {Promise<void>}
    */
   async _onDhtConnection(connection, info) {
@@ -416,7 +416,9 @@ export class Discovery extends TypedEmitter {
       topicBuffer,
       mdns: mdnsActive ? new MdnsDiscovery() : undefined,
       dht:
-        dhtActive && this.dht.join(topicBuffer, { server: true, client: true }),
+        dhtActive && this.dht
+          ? this.dht.join(topicBuffer, { server: true, client: true })
+          : undefined,
     })
 
     this.#topics.set(topicHex, topic)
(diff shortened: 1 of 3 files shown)


lib/discovery/index.js
```diff
@@ -552,7 +554,7 @@ export class Discovery extends TypedEmitter {
       topic.mdns?.lookup(serviceType)
     }
 
-    if (dhtActive) {
+    if (dhtActive && this.dht && topic.dht) {
```
achou11, line 557: probably the most unsure code change

types/modules.d.ts
```diff
@@ -70,7 +70,130 @@ declare module 'time-ordered-set' {
...
+    banned: boolean
+    tried: boolean
+    explicit: boolean
+    queued: boolean
+    client: boolean
+    topics: Buffer[]
+    attempts: number
+    priority: number
+
+    constructor(opts: { publicKey: Buffer; relayAddresses: any[] })
```
achou11, line 134: slightly out of scope for this PR, but @sethvincent do you know what the type for the `relayAddresses` is?

lib/discovery/index.js
```diff
@@ -1203,21 +1205,21 @@ export class Topic extends TypedEmitter {
    * @param {TopicServiceStatus} [options.dhtStatus]
    * @param {TopicServiceStatus} [options.mdnsStatus]
    * @param {MdnsDiscovery} [options.mdns]
-   * @param {Hyperswarm} [options.dht]
+   * @param {import('hyperswarm').PeerDiscoverySession} [options.dht]
```
achou11, line 1208: based on usage, guessed that it was the discovery session we want to pass here, not the hyperswarm instance itself

lib/discovery/index.js
sethvincent: :+1: i think this makes sense

lib/discovery/index.js
sethvincent: ah yeah, that looks right

types/modules.d.ts
sethvincent: it looks like an array of objects with `host` and `port` properties based on hyperswarm/dht: https://github.com/hyperswarm/dht/blob/bb21a0191c1d96a727dcf82ffe3cb8967fb7c474/test/messages.js#L258

sethvincent (approved):
:+1:

types/modules.d.ts
achou11: ah nice! I'll update in a follow-up

PR #42, opened 2022-10-18
chore: add type annotation for relay address

achou11:
Towards #14. Follow up for https://github.com/digidem/mapeo-core-next/pull/41#discussion_r997452665

Was thinking about defining the `RelayAddress` interface in `dht-rpc` but that seemed to low-level. Placing it in `@hyperswarm/dht` feels appropriate to me

### DIFF
--- a/types/modules.d.ts
+++ b/types/modules.d.ts
@@ -72,7 +72,7 @@ declare module 'time-ordered-set' {
 }
 declare module 'hyperswarm' {
   import EventEmitter from 'events'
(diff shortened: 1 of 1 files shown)


achou11:
hmm just realized that the `RelayAddress` has the same interface as `DhtNode`. @sethvincent thoughts on whether this should be unified or distinct? Based on context, seems like they're technically different but maybe they're just loose with terminology 😄

sethvincent (approved):

PR #44, opened 2022-10-18
chore: add types for b4a

achou11:
Towards #14 

- Only adds definitions for the Node implementation, not the browser one, which seems out of scope (interested to have them but don't know how to create the separation properly in a declaration file, if possible)


achou11:
For reference, maybe this is how you can declare for different environments? https://github.com/speakeasyjs/base32.js/blob/master/index.d.ts

PR #55, opened 2022-12-21
staged sync

sethvincent:


lib/CoreReplicator.js
```diff
@@ -0,0 +1,46 @@
...
+* @module CoreReplicator
+ */
+
+import { CoreIdCache } from "./authstore/coreIdCache"
+import { Sqlite } from "./sqlite"
+
+export class CoreReplicator {
+  /** @type {string} */
+  #coreId
+  #masterCore
```
sethvincent, line 11: let's just call this `#core`

lib/CoreReplicator.js
```diff
@@ -0,0 +1,46 @@
...
+    this.#coreId = coreId
+    this.#store = store
+    this.#coreIdCache = coreIdCache
+    this.#masterCore = this.#store.get({key:this.#coreId})
+  }
+
+  /**
+* @param {StoreType} type
+  */
+  replicate(type){
```
sethvincent, line 36: This name would get a bit confusing because of the existing `hypercore.replicate` method. We could instead use something like `CoreIdExtension` and maybe this method could be called `share`? I don't know if those ideas are very good, though. This is a difficult one to name, haha.

lib/authstore/coreIdCache.js
```diff
@@ -0,0 +1,57 @@
+/*
+* @module CoreIdCache
+*/
+import { Sqlite } from "./../sqlite.js"
```
sethvincent, line 4:
```suggestion
```

lib/CoreReplicator.js
```diff
@@ -0,0 +1,46 @@
...
+  #extensionNamespace = 'replicator'
+  /** @type {any} */
+  #extension
+  /**
+* Create a CoreReplicator instance. This class is in charge of replicating cores belonging to a namespace in a corestore with other peers. This is acomplished by registering an hypercore extension.
+* @param {string} coreId - core public key to use as master core to register the extension in
+* @param {Corestore} store - set of cores to share
+* @param {CoreIdCache} coreIdCache  
+  */
+  constructor(coreId, store, coreIdCache){
```
sethvincent, line 26: It might be best to pass in the core rather than the coreId. This is something that I'm changing in DataType code right now, too, so that it takes a core rather than a keyPair for a core.

sethvincent:
I think it would make sense to move both coreIdCache.js and CoreReplicator.js into the new lib/sync directory I just made today.

tomasciccola:
awesome, thanks for the comments!
sync seems to be on the root dir, not on lib.But I'll move the CoreIdCache and CoreIdExtension to sync

sethvincent:
@tomasciccola This is looking cool! I updated this and the sync branch with latest changes from the authstore branch.

Some tests for `CoreIdExtension` would be good and then I think we can get this merged.

Since we realized that we only need to send the authstore keys over the extension and the rest of the keys will be shared through blocks in authstore that means the test for `CoreIdExtension` can be pretty simple, probably just making sure that two peers can exchange their keys (though if you have ideas for other tests I'm not thinking of that would also be good).

tomasciccola:
yeah, right. I'll update CoreIdExtension and add some tests to it. 
So, instead of sharing through it smth like `{coreIds:[], namespace: '', identityId:''}` we are only sharing the list of coreIds, right? we're not passing neither namespace (since always 'auth') nor identityId (authstore will provide this?)

sethvincent:
Right, the extension doesn't need to send the namespace. I'm realizing now that it should send all the coreIds it knows about (not just the identity's cores but all cores), so it could send something like:

```js
[
  {
    coreId: String,
    identityId: String
  },
  ...
]
```

I think the CoreIdCache is still useful for tracking which cores exist so we can do things like get them all through `corestore.get` when first loading an app.

gmaclennan:
I've just been reading through this and thinking about how the extensions and replication will work in practice.

The problem we are solving is that while hypercore will replicate every hypercore that is injected into the replication stream (via `core.replicate(stream)`), two peers may not have prior knowledge of the public key, which is needed to create the instance of the core in the first place.

The solution here is an "push" approach that sends all keys on first connection. However we need to also handle how we handle additional cores being added to the stream during replicaiton, e.g. Peer A connects to Peer B, sends a list of cores. Then Peer C connects to Peer A, and Peer A gets new cores. Peer A then needs to send any new cores it has received to Peer B. However to know which are "new" then Peer A needs to track what it has already sent to B, unless we just send the complete list. However sending the complete list can result in cycles of traffic, e.g. Peer B now connects to Peer C, so when B receives a new list of cores from A, then it sends a list to C. This is manageable but requires a lot of code to track.

I think a better solution might be to switch to a "pull" approach. `Hypercore.createProtocolStream` gives us an [`ondiscoverykey` callback](https://github.com/holepunchto/hypercore/blob/ab79e538350b109adfa528cb752d275dd0e90882/index.js#L153-L155) that is called for each hypercore which is added to the stream (undocumented unfortunately, like many things in Hypercore). We can register two extensions: "want core" - a message is just a discovery key, and "have core" - a core public key.

For an example, two peers, A & B. A has cores X, Y. B has no cores.

1. `A` & `B` connect.
2. `A` adds cores X and Y to the stream
3. `ondiscoverykey()` for `B` will fire with `discoveryKey(X)` and `discoveryKey(Y)`. This is not enough to start replicating - `B` needs the public keys for X and Y.
4. `B` sends two extension messages: "want core" `discoveryKey(X)` and "want core" `discoveryKey(Y)`.
5. `A` receives these messages, and validates that these cores are auth store keys (since we other public keys are shared in auth store records).
6. `A` sends two extension message: "send core" `publicKey(X)` and "send core" `publicKey(Y)`.
7. `B` receives the two public keys, and can now inject them into the stream and replicate them with `A`

This results in more, smaller messages over the protocol stream, but when two peers are missing core keys. If they each have all the keys then no messages need to be send (`ondiscoverykey` will fire, and each client will see they already have the key they need).

We can improve message efficiency by just binary encoding of the extension, it can just be a 32-byte buffer (we should validate the buffer length, maybe just ignore and log if we receive not 32-bytes).

---

I was also looking at how [multifeed manages this](https://github.com/kappa-db/multifeed/blob/master/mux.js), and I don't think we need that much complexity. As I understand it, protomux now handles versioning of the protocol (we can check with Mafintosh maybe) and the way our replication works, it doesn't actually need to know about the extension messages if each peer has prior knowledge of the public keys — in that case it is just normal hypercore replication.

lib/sync/CoreIdExtension.js
```diff
@@ -0,0 +1,39 @@
...
+  * @param {Object} peer - socket peer
+  */
+  /**
+  * Create a CoreReplicator instance. This class is in charge of sharing a set of cores by registering an hypercore extension.
+  * @param {Hypercore} core - core to use as master core to register the extension in
+  * @param {import('./CoreIdCache.js').CoreIdCache} coreIdCache the cache to retrieve the set of coreIds to share
+  * @param {OnMessage} [onmessage] - callback that handles recieved messages from a peer
+  */
+  constructor(core, coreIdCache, onmessage = () => {}){
+    this.#core = core
```
gmaclennan, line 24: It's really important that these extensions are _only_ added to the project creator auth store core (the core that is used for the initial connection). We should document that somewhere clearly.

lib/sync/CoreIdCache.js
```diff
@@ -0,0 +1,66 @@
...
+export class CoreIdCache {
+  #sqlite
+  #tableName = 'coreIdCache'
+  #types = `'auth', 'data', 'blob'`
+  /**
+    * Create a CoreIdCache instance. This class handles a table to track cores belonging to different namespaces/stores (Authstore, Datastore, Blobstore)
+    * @param {import('../sqlite.js').Sqlite} sqlite an instance of better-sqlite3
+  */
+  constructor(sqlite){
+    this.#sqlite = sqlite
```
gmaclennan, line 13: I'm not sure it makes sense to use our Sqlite class here? It does extra work parsing links and forks fields, which don't exist here, and creates an abstraction that makes it harder to parse what is happening. It also does not allow us to pre-prepare the statement to improve efficiency of gets.

lib/types.js
```diff
@@ -98,10 +98,25 @@
...
+/**
+  * @typedef {GeneralObject} CoreIdRecord
+  * @property {CoreId} coreId
+  * @property {StoreNamespace} namespace
+  * @property {IdentityId} identityId
+ */
+
+/**
+  * @typedef {GeneralObject} CoreIdRecordAggregate
+  * @property {CoreId[]} coreIds
```
gmaclennan, line 114: In what scenario would be have more than one coreId for a given namespace and identityId? Would that not be an error?

lib/sync/CoreIdCache.js
```diff
@@ -0,0 +1,66 @@
...
+    this.#sqlite.run(`
+    CREATE TABLE IF NOT EXISTS ${this.#tableName}
+                  (
+                    coreId TEXT PRIMARY KEY NOT NULL,
+                    namespace TEXT CHECK(namespace IN(${this.#types})) NOT NULL, 
+                    identityId TEXT NOT NULL
```
gmaclennan, line 19: Maybe index by discovery key (e.g. make discovery key the primary key). See https://github.com/digidem/mapeo-core-next/pull/55#issuecomment-1384001134 - we would likely want to lookup a core by discovery key.

PR #61, opened 2023-01-17
feat: core-manager

gmaclennan:
These are some ideas, ~~WIP~~, for a "CoreManager" class for Mapeo.

It eagerly loads all the hypercores in the project, including writer cores for each namespace.

I am imagining that each module might use the `getCores(namespace)` method to get the cores to add to the indexer, and would also listen to the `add-core` event. And each module would also use the `getWriter(namespace)` method.

A thought about the extensions to share keys: what if we don't use extensions at all, and use the data in authStore for all keys? This would require the public key of the auth store from the peer being invited, and for us to write it as part of the record that adds that peer to the project.

That way auth store would be solely responsible for calling the `addCore` method as needed, but auth store would also need to be responsive as new data is added during sync.

~~Will write more tomorrow when my head is clearer.~~

### DIFF
--- a/lib/core-manager/core-index.js
+++ b/lib/core-manager/core-index.js
@@ -1,0 +1,121 @@
+import crypto from 'hypercore-crypto'
+import { TypedEmitter } from 'tiny-typed-emitter'
+
+/** @typedef {import('./index.js').Namespace} Namespace */
+/** @typedef {{ core: import('hypercore').default, key: Buffer, namespace: Namespace }} CoreRecord */
+/**
+ * @typedef {Object} CoreIndexEvents
+ * @property {(coreRecord: CoreRecord) => void} add-core
+ */
+
+// WARNING: If changed once in production then we need a migration strategy
+const TABLE = 'cores'
+const CREATE_SQL = `CREATE TABLE IF NOT EXISTS ${TABLE} (
+  publicKey BLOB NOT NULL,
+  namespace TEXT NOT NULL
+)`
+
+/**
+ * An in-memory index of open cores that persists core keys & namespaces to disk
+ *
+ * @extends {TypedEmitter<CoreIndexEvents>}
+ */
+export class CoreIndex extends TypedEmitter {
+  /** @type {Map<string, CoreRecord>} */
+  #coresByDiscoveryId = new Map()
+  /** @type {Map<Namespace, CoreRecord>} */
+  #writersByNamespace = new Map()
+  #addCoreSqlStmt
+
+  /**
+   * @param {object} options
+   * @param {import('better-sqlite3').Database} options.db better-sqlite3 database instance
+   */
+  constructor ({ db }) {
+    super()
+    // Make sure table exists for persisting known cores
+    db.prepare(CREATE_SQL).run()
+    // Pre-prepare SQL statement for better performance
+    this.#addCoreSqlStmt = db.prepare(
+      `INSERT OR IGNORE INTO ${TABLE} VALUES (@publicKey, @namespace)`
+    )
+  }
+
+  [Symbol.iterator]() {
+    return this.#coresByDiscoveryId.values()
+  }
+
+  /**
+   * NB. Need to pass key here because `core.key` is not populated until the
+   * core is ready, but we know it beforehand.
+   *
+   * @param {Object} options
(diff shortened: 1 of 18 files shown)


gmaclennan:
> A thought about the extensions to share keys: what if we don't use extensions at all, and use the data in authStore for all keys? This would require the public key of the auth store from the peer being invited, and for us to write it as part of the record that adds that peer to the project.

Thinking about this more overnight, I don't think this is a good idea because it adds complexity to the invite process - it would add another step, since auth store keys are derived from the project key:

1. Joiner shares public key
2. Inviter encrypts project key and sends to inviter
3. Joiner generates auth store key and sends to inviter
4. Inviter writes auth store key to permissions record in their auth store

Step (3) would make it more complex to complete an invitation, so I think it is better to use an extension to share auth store keys.

gmaclennan:
Ok this is now working and in a good place to review. Most things are documented inline in JSDoc comments.

To initiate the CoreManager you need a `rootKey` — this is the project key (should probably rename - in fact feedback welcome on all naming decisions)

You can get the writer core for any namespace with `coreManager.getWriteCore()`. The returned value from this (and other methods) is a "CoreRecord", which is of type `{ core: Hypercore, key: Buffer, namespace: string }`. This is because `core.key` is null until the core is ready, but we already know the key. We could Proxy the core instance and append the key and namespace properties, but that might be confusing for future maintainers.

You can get all cores in a namespace with `coreManager.getCores(namespace)`.

All auth cores are gossiped between replicating peers via extensions on the root core and will managed internally, but cores in other namespaces need to be added. The auth store should do this: adding cores that are read from the auth cores with `addCore(key, namespace)`.

The replication function is not as flexible as `hypercore.replicate()`. It expects a Noise Secret Stream (e.g. `@hypercore/secret-stream`) as the argument.

I am thinking that the CoreManager should also be responsible for calling `core.download()` on all cores too, but not sure yet.

This needs some more tests too to catch different edge cases. However I think it's in a good place for an initial review.

gmaclennan:
I have implemented a single `mapeo` extension message type that I think we can use for all extension messages that we want to use (e.g. for gossiping sync state in the future). It's protobuf encoded, so we can add messages as optional fields (like [Hyperbee does](https://github.com/holepunchto/hyperbee/blob/master/schema.proto)). Hypercore uses [compact-encoding](https://github.com/compact-encoding/compact-encoding) for messages, but I think writing and maintaining that is more difficult than prototype buffers.

gmaclennan:
This builds upon and is a replacement for #55

lib/messages.js
```diff
@@ -0,0 +1,40 @@
+// @ts-nocheck
+/* eslint-disable */
+import _m0 from "protobufjs/minimal.js";
+function createBaseExtension() {
+    return { authCoreKeys: [], wantCoreKeys: [] };
+}
+export var Extension = {
```
sethvincent, line 7: seems like this would this be in the core-manager directory.

it could use a more detailed name. something like `CoreExchangeExtension`?

gmaclennan: I was thinking we would use a single extension for all messages we send (e.g. for sharing sync state), rather than register a separate extension for each one. This is inspired by how [hyperbee uses extensions](https://github.com/holepunchto/hyperbee/blob/master/schema.proto). For additional messages we would them as optional messages attached to the extension class.

That said, maybe `MapeoExtension` is a better name? And since I think the core manager would likely manage all extensions, I think it makes sense to move to there.

sethvincent:
The base for this PR will need to be changed from `sync` to `main`.

gmaclennan:
Had trouble rebasing this, ended up squashing all commits then applying just the commits since the `sync` branch onto `main`. I hope this doesn't mess things up.

sethvincent:
I think this is ready to merge unless you wanted to do anything else with it @gmaclennan.

It makes sense to merge this before #60 and then I'll make needed revisions to incorporate the core manager.

gmaclennan:
Let me just spend the next 30 mins reviewing coverage and adding missing tests.

gmaclennan:
ok, go ahead and merge. I've created [an issue](https://github.com/digidem/mapeo-core-next/issues/71) for missing tests, can be done in a follow up

PR #62, opened 2023-01-23
feat: blobstore

### LINKED ISSUE #52: BlobStore
To store images and other types of blobs we'll implement a module called something like blobstore.

This will include:

- using [hyperdrive-next](https://github.com/hypercore-protocol/hyperdrive-next)
- storing original images and smaller preview versions separately
- allowing mobile devices to sync only preview versions and desktop devices to sync previews and originals
- keeping the possibility open to add new blob types: audio, video, etc.
  - this will likely be modeled in a similar way to the current `DataStore` implementation. We might have a BlobStore with multiple BlobTypes similar to how DataStore has multiple DataTypes.
- _other details I'm not thinking of right now_

gmaclennan:
An initial implementation of a BlobStore, fixes #52


gmaclennan:
Rebased on main

sethvincent (approved):
Like with the invites PR I read through this again and it looks good to me. Excited to merge it and work on integrating everything 👍

gmaclennan:
Read through this again today and did some cleanup and additional tests. Before merging I want to:

**Additional tests**
- [ ] Cancel download at different stages
- [ ] get() when data has not synced
- [ ] createReadStream() when data has not sync

**Ok for after merge**
- [ ] #81 
- [ ] #80

sethvincent:
### downloading multiple types and sizes

The behavior of `download` is somewhat confusing when someone tries to call it with these types and sizes:

```js
blobstore.download({
  types: ['photo', 'audio']
  sizes: ['original', 'preview']
})
```

Silently ignore `/audio/preview`? Throw error? I'd lean toward throwing an error because this is a programmer mistake and we need to know `/audio/preview` doesn't exist.

It seems like there's not a clean solution to downloading multiple types with multiple sizes aside from doing something like:

```js
{
  photo: ['original', 'preview'],
  audio: ['original']
}
```

Or maybe an array of objects with `type` and `sizes` properties.

The type checking could also know which sizes are appropriate for each blob type.

### variant instead of size

Using `variant` instead of `size` might be useful for storing different representations of a blob.

Example: If we wanted to store a frame of a video to create a thumbnail we could have a `thumbnail` variant for videos and the `variant` term is more appropriate.

### side note: template literal types

As an aside we could use [template literal types](https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html) for filepaths if that ever seems needed. Probably not but it's a fun types rabbit hole.

gmaclennan:
Thanks @sethvincent, this is helpful feedback. I made the suggested changes. Regarding template literals, yeah, definitely possible, although I think keeping the concept of "paths" as an implementation detail is helpful - it keeps the module focussed on what is needed for the Mapeo API.

gmaclennan:
Ok tests are added, this is good to merge from my perspective. @sethvincent go ahead and merge if you approve of the changes.

sethvincent:
:tada:

PR #70, opened 2023-02-07
feat: invites

gmaclennan:
Support project invites over an RPC channel that can be muxed on the same stream as hypercore replication.

Implemented as an RPC class, allowing future support of other communication over RPC as needed (e.g. for app updates)

In a follow-up PR: Add additional properties to the invite (e.g. project name) as needed by designs


gmaclennan:
Rebased on main

sethvincent (approved):
This looks good to me! After looking through it again I still don't have any specific concerns. It'll be best to merge this and make issues for anything that comes up during integration with other modules.

PR #85, opened 2023-03-02
fix: persist cores

### LINKED ISSUE #86: bug: persisted cores are not loaded
Somehow I forgot to load persisted cores when starting a new instance of CoreManager.

Failing test: https://github.com/digidem/mapeo-core-next/pull/85/commits/52269cd4fa17a9d20adc2032eba4e66a391afb3e

Work to fix: #85

gmaclennan:
Persist cores that are added to the core manager. Fixes #86 

Todo:

- [x] Add coreManager.close() method
- [x] Load and open persisted cores from storage.

### DIFF
--- a/lib/core-manager/core-index.js
+++ b/lib/core-manager/core-index.js
@@ -1,45 +1,17 @@
 import crypto from 'hypercore-crypto'
-import { TypedEmitter } from 'tiny-typed-emitter'
 
 /** @typedef {import('./index.js').Namespace} Namespace */
-/** @typedef {{ core: import('hypercore').default, key: Buffer, namespace: Namespace }} CoreRecord */
-/**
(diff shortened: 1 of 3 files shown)


sethvincent (approved):
:+1: looks cool!

The only thing I was confused by was moving the `CREATE_SQL` and `addCoreSqlStmt` to core-manager/index.js. It seemed like having them in core-index.js was a good separation. Not really a big deal though if the CoreIndex class is internal to CoreManager.

gmaclennan:
Yeah, I did not like that either. Because events need to be added to cores loaded from persistence (which happens in the parent class) it became a very leaky abstraction to keep the persistence in the index. Couldn't figure out a neat way to separate it, but yeah, it's not pretty but it works

PR #87, opened 2023-03-03
feat: support block encryption

gmaclennan:
Add support for block-level encryption of all hypercores. Separate encryption key for each namespace, to allow for future features (e.g. only sharing encryption key for specific namespaces). 

TODO after PR is merged:

- [ ] Add tests for encryption support to BlobStore #90


sethvincent (approved):
:+1:

PR #88, opened 2023-03-03
feat: add projectConfig & encryptionKeys to invite

gmaclennan:
Adds support for sending multiple encryption keys (one for each namespace) along with an invite, and also adds support for an optional "projectConfig" buffer.

Actual handling of the projectConfig in the invite will need to be done in a follow up PR.

### DIFF
--- a/lib/rpc/index.js
+++ b/lib/rpc/index.js
@@ -110,7 +110,7 @@ class Peer {
 
 /** @extends {TypedEmitter<MapeoRPCEvents>} */
 export class MapeoRPC extends TypedEmitter {
-  /** @type {Map<string, Peer} */
+  /** @type {Map<string, Peer>} */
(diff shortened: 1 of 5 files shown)


sethvincent (approved):
This looks good to me!

As an aside I'm wondering if we'd be able to cycle the keys used for encrypting the different cores or not. Probably need an issue for that if it's something we want to aim for.

gmaclennan:
> As an aside I'm wondering if we'd be able to cycle the keys used for encrypting the different cores or not. Probably need an issue for that if it's something we want to aim for.

I'm not sure how we would implement that (namely the coordination in a p2p system, but also working within disk space limitations), but it's an interesting challenge! I created [an issue](https://github.com/digidem/mapeo-core-next/issues/89) to track it.

Issue #65, opened 2023-01-31
ConfigStore


sethvincent:
The way we store project config will change in this new version of mapeo-core.

We'll create a ConfigStore that's based on DataStore that's somewhat similar to the implementation of AuthStore.

To start this will have a couple of document types:

- DeviceInfo
  - deviceName
- ProjectInfo
  - projectName

Full schemas TBD.

gmaclennan:
Just to clarify what we mean by project config in this context. I think there are distinct pieces of data:

1. Metadata about a project, like project name, maybe in the future "description" or "dates" etc.
2. Preset(s) - match tags to display data / collect data according to a certain schema
3. Field(s) - define how to display tags on a record
4. Icon(s) - SVGs or PNGs associated with a preset

We had discussed these being stored in the auth cores so that they are available to a newly invited project member as soon as the auth stores sync, to avoid needing to wait for the data cores to sync. I'm not so sure about that now — it introduces a lot of "noise" in the auth cores. However we need to think how we handle invites and syncing of data beyond just the auth cores.

We could store all of this as a single "project config" record, or normalize it into separate records for each preset, field and icon. For future-proofing I think normalized data is better, since it will reduce conflicts if we enable config editing on the client in the future, and it will also make it easier to match using historical data (e.g. imagine a preset changing so that it no longer matches previously collected data, if we had a preset history it would be easier to match, rather than matching against a history of the entire project config record).

TL;DR, I think we should store a record for each preset, each field, and each icon (I think icons in the data store rather than the blob store? just because they are more closely "attached" to presets, and don't have the same concepts of previews and originals). I think we should store them in the data store and not the auth store.

sethvincent:
Working on this today I realized this could pretty much just be an instance of `DataStore`.

Right now I'm making a `ConfigStore` class that extends `DataStore` just in case we might need some unique config-related methods.

A couple methods that may be needed:
- take all the records for one type of config and reduce them to one object
- query all config records of all types and return the big config object that we get in a mapeosettings file

ximenabb:
Gregor to help on this ticket

gmaclennan:
Do you think this should wait on #111 or go ahead with what is there currently so we can get something working?

gmaclennan:
Closing in favour of:

- https://github.com/digidem/mapeo-core-next/pull/187
- https://github.com/digidem/mapeo-core-next/issues/182
- https://github.com/digidem/mapeo-core-next/pull/174

PR #92, opened 2023-03-09
wip: initial work on configstore

sethvincent:


lib/configstore/config-types.js
```diff
@@ -0,0 +1,296 @@
+// TODO: move to / use mapeo-schema
+
```
sethvincent, line 2: @gmaclennan this file is a rough start to the schemas (partially pulling from mapeo-schema) that has a few questions about the schemas and how certain types of config should be handled

gmaclennan:
Re. from #65 

> A couple methods that may be needed:
> - take all the records for one type of config and reduce them to one object
> - query all config records of all types and return the big config object that we get in a mapeosettings file

I don't think either of these are needed - on mobile we actually split up this single object and [turn presets and fields into arrays](https://github.com/digidem/mapeo-mobile/blob/develop/src/frontend/api.ts#L489-L500). We just need to add a `getMany()` or `list()` function to the Datastore that will also be used for presets and fields (we will need this for observations too).

It could potentially be useful to join fields to presets on the server, although I don't see a huge issue if we do the join on the client (e.g. making multiple request, first for the preset, then for each field). Since we have very low latency I don't think this should be an issue?

I am wondering about the preset <-> fields relationship: should we maybe store the field versionId as well as the field docId? This would be useful if we are using historical presets to match old data: it would ensure we knew that data was collected with a previous version of a field definition. Not sure about this though, needs more thinking...

gmaclennan:
Fixes #65

PR #109, opened 2023-04-10
Block non-local connections on mdns server 

### LINKED ISSUE #68: Block non-local connections on mdns server
We [create a tcp server](https://github.com/digidem/mapeo-core-next/blob/f296696881863d74cc23aca0cf694840038edb5b/lib/discovery/index.js#L73) for listening to mdns connections, but theoretically this could receive connections from outside the local network.
We should reject/block incoming connections from non-local addresses. We can maybe do this by checking [`socket.remoteAddress`](https://nodejs.org/api/net.html#net_socket_remoteaddress) against private IP ranges with something like https://github.com/mafintosh/bogon
The reason for doing this is a security precaution, particularly if we are sending invites to peers discovered through mdns, and we want to be sure that incoming connections are local network only.

tomasciccola:
This should close #68

### DIFF
--- a/lib/discovery/index.js
+++ b/lib/discovery/index.js
@@ -5,6 +5,7 @@ import NoiseSecretStream from '@hyperswarm/secret-stream'
 import Hyperswarm from 'hyperswarm'
 import { MdnsDiscovery } from 'multicast-service-discovery'
 import base32 from 'base32.js'
+import bogon from 'bogon'
 
 /**
  * @typedef {Object} DiscoveryEvents
@@ -234,6 +235,9 @@ export class Discovery extends TypedEmitter {
    * @returns {Promise<void>}
    */
   async _onMdnsConnection(socket) {
+    // end connection if its from a non-local node
+    if(!bogon(socket.remoteAddress)) socket.end()
+
     if (socket.destroyed) {
       throw new Error('Socket destroyed')
     }


tomasciccola:
For this to be merged I think we need a way to test it. 
This would maybe mean a vps that makes a request to join a swarm id'ed via a topic. This request could be trigger from a request made inside the test to an endpoint. 
I don't know enough about mDNS to know if this would suffice...

sethvincent:
Yeah, that's a hard thing to write a test for, and in a way duplicates [bogon's tests](https://github.com/mafintosh/bogon/blob/master/index.js).

I think ensuring this works as expected is something that would be covered by an integration test suite (that doesn't exist yet and would take some effort to set up) so would be best to make an issue for making that test suite and referencing this pull request as one of the needed tests.

sethvincent (approved):

sethvincent:
@tomasciccola Approved with the small caveat that bogon should be in package.json dependencies.

lib/discovery/index.js
```diff
@@ -234,6 +235,9 @@ export class Discovery extends TypedEmitter {
    * @returns {Promise<void>}
    */
   async _onMdnsConnection(socket) {
+    // end connection if its from a non-local node
+    if(!bogon(socket.remoteAddress)) socket.end()
```
gmaclennan, line 239:
```suggestion
    if(!bogon.isPrivate(socket.remoteAddress)) socket.end()
```
We're concerned about checking this is from a local IP, not any Bogon.

gmaclennan (changes requested):
Yes, testing is hard for this. One way forward is to set up a virtual network with [mininet](http://mininet.org), but that would be a lot of work. I think it's ok to merge without tests for now, and we should ensure that we do manual integration tests for this (e.g. actually testing on real devices) at some point when the apps are ready.

lib/discovery/index.js
tomasciccola: Ooops, yeah, fixed that. 
Additionally, `socket.remoteAddress` returns an IPv6 with the IPv4 emebed (is it always like that or does it depend on some flag?); so I need to strip the `:ffff::` to avoid false positives on the check

sethvincent:
It looks like we're now good to merge this.

tomasciccola:
nop, I need to address this yet ->

> Ooops, yeah, fixed that.
Additionally, socket.remoteAddress returns an IPv6 with the IPv4 emebed (is it always like that or does it depend on some flag?); so I need to strip the :ffff:: to avoid false positives on the check

> This is probably not a good idea. By default `server.listen()` is [listening on the unspecified IPv6 (`::`) address](https://nodejs.org/dist/latest-v18.x/docs/api/net.html#serverlistenport-host-backlog-callback), then your OS, if not using IPv6, is [mapping IPv4 to IPv6](https://en.wikipedia.org/wiki/IPv6_address#Representation). IPv6 does not have any private addresses, so doing this regex will actually make some IPv6 addresses appear private, which is not what we want.
> 
> I think the solution to this is to remove the string.replace() here and to ensure that when we listen, we listen only on IPv4, e.g. with `net.listen(0, '0.0.0.0')`

tomasciccola:
I wrapped `this.#tcp.listen` in a promise and now it works and only listens to IPv4. 
For some reason `util.promisify` provided by node didn't work. 
The idea is that any fn that has a callback as a last parameter can be wrapped like so => 
```js
const = promisify(this.#tcp?.listen)
await listen(0,'0.0.0.0')
```
Another issue is that in a promise `resolve` is typed as `(value:any) => void` while `server.listen` expects a `() => void`.
Is there any way to 'cast' a callback?

Additionally, the `server.listen` callback can't error so I'm not handling a rejection; I don't know if this is correct.
Otherwise I think this is ready to merge

gmaclennan:
With this last commit, you would still get an error because server.address() could be null (if the listen() fails)

PR #112, opened 2023-06-06
Deterministically de-dupe mdns connections

### LINKED ISSUE #66: Deterministically de-dupe mdns connections
The [existing code to de-dupe connections](https://github.com/digidem/mapeo-core-next/blob/f296696881863d74cc23aca0cf694840038edb5b/lib/discovery/index.js#L263) can result in a race condition where peers do not agree which one of a duplicated connection they should close. We should do this deterministically, probably borrowing from [how hyperswarm does it](https://github.com/holepunchto/hyperswarm/blob/master/index.js#L239).

tomasciccola:
This should close #66 
This is done comparing the remote and local publicKeys using `b4a.compare` that sorts the keys.
On this branch there's also other changes, mainly:
* close mdns connection if not from a private ip (there's also [this PR](https://github.com/digidem/mapeo-core-next/pull/109) related to this, which does it differently...)
*  destroy connections instead of closing them (following hypercore implementation)
* setting event handlers for errors as noop
* replace throwing errors that can be handled from outside (like `socket` errors) with logging.
* PeerInfo had the host and port binded to local host and port, now they are binded to remote host and port (also, don't know why we are tracking the port...) @sethvincent does this change make sense??


gmaclennan:
@tomasciccola I'm concerned about merging work like this without any tests. Can we add tests in a follow-up PR? Also at a quick glance the `private-ip` package looks to check IPs in a very inefficient way. It may not be a huge bottleneck, but [bogon](https://www.npmjs.com/package/bogon) should be a lot more efficient, and has a similar API

lib/discovery/index.js
```diff
@@ -233,64 +235,84 @@ export class Discovery extends TypedEmitter {
...
+//       socket.address()
+//     )
 
-    if (!socketAddress.address) {
-      throw new Error('Socket not connected')
+    if (!socket.remoteAddress) {
+      console.error('Socket not connected')
+      return
     }
+    if (!isIpPrivate(socket.remoteAddress)) socket.destroy()
```
gmaclennan, line 253: @tomasciccola this should return here. Also there are no error handlers on socket at this point so this will crash with an unhandled error

gmaclennan:
> the `private-ip` package looks to check IPs in a very inefficient way.

I ran a quick benchmark and `private-ip` is >100 times slower than bogon for evaluating non-private IPs. Still negligible at 0.01ms per check, but would be better to avoid having that code in our codebase and stick to bogon.

gmaclennan:
> @tomasciccola I'm concerned about merging work like this without any tests.

Obviously the check on remote ips is hard to test, but it would be good to simulate some situations where we can test the dedupe code. Running coverage checks can ensure that the dedupe code paths are being tested.

tomasciccola:
You're right, I'll open an issue to add tests to the dedup code. Additionally, change private-ip to bogon

PR #117, opened 2023-07-06
feat: serve blobs over http

### LINKED ISSUE #115: Provide photos in a way that can be consumed by the front-end
## Description

We store photos in the blob store, and we need to display these photos in the frontend. In Mapeo 5 photos are just stored as files, and to display them we pass a file path to the front end, which sets the file path as the src for an image. This is quick and efficient. We don't have a file path for images stored in the blob store.

Passing the images around as blobs is not really an option. React Native (and the bridge we use) does not support buffers, so they would need to be encoded to Base64 or equivalent. This has big memory and performance overheads. In addition, in React Native images are actually drawn on the UI thread, not the thread that the JS code is running, so this blob needs to go across another bridge (the React Native -> mobile Native world) to be displayed. This will be really expensive for performance.

A couple of options:

1. Serve the images via a local http server
2. Use [MirrorDrive](https://docs.holepunch.to/helpers/mirrordrive) to copy the hyperdrive to disk
3. Copy files to disk on-demand to cache storage and cleanup after, e.g. add a `mapeo.getBlobFilePath(blobId)` method that will copy to cache storage, and cleanup if not accessed for a while.

I think (2) will use up too much disk space and is not really an option due to space restrictions on mobile. (3) will use less disk space, but it can still be an issue. (1) is clunky, but it might end up being less work than trying to deal with the caching of (3).

We also need to bear in mind the upcoming audio and video file storage, which will have similar requirements.

## Tasks

- [ ] Decide which approach is going to work best
- [ ] Implement the API for this

gmaclennan:
Fixes #115 

- [x] Add BlobServer class
- [x] Write tests


gmaclennan:
This initial implementation is done, but I have some questions:

1. Do we need to provide a mimetype header? If so, what is the best way to do it?
    a. Store the mimetype as [Hyperdrive metadata](https://github.com/holepunchto/hyperdrive#const-ws--drivecreatewritestreampath-options) and make two reads on the server (get the entry, then get the stream)
    b. Store the mimetype as metadata, and in the blobStore.createReadstream function, emit the mimetype as an event when read. Would need some magic to start the stream to get the event before piping to the response stream (because headers need to be set before piping the response).
    c. Guess the mimetype from the [magic bytes](https://www.npmjs.com/package/magic-bytes.js). Can get these bytes with [buffer-peek-stream](https://www.npmjs.com/package/buffer-peek-stream) before piping to the response stream.
2. Maybe just use Fastify? Although relatively simple, there a a few things that will require boilerplate to get working without fastify:
    a. Error handling to return 404 for missing routes, vs 400 bad request, 500 server error etc.
    b. Logging requests
    c. Adding cache headers / etag e.g. https://github.com/fastify/fastify-caching
    d. Promisified server.listen (already in code, copied from Fastify, more complex than first seems).
I have a feeling that over time we will end up adding these ourselves when we would get them from Fastify.
It does create the issue of potentially multiple Fastify versions. I think publishing as Fastify plugins can help this.
@achou11 thoughts?

gmaclennan:
Discussed with @achou11 in-person.

> Do we need to provide a mimetype header? If so, what is the best way to do it?

Yes. Store in hyperdrive metadata. Read metadata as `drive.entry()`, write mimetype header, then return the stream from `(await drive.getBlobs()).createReadStream(entry.value.blob)`

> Maybe just use Fastify?

Yes, although avoiding another fastify instance might be preferable, the simplicity gains of writing a raw http server are quickly lost as we add error handling and tackle edge-cases. In the future we can convert everything to fastify plugins and set fastify as a peer dependency so that we do not end up with multiple fastify instances.

gmaclennan (commented):
Implementation all looks good. Just needs some more tests I think:

- Extract logic to populate the server with fixtures, to make tests easier to read (and clarify what they are testing)
- Test for valid URL but non-existent blob name and/or variant, should return 404
- Test for trying to read blob that exists but has not been replicated, should also return 404

lib/blob-server/fastify-plugin.js
```diff
@@ -0,0 +1,120 @@
...
+  fastify.get(
+    '/:driveId/:type/:variant/:name',
+    { schema: { params: PARAMS_JSON_SCHEMA } },
+    /**
+     * @param {import('fastify').FastifyRequest<{ Params: ParsedParams }>} request
+     */
+    async (request, reply) => {
+      const { driveId, type, variant, name } = request.params
+
+      const blobId = constructBlobId(driveId, type, variant, name)
```
achou11, line 63: wasn't keen on trying to figure out how to fully reflect the discriminate union of the BlobVariant in JSON Schema format, so using `contructBlobId` to assert and guarantee that the provided combination of `type` and `variant` are valid at runtime

lib/blob-store/index.js
```diff
@@ -9,12 +9,14 @@ import { LiveDownload } from './live-download.js'
...
 // value = array of blob variants supported for that type
-export const SUPPORTED_BLOB_VARIANTS = /** @type {const} */ ({
+const SUPPORTED_BLOB_VARIANTS = /** @type {const} */ ({
   photo: ['original', 'preview', 'thumbnail'],
   audio: ['original'],
-  video: ['original']
+  video: ['original'],
 })
 
+export { SUPPORTED_BLOB_VARIANTS }
```
achou11, line 18: for posterity, needed to export on a separate line because the jsdoc type cast syntax doesn't play nicely with the export keyword

lib/blob-store/index.js
```diff
@@ -89,34 +100,70 @@ export class BlobStore {
...
+  async getEntryBlob(driveId, entry, { length } = {}) {
+    const drive = this.getDrive(driveId)
+    const blobs = await drive.getBlobs()
+
+    if (!blobs)
+      throw new Error(
+        'Hyperblobs instance not found for drive ' + driveId.slice(0, 7)
+      )
+
+    return blobs.get(entry.value.blob, { wait: false, start: 0, length })
```
achou11, line 155: okay to be opinionated about these options, at least initially?

lib/blob-store/index.js
```diff
@@ -89,34 +100,70 @@ export class BlobStore {
...
+  }
+
   /**
    *
    * @param {Omit<BlobId, 'driveId'>} blobId
    * @param {Buffer} blob
    * @param {object} [options]
-   * @param {any} [options.metadata] Metadata to store with the blob
+   * @param {{[key: string]: any, mime: string}} [options.metadata] Metadata to store with the blob
```
achou11, line 163: okay with using `mime` as the metadata field that we store in the drive entry?

package.json
```diff
@@ -10,7 +10,9 @@
     "./datastore": "./lib/datastore/index.js",
     "./discovery": "./lib/discovery/index.js",
     "./datatype": "./lib/datatype/index.js",
-    "./sqlite": "./lib/sqlite.js"
+    "./sqlite": "./lib/sqlite.js",
+    "./blob-store": "./lib/blob-store/index.js",
+    "./blob-server": "./lib/blob-server/index.js"
```
achou11, line 15: should we also have an export for the fastify plugin itself? (located at `./lib/blob-server/fastify-plugin.js`)

achou11:
Todo: update implementation to support multiproject i.e.

- add `:projectId` param to route registration
- pass `getBlobStore(projectId)` to fastify in order to get the project-specific blobstore

lib/blob-store/index.js
gmaclennan: Yes, that's fine

lib/blob-server/fastify-plugin.js
gmaclennan: That's valid. I switched to a typescript type guard for slightly stricter typing.

lib/blob-store/index.js
gmaclennan: Switched to mimeType, personal pref

package.json
gmaclennan: At this stage I'm not sure we need any of these exports, so ok to leave as-is for now.

gmaclennan (commented):
Made some fixes, good to merge I think

gmaclennan:
Added some tests to get 100% coverage on `lib/blob-server/**`

Issue #139, opened 2023-07-31
Internal architecture: adjust DataStore & DataType responsibilities


gmaclennan:
## Description

As the codebase has evolved along with the proposed public API, some changes to the responsibilities of the internal modules `DataStore` and `DataType` can simplify integration and reduce dependencies between different parts of the codebase.

Proposed changes:

### `DataStore`

A`DataStore` should be an abstraction over a particular "Mapeo Store", e.g. a collection of hypercores that store a particular datatype. Responsibilities:

- Write a document to the write core of the store.
- Read a document based on versionId.
- Manage read indexing and return current indexing state.
- Handle decoding/encoding of data

With this we can isolate the boundary with hypercore / core store to this module (apart from within the blob store), and largely isolate indexing: The write method can resolve when the document is indexed.

```ts
const store = new DataStore({
  coreManager,
  namespace: 'auth' | 'data',
  // DataStore will consider a block "indexed" once this function resolves.
  indexDocs: (blocks: Buffer[]) => Promise<void>
})

// Guarantee that `doc` is indexed once this resolves
store.write(doc: Exclude<MapeoDoc, 'versionId'>): Promise<MapeoDoc>

store.read(versionId: string): Promise<MapeoDoc>

store.getIndexState() // return whether is currently indexing, and estimate of indexing remaining to be done

store.on('index-state') // same as above.
```

### `DataType`

Implements the [`MapeoDoc` methods from the public API](https://github.com/digidem/mapeo-core-docs/blob/f6a5498212567642423dabde06bd9c73518077af/client-docs/README.md#mapeodoc). Needs a DataStore instance, probably also needs a way to get permissions. One dataType instance for each MapeoDoc type, but not all will be exposed in the public API. Needs a way to define what data types it will handle, possibly a Drizzle schema.

```ts
const dataType = new DataType({
  dataStore,
  schema,
  getPermissions // function to return current permissions (read/write, own vs. others)
})

dataType.create()
dataType.getByDocId()
dataType.getByVersionId()
dataType.getMany()
dataType.update()
dataType.delete()
```

### `IndexWriter`

One instance that does all writing to the indexes. Only place where SQLite is written to (all other places are read-only SQLite). Just exports one method `batch()` which writes one or more blocks to the index.

- Decodes blocks from buffers
- Ignores invalid blocks
- Migrates previous schema versions if required
- Writes each data type to a separate sqlite table
- `batch()` resolves when documents are written to the index
[...]

PR #140, opened 2023-08-01
feat: generate SQL table schema from JSON Schemas [1/n]

gmaclennan:
Generate drizzle-orm SQLite table schemas from Mapeo types & JSONSchemas.
(Currently uses unpublished version of `@mapeo/schema` that exports JSONSchemas with expanded $refs)


gmaclennan:
Now finished and ready for review, but awaiting publishing of new version of `@mapeo/schema`.

Will be stacking other PRs on this, working on #139

lib/schema/schema-to-drizzle.js
```diff
@@ -0,0 +1,302 @@
...
+/**
+Convert optional properties to nullable
+@template {object} T
+@typedef {{ [K in keyof T]-?: undefined extends T[K] ? T[K] | null : T[K] }} OptionalToNull
+ */
+
+/**
+Convert a readonly array/object to writeable
+@template T
+@typedef {{ -readonly [P in keyof T]: T[P] }} Writable
```
achou11, line 12: out of curiosity, what are the hyphens in the types?

lib/schema/schema-to-drizzle.js
```diff
@@ -0,0 +1,302 @@
...
+
+/**
+ * Check if a value is an array of strings of length at least one
+ * @param {any} value
+ * @returns {value is [string, ...string[]]}
+ */
+function isStringArray(value) {
+  return (
+    isArray(value) &&
+    value.every((v) => typeof v === 'string') &&
```
achou11, line 299: was going to suggest prematurely optimizing this using `Array.some`, but the length `value` will be small here like 99% of the time right?

achou11 (approved):
overall looks good! curious about what else to add to the tests, if anything

lib/schema/schema-to-drizzle.js
gmaclennan: They are called "[mapping modifiers](https://www.typescriptlang.org/docs/handbook/2/mapped-types.html#mapping-modifiers)" and can be used to remove readonly from a type.

lib/schema/schema-to-drizzle.js
gmaclennan: Yes, will be small e.g. max will be the number of enum values. Also this only runs once on app startup so it's not something that needs to be fast.

gmaclennan:
> overall looks good! curious about what else to add to the tests, if anything

I think more type tests would be good, but probably make more sense to be end-to-end tests rather than unit-testing just this function.

lib/schema/schema-to-drizzle.js
achou11: oh wow, very cool!

gmaclennan:
Merged in #149

PR #141, opened 2023-08-02
feat: updated DataStore [2/n]

gmaclennan:
From #139:

A `DataStore` should be an abstraction over a particular "Mapeo Store", e.g. a collection of hypercores that store a particular datatype. Responsibilities:

- Write a document to the write core of the store.
- Read a document based on versionId.
- Manage read indexing and return current indexing state.
- Handle decoding/encoding of data

With this we can isolate the boundary with hypercore / core store to this module (apart from within the blob store), and largely isolate indexing: The write method can resolve when the document is indexed.

```ts
const store = new DataStore({
  coreManager,
  namespace: 'auth' | 'data',
  // DataStore will consider a block "indexed" once this function resolves.
  indexDocs: (blocks: Buffer[]) => Promise<void>
})

// Guarantee that `doc` is indexed once this resolves
store.write(doc: Exclude<MapeoDoc, 'versionId'>): Promise<MapeoDoc>

store.read(versionId: string): Promise<MapeoDoc>

store.getIndexState() // return whether is currently indexing, and estimate of indexing remaining to be done

store.on('index-state') // same as above.
```

### DIFF
--- a/lib/core-manager/index.js
+++ b/lib/core-manager/index.js
@@ -10,7 +10,12 @@ import { ReplicationStateMachine } from './replication-state-machine.js'
 
 // WARNING: Changing these will break things for existing apps, since namespaces
 // are used for key derivation
-const NAMESPACES = /** @type {const} */ (['auth', 'data', 'blobIndex', 'blob'])
+export const NAMESPACES = /** @type {const} */ ([
+  'auth',
+  'data',
+  'blobIndex',
+  'blob'
+])
 // WARNING: If changed once in production then we need a migration strategy
 const TABLE = 'cores'
 const CREATE_SQL = `CREATE TABLE IF NOT EXISTS ${TABLE} (
@@ -18,10 +23,11 @@ const CREATE_SQL = `CREATE TABLE IF NOT EXISTS ${TABLE} (
   namespace TEXT NOT NULL
 )`
 
+/** @typedef {import('hypercore')<'binary', Buffer>} Core */
 /** @typedef {(typeof NAMESPACES)[number]} Namespace */
-/** @typedef {{ core: import('hypercore').default, key: Buffer, namespace: Namespace }} CoreRecord */
+/** @typedef {{ core: Core, key: Buffer, namespace: Namespace }} CoreRecord */
 /** @typedef {import('streamx').Duplex} DuplexStream */
-/** @typedef {{ rsm: ReplicationStateMachine, stream: DuplexStream, cores: Set<Hypercore> }} ReplicationRecord */
+/** @typedef {{ rsm: ReplicationStateMachine, stream: DuplexStream, cores: Set<Core> }} ReplicationRecord */
 /**
  * @typedef {Object} Events
  * @property {(coreRecord: CoreRecord) => void} add-core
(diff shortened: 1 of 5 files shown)


gmaclennan:
Now largely complete, will be best to review once all items in #139 are done, because this might still change.

lib/core-manager/index.js
```diff
@@ -285,6 +291,7 @@ export class CoreManager extends TypedEmitter {
         'Passed an existing protocol stream to coreManager.replicate(). Other corestores and core managers replicated to this stream will no longer automatically inject shared cores into the stream'
       )
     }
+    // @ts-ignore - too complex to type right now
```
achou11, line 294: I generally like to use `@ts-expect-error` for these kinds of these things, in case they ever do get fixed
```suggestion
    // @ts-expect-error - too complex to type right now
```

lib/datastore/data-store-new.js
```diff
@@ -0,0 +1,175 @@
...
+      batch: (entries) => this.#batch(entries, indexEntries),
+    })
+
+    // Forward events from coreIndexer
+    this.on('newListener', (eventName, listener) => {
+      if (['newListener', 'removeListener'].includes(eventName)) return
+      this.#coreIndexer.on(eventName, listener)
+    })
+    this.on('removeListener', (eventName, listener) => {
+      if (['newListener', 'removeListener'].includes(eventName)) return
```
achou11, line 69: extract these events to a constant? i.e. `const LISTENER_EVENTS = ['newListener', 'removeListener']` (you can come up with a better variable name 😄 )

lib/datastore/data-store-new.js
```diff
@@ -0,0 +1,175 @@
...
+ */
+function parseVersionId(versionId) {
+  const items = versionId.split('/')
+  if (!items[0] || !items[1]) throw new Error('Invalid versionId')
+  const coreKey = Buffer.from(items[0], 'hex')
+  if (coreKey.length !== 32) throw new Error('Invalid versionId')
+  const index = Number.parseInt(items[1])
+  if (isNaN(index)) throw new Error('Invalid versionId')
+  return { coreKey, index }
+}
```
achou11, line 175: can remove these and use the exported helpers from `@mapeo/schema` instead

tests/datastore-new.js
```diff
@@ -0,0 +1,91 @@
...
+    'data returned from write matches data returned from read'
+  )
+  t.alike(
+    indexedVersionIds,
+    [written.versionId],
+    'The indexEntries function is called with all data that is added'
+  )
+})
+
+test.solo('index events', async (t) => {
```
achou11, line 56: remove the `solo`?

achou11 (approved):
couple of minor suggestions. also need to remove a `test.solo` call to see if all the added tests pass, but otherwise seems good to me

lib/datastore/data-store-new.js
gmaclennan: I'd prefer not to because I'd like this to be explicit in the code, rather than the indirection of a constant - we'd end up naming it "newListenerAndRemoveListenerEventNames". Making is a constant might be a slight performance improvement, but this is not somewhere that is a bottleneck (events are added rarely).

gmaclennan:
Rebased onto changes in #140

gmaclennan:
Merged in #149

PR #144, opened 2023-08-05
feat: `IndexWriter` for all things indexy [3/n]

gmaclennan:
All indexing goes through this class, which just exports a single method `batch()`. Later we can move the index code into a separate thread for performance.

Depends on:

- [x] https://github.com/digidem/mapeo-sqlite-indexer/pull/11
- [x] https://github.com/digidem/mapeo-sqlite-indexer/pull/14

Makes more sense to e2e test rather than unit test, which depends on
the new DataType class, which is next up.


gmaclennan:
Merged in #149

PR #151, opened 2023-08-08
chore: upgrade Hyperdrive to 11.5.3, fix blob-store tests, fix hyperdrive blob namespace

### LINKED ISSUE #147: Update to hyperdrive@11.5.3
## Description

We have been using a fork/patch of hyperdrive in order to support passing the `wait` option to `get()`. This has now landed in https://github.com/holepunchto/hyperdrive/pull/348 and https://github.com/holepunchto/hyperdrive/pull/351.

We should update to the latest (currently v11.5.3) instead of the branch we currently use (github:holepunchto/hyperdrive#3db28f3). However the change to how the blob core is named in https://github.com/holepunchto/hyperdrive/pull/355/files will break our "[fake corestore](https://github.com/digidem/mapeo-core-next/blob/8e64abdf736b31d1a13a1d9791fd9f677118d8d2/lib/blob-store/index.js#L269-L271)" so will need a fix. Change in behaviour to [throwing if not found](https://github.com/holepunchto/hyperdrive/pull/352/files) might also require some changes - I think we have a `null` check in there right now.

## Tasks

- [ ] Update to hypercore@11.5.3
- [ ] Fix errors and ensure blob-store and blob-server tests pass

sethvincent:
This pr upgrades the hyperdrive package to latest and fixes a couple small errors:
- blobs are now always [prefixed by the internal hyperbee's core id](https://github.com/holepunchto/hyperdrive/pull/355/files)
- A few tests needed to check for a thrown exception rather than a null return value

Closes #147

### DIFF
--- a/lib/blob-store/index.js
+++ b/lib/blob-store/index.js
@@ -266,7 +266,7 @@ class PretendCorestore {
       )
     } else if (opts.name === 'db') {
       return this.#coreManager.getWriterCore('blobIndex').core
-    } else if (opts.name === 'blobs') {
+    } else if (opts.name.includes('blobs')) {
       return this.#coreManager.getWriterCore('blob').core
     } else {
       throw new Error(
--- a/package-lock.json
+++ b/package-lock.json
@@ -22,7 +22,7 @@
         "fastify-plugin": "^4.5.0",
         "hypercore": "^10.9.0",
         "hypercore-crypto": "^3.3.1",
-        "hyperdrive": "github:holepunchto/hyperdrive#3db28f3",
+        "hyperdrive": "^11.5.3",
         "hyperswarm": "^4.4.1",
         "magic-bytes.js": "^1.0.14",
         "mapeo-schema": "github:digidem/mapeo-schema#protobufsTypescript",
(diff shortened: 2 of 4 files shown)


lib/blob-store/index.js
```diff
@@ -266,7 +266,7 @@ class PretendCorestore {
       )
     } else if (opts.name === 'db') {
       return this.#coreManager.getWriterCore('blobIndex').core
-    } else if (opts.name === 'blobs') {
+    } else if (opts.name.includes('blobs')) {
```
achou11, line 269: this seems like a reasonable change but not sure how this works in practice. wanted to highlight in order to get clarity

achou11 (approved):
one comment that would be helpful to get some clarification on, but otherwise lgtm

lib/blob-store/index.js
gmaclennan: We don't actually name any cores, so this will be fine. It's just a way of making sure we return the right core when hyperdrive asks for it. For existing hyperdrives it always reads the blob core key from the blob index anyway. In an ideal world the hyperdrive constructor would accept the blob core and the hyperbee core, but this is a necessary workaround for now.

gmaclennan (approved):
Looks good to me. Can you create a follow up issue to resolve the todo comments in the code around wait options?

PR #154, opened 2023-08-09
chore: code cleanup and re-org

gmaclennan:
- Remove code replaced by new DataStore, DataType and IndexWriter
  implementations
- Skip failing authStore tests for now - get tests running and passing again (pending updates to AuthStore for new Mapeo Schema)
- Move all code to `src` folder to prepare for compiling typescript
  declaration files for package publishing
- Update internal READMEs
- Format everything with prettier
- Add missing deps
- Fix protobuf build script

Why move everything into `src`? For publishing we need to compile Typescript declarations (`.d.ts`). It's difficult to manage outputting these alongside the source files, it's easier to manage if all declarations go in a `dist` folder. However if there are JS files that need type-checked in the top-level project folder, then Typescript config can be a pain. Typescript is easiest to configure when all code that needs to be compiled is in a single folder, for which `src` makes the most sense.


src/core-manager/messages.js
achou11, line 1: is this a generated file? if so, is there a way to add a comment here to indicate that?

tests/authstore.js
achou11, line 1: would `test.todo` be more appropriate for these or does it not really matter?

achou11 (approved):
Couple of minor questions but otherewise lgtm!

src/core-manager/messages.js
gmaclennan: Yes. Not sure. Let's address in a separate issue (this PR doesn't change this file)

tests/authstore.js
gmaclennan: I'll just remember for now, will be working on this next week.

PR #173, opened 2023-08-14
fix: properly generate column names in projectKeys table

achou11:
Fixes a bug introduced in #169, as I used a `for...in` instead of a `for...of` loop...

Also creates the drizzle files becaues I forgot to do that in the other PR. My understanding is that it should be okay to overwrite the original migration files, as opposed to creating a new migration

Referencing #165 for linking purposes

### DIFF
--- a/drizzle/client/0000_unknown_umar.sql
+++ b/drizzle/client/0000_unknown_umar.sql
@@ -2,6 +2,16 @@ CREATE TABLE `project_backlink` (
 	`versionId` text PRIMARY KEY NOT NULL
 );
 --> statement-breakpoint
+CREATE TABLE `projectKeys` (
+	`projectId` text PRIMARY KEY NOT NULL,
+	`projectSecretKey` blob,
+	`0EncryptionKey` blob,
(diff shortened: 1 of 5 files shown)


drizzle/client/0000_unknown_umar.sql
```diff
@@ -2,6 +2,16 @@ CREATE TABLE `project_backlink` (
...
 );
 --> statement-breakpoint
+CREATE TABLE `projectKeys` (
+	`projectId` text PRIMARY KEY NOT NULL,
+	`projectSecretKey` blob,
+	`0EncryptionKey` blob,
+	`1EncryptionKey` blob,
+	`2EncryptionKey` blob,
+	`3EncryptionKey` blob,
+	`randomEncryptionKey` blob
```
achou11, line 12: Major bug exposed 🙃 fixing

achou11: fixed via 8c53afe

gmaclennan (approved):
oops I should have caught this in review

PR #174, opened 2023-08-15
feat: Add config store and preset & field types

gmaclennan:
~~Currently depends on https://github.com/digidem/mapeo-schema/pull/117 and a new version of @mapeo/schema being published.~~

This adds two new data types: preset and field. They are stored in a new namespace called "config". The reason for keeping these data types in a separate namespace is to allow sync of only the config prior to syncing the rest of the data. We're not using the auth namespace because Mapeo Cloud needs the encryption key for that namespace, and we want to minimise the data that a server can see as unencrypted.

### DIFF
--- a/package.json
+++ b/package.json
@@ -82,7 +82,7 @@
     "@fastify/type-provider-typebox": "^3.3.0",
     "@hyperswarm/secret-stream": "^6.1.2",
     "@mapeo/crypto": "^1.0.0-alpha.4",
-    "@mapeo/schema": "file:../mapeo-schema/mapeo-schema-3.0.0-next.5.tgz",
+    "@mapeo/schema": "^3.0.0-next.6",
     "@mapeo/sqlite-indexer": "^1.0.0-alpha.5",
     "@sinclair/typebox": "^0.29.6",
     "b4a": "^1.6.3",
--- a/package.json
+++ b/package.json
@@ -82,7 +82,7 @@
     "@fastify/type-provider-typebox": "^3.3.0",
     "@hyperswarm/secret-stream": "^6.1.2",
     "@mapeo/crypto": "^1.0.0-alpha.4",
-    "@mapeo/schema": "^3.0.0-next.5",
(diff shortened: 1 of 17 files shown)


package.json
```diff
@@ -82,7 +82,7 @@
     "@fastify/type-provider-typebox": "^3.3.0",
     "@hyperswarm/secret-stream": "^6.1.2",
     "@mapeo/crypto": "^1.0.0-alpha.4",
-    "@mapeo/schema": "^3.0.0-next.5",
+    "@mapeo/schema": "file:../mapeo-schema/mapeo-schema-3.0.0-next.5.tgz",
```
achou11, line 85: Is this change intentional?

test-e2e/project-crud.js
```diff
@@ -3,54 +3,116 @@ import { randomBytes } from 'crypto'
...
-    }
-  })
-  for (const value of obs) {
-    await project.observation.create(value)
+test('CRUD operations', async (t) => {
+  for (const value of fixtures) {
+    const { schemaName } = value
+    t.test(`create and read ${schemaName}`, async (t) => {
+      const project = await createProject()
+      // @ts-ignore - TS can't figure this out, but we're not testing types here to ok to ignore
```
achou11, line 68:
```suggestion
      // @ts-ignore - TS can't figure this out, but we're not testing types here so ok to ignore
```

test-e2e/project-crud.js
```diff
@@ -82,3 +144,13 @@ function createProject({
 function stripUndef(obj) {
   return JSON.parse(JSON.stringify(obj))
 }
+
+/**
+ *
+ * @param {number} value
+ * @param {number} decimalPlaces
+ * @returns
```
achou11, line 152:
```suggestion

```

achou11 (changes requested):
Requesting change only because of the `package.json` change, otherwise lgtm!

package.json
gmaclennan: Yeah, until we merge and publish @mapeo/schema - a reminder not to merge this yet - tests won't pass

achou11 (approved):

PR #176, opened 2023-08-15
chore: update drizzle files for client db to add configEncryptionKey column

achou11:
Didn't catch this when reviewing #174 

although not a high priority for now, would be ideal to set up some sort of check to not forget to do this


achou11:
Closing since no longer relevant with #178

PR #177, opened 2023-08-15
Implement MdnsDiscovery

tomasciccola:


tomasciccola:
On this last commit I'm having this issue where on an initiated connection `socket.remoteAddress` (or `socket.address`) can be undefined, I don't know why but I suspect that is related to the reusing of `this.#handleConnection` listener for the callback of the started tcp server and the connection to peers adverticing mapeo. 
The thing is that if I uncomment line 78 ( `if(!socket.remoteAddress) return`) then peers won't be able to find each other. I started working on this at the end of the day so the culprit is probably my brain...

tomasciccola:
Heh, had a really silly bug passing the connection handler to the tcp client. Still getting some duplicated connections though...

src/discovery/mdns.js
```diff
@@ -0,0 +1,119 @@
...
+   */
+  async #handleConnection(isInitiator, socket) {
+    if(!socket.remoteAddress) return
+
+    const remoteAddress = socket.remoteAddress
+    if (this.#connections.has(remoteAddress)) {
+      socket.destroy()
+      return
+    }
+
```
gmaclennan, line 87: We should add this to the connections set immediately

```suggestion
this.#connections.add(remoteAddress)
```

src/discovery/mdns.js
```diff
@@ -0,0 +1,119 @@
...
+    const secretStream = new NoiseSecretStream(isInitiator, socket, {
+      keyPair: this.#identityKeypair,
+    })
+
+    secretStream.on('close', () => {
+      this.#connections.delete(remoteAddress)
```
gmaclennan, line 93: Need to test what happens when:
1. secretStream throws an error
2. socket throws an error before secretStream connects
Regardless I believe we need to attach an error listener to secretStream - Node has a funny thing where you need to attach error listeners or it will throw an uncaught error and crash.

```suggestion
      this.#socketConnections.delete(remoteAddress)
      this.#noiseConnections.delete(secretStream.remotePublicKey)
```

src/discovery/mdns.js
```diff
@@ -0,0 +1,119 @@
...
+ * @property {Object<string, any>} txt - TXT records of the service
+ * @property {string[]} [addresses] - addresses of the service
+ */
+
+
+export class MdnsDiscovery extends TypedEmitter {
+  #identityKeypair
+  #server
+  /** @type {Set<string>} */
+  #connections = new Set()
```
gmaclennan, line 27:
```suggestion
  #socketConnections = new Set()
  #noiseConnections = new Set()
```
We're actually doing two stages of dedupe - not sure the second is strictly necessary, but good to have just in case.

src/discovery/mdns.js
```diff
@@ -0,0 +1,119 @@
...
+      keyPair: this.#identityKeypair,
+    })
+
+    secretStream.on('close', () => {
+      this.#connections.delete(remoteAddress)
+    })
+
+
+    secretStream.on('connect', async () => {
+      // deterministically choose the connection to keep,
```
gmaclennan, line 98:
```suggestion
    secretStream.on('connect', async () => {
      const isDuplicate = this.#noiseConnections.has(secretStream.remotePublicKey)
      if (isDuplicate) {
        // check whether to keep, if not then destroy the socket here and return
      }
      this.#noiseConnections.add(secretStream.remotePublicKey)
      this.emit('connection', secretStream)
```

src/discovery/mdns.js
```diff
@@ -0,0 +1,119 @@
...
+      // by ordering the {remote,local}public keys
+      const keep = !secretStream.isInitiator ||
+        b4a.compare(secretStream.publicKey, secretStream.remotePublicKey) > 0
+      if(keep){
```
gmaclennan, line 102: We don't want to choose to keep based on `isInitiator` - both sides of the connection will make the same determination.

The reason for this code Hyperswarm is to check for duplicate outgoing connections, e.g. a device A tries to connect to device B twice. This can happen if device B starts advertising, so device A connects, then device B stops advertising then starts again, without closing the original connection. Then device A will see the advertising re-appear and re-try to connect.

To solve this I think we actually need to change `this.#noiseConnections` to a `Map<remotePublicKey, noiseStream>`. Then within `secretStream.on('connect')` we can check:

1. is there a previous connection where we were the initiator?
2. if we were the initiator of this new connection, then drop the new connection (e.g. keep the existing one)

src/discovery/mdns.js
```diff
@@ -0,0 +1,119 @@
+import { TypedEmitter } from 'tiny-typed-emitter'
+import net from 'node:net'
+import NoiseSecretStream from '@hyperswarm/secret-stream'
+import { once } from 'node:events'
+import dnssd from 'dnssd'
```
gmaclennan, line 5: This is no longer maintained. We have been using the fork [@gravitysoftware/dnssd](https://www.npmjs.com/package/@gravitysoftware/dnssd). Seth had written [some type declarations](https://github.com/digidem/multicast-service-discovery/blob/main/types/gravitysoftware__dnssd/index.d.ts).

There is also https://github.com/homebridge/ciao, but there seem to be some outstanding issues with that and maintenance seems to be focussed just on supporting the Homebridge app e.g. see https://github.com/SignalK/signalk-server/issues/1353.

There is also https://www.npmjs.com/package/bonjour-service but it doesn't seem to suggest any RFC compliance.

I think maybe @gravitysoftware/dnssd might be the most reliable option.

tests/discovery.js
```diff
@@ -1,439 +1,461 @@
...
-
-  const discover2 = new Discovery({
-    identityKeyPair: identity2.identityKeyPair,
-    mdns: true,
-    dht: false,
+  const mdnsDiscovery1 = new MdnsDiscovery({identityKeypair:identityKeypair1})
+  mdnsDiscovery1.on('connection', (stream) => {
+    console.log('found peer 2')
   })
+  await mdnsDiscovery1.start()
```
gmaclennan, line 22: call `start()` on both discovery instances at the same time (e.g. don't await the first) so that we can catch potential duplicate connections.

gmaclennan (commented):
Great work on this - this is tricky stuff to get right. Left a bunch of comments that hopefully get this working - I can be available to work together on this today if needed.

gmaclennan (commented):
Great work on this - this is tricky stuff to get right. Left a bunch of comments that hopefully get this working - I can be available to work together on this today if needed.

src/discovery/mdns.js
```diff
@@ -0,0 +1,119 @@
...
+    await this.#advertiser.start()
+
+    // find all peers adverticing Mapeo
+    this.#browser = await dnssd
+      .Browser(dnssd.tcp(SERVICE_NAME))
+      .on('serviceUp',
+        /** @param {Service} service */
+        (service) => {
+          const socket = net.connect(
+            service.port,
```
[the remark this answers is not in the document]
gmaclennan: I understand from the typing that service.addresses can be undefined, or an empty array. I'm not sure what circumstances might result in that, maybe when the mdns service cannot resolve an IP address. So what I mean by "does not have an address we can use" is that the browser returns addresses as undefined or an empty array.

tomasciccola:
I've been noticing instability in terms of peers finding each other. With no changes to the code it seems that sometimes two peers find each other, and sometimes they don't... 
I don't know what to make up of it, but maybe I have a bug from the beginning that I'm not seeing...

gmaclennan:
Try logging the initial dedupe based on remoteaddress. I have a feeling that there is a race condition here where both clients can end up closing both sockets. It might be better to remove this check altogether and just use the dedupe on the noise stream.

tomasciccola:
Missing for this to be merged are more tests (testing with more peers, and probably going to look to test that exist on `main` now an add them)

src/discovery/mdns.js
```diff
@@ -0,0 +1,201 @@
...
+    log(
+      `${isInitiator ? 'outgoing' : 'incoming'} connection ${
+        isInitiator ? 'to' : 'from'
+      } ${socket.remotePort}`
+    )
+    if (!socket.remoteAddress) return
+    if (!isIpPrivate(socket.remoteAddress)) return
+
+    const { remoteAddress } = socket
+    this.#socketConnections.add(remoteAddress)
```
gmaclennan, line 114: I think we can remove this now?

src/discovery/mdns.js
```diff
@@ -0,0 +1,201 @@
...
+  stop() {
+    const port = this.#server.address()?.port
+    this.#browser.removeAllListeners('serviceUp')
+    this.#browser.stop()
+    this.#advertiser.stop(true)
+    // eslint-disable-next-line no-unused-vars
+    for (const [_, socket] of this.#noiseConnections) {
+      socket.destroy()
+    }
+    this.#server.close() // wait for close
```
gmaclennan, line 180: Should wait for close

gmaclennan (changes requested):
The tests still need some work.

What we need to test is:

1. Every device will find every other device and establish a connection
2. There is no more than one connection between each device

We want to test this in a scenario that is as close as we can get to how the app will be used.

The current multi-connection test waits for 10 connection events and then stops, and doesn't do any checks about whether all clients have connected or whether there are any duplicate connections. 10 connection events = 5 connections. For 10 peers to connect to each other we need n(n-1)/2 connections, or n(n - 1) connection events.

In order to test for duplicate connections there is no great way to do this other than just "wait" for a certain period of time. I think 5 seconds should be enough.

To simulate a real-life situation, we should not start all the discoveries at the same time - although that is worth testing too, because simultaneous start can catch certain race conditions.

One way of doing all this:

1. Start each discovery at a random time in the first 2 seconds of the test
2. On each connection event push the `[localPublicKey, remotePublicKey]` tuple to an array.
3. Wait until 5 seconds total have passed.
4. Check the array of connection events has `nPeers * (nPeers - 1)` entries
5. Check that each peer is connected to every other peer e.g.
```
for (publicKey of peers) {
   const peerConnections = allConnections.filter(([localKey, remoteKey]) => localKey === publicKey).map(([ , remoteKey]) => remoteKey)
   // Check peerConnections is the same as peers.filter(peerKey => peerKey !== publicKey)
}
```

If you need help with this I can work on this with you.

I don't think the hypercore replication test is necessary. We will test for that in the replication code - it's not testing anything in the MdnsDiscovery class: it's testing whether you can replicate a hypercore over a noisestream.

tomasciccola:
Things that maybe can be improved or checked upon: 
* IPv6 handling: [we are skipping IPv6 addresses](https://github.com/digidem/mapeo-core-next/blob/c9e746d0769f515d61b2badb4f4e64502e32ece9/src/discovery/mdns.js#L83)  and this can lead to tests failing. We can disabling the resolver of `dnssd.Browser` and setting it up ourselves ([see](https://gitlab.com/gravitysoftware/dnssd.js/-/tree/master?ref_type=heads#dnssdresolveaname-options))
* Tests right now hangs the process after finishing. I put a timeout of 5 minutes in the CI workflow. But we should look at why its happening (it started to happen after adding the last tests)
* There's an unpredictable relation of `nPeers` and `timeout` when testing multiple peers. Some combinations can lead to failing tests (not every peer is connected to each other after `timeout`)
* Tests can be more DRY, maybe

gmaclennan:
I went into a rabbit hole investigating this today. After testing locally I've concluded that @gravitysoftware/dnssd is unreliably resolving the IP addresses for any service. Even when only using it as a browser, it is sometimes only resolving `127.0.0.1` for a service, or only ipv6 addresses. After reading through the source code I'm also not convinced it is fully complying with the RFCs [6763](https://www.ietf.org/rfc/rfc6763.html) and [6762](https://www.ietf.org/rfc/rfc6762.html). The problem with not fully implementing the RFC is that it can cause excessive multicast network traffic. This is a particular issue for Mapeo because we will typically have multiple devices on a network, all advertising a service and all browsing for services, and often on cheap network equipment that might not be able to handle high loads of multicast traffic.

The only mdns module I've found that appears to be fully spec compliant is [`@homebridge/ciao`](https://github.com/homebridge/ciao) which also [passes](https://github.com/homebridge/ciao/blob/master/BCT-Results-CIAO-PI-en0.txt) the [Bonjour Conformance Test](https://developer.apple.com/bonjour/) as defined and required by Apple.

Unfortunately `ciao` can only advertise a service, it does not currently support browsing to discover services on the local network. However for service advertising it seems like the best option out there.

I also found some additional modules for dns-sd:

- [`ya-dns-sd`](https://github.com/earthstar-project/dns_sd/tree/main) by [Earthstar](https://earthstar-project.org/), which looks promising. However I found several bugs:
    -  it attempts to bind the dgram socket to port 5353 without the `exclusive: false` option, which causes an error when run on a system that is already running a system service on port 5353. 
    - I cannot get the advertiser to work - I can't browse advertised services either with the built-in browse option or with avahi browse
    - The browser [emits duplicate service events when a service goes down.](https://github.com/earthstar-project/dns_sd/blob/a27d48d412ca4e55c5456f60a39cf12cc588820c/src/mdns/query.ts#L326-L338). It does appear to be following the RFC though, so this is just an API issue that could be worked around.
    It would require a patch (and a PR upstream probably) to add `exclusive: false`. It probably works ok in CI environments because they generally don't have avahi or equivalent installed, and I think their tests are mainly with Deno, which uses something different to the node gram module.
- [`bonjour-service`](https://github.com/onlxltd/bonjour-service), which also looks good. It's built on [`multicast-dns`](https://www.npmjs.com/package/multicast-dns) which is fairly battle-tested, and it has 7 million downloads a month, but it does have [fairly limited tests](https://github.com/onlxltd/bonjour-service/tree/master/test) and does not mention full RFC compliance in its redme
- [`mdns`](https://github.com/agnat/node_mdns) this is a wrapper for native mdns, e.g. avahi on linux and Apple's mdns implementation on Android and MacOS. This is the most reliable option if it is available, because [there can be some issues running multiple mDNS responders on the same machine](https://datatracker.ietf.org/doc/html/rfc6762#section-15).

I tested `ciao` and it worked well for advertising, but browsing also requires running an mDNS resolver, and having two separate resolvers running (`ciao`'s resolver, and the resolver for browsing) seems like a ticket to issues such as those described in the RFC.

I think regardless of choice it will be good to build an abstraction for the mdns discovery that we need (e.g. not the work connecting to discovered devices and establishing a noise stream, but the act of advertising and browsing). That way we can swap out the mdns library in the future without too much issue.

The other issue with any JS implementation of mDNS is how it handles device sleep and network interfaces coming and going. Particularly for mobile there are not events available on the node side for when, for example, airplane mode is turned off. This was not too much of an issue with Mapeo 5 because we only ran discovery when on the sync screen, but if we want discovery to be continuous in the background, it requires managing connections coming and going.

I think ultimately the best solution for Android is going to be to use the native implementation of dns-sd, via [network service discovery](https://developer.android.com/training/connect-devices-wirelessly/nsd). I'm not sure how to bridge between Java and Node, so we would need to interact with NSD via React Native with [`react-native-zeroconf`](https://github.com/balthazar/react-native-zeroconf). The node process would need to send a message when it starts listening for connections, with the port number. React-native-zeroconf could then start advertising and browsing, and send a message back to node when it finds a service. If we build an abstraction around advertising and browsing then we can swap in this android specific implementation when needed. I think something like this is going to be necessary on iOS anyway, and probably on Android in the future, because of rules restricting app's use of multicast discovery.

gmaclennan:
### Instance names

Instance names in dns-sd are [meant to be human-friendly names](https://www.ietf.org/rfc/rfc6763.html#section-4.1.1) for identifying devices on the network. However they must also be unique, so dns-sd implementations must check on the network to see if a name is already taken, then try a different name, normally by appending a number to the end of the name.  

The dns-sd implementation that we have been using defaults to using the hostname if the instance name is not defined. This is causing an issue with tests because the RFC 6762 says that after 15 attempts at a unique name in 10 seconds, the host must wait an additional 10 seconds. Because tests have been using the hostname, each device will be a conflict, resulting in multiple attempts, and throttling. This is likely one reason that tests are not behaving as we would like.

We are going to start asking users to name their devices, so we could use that for the service name, and using a user-friendly name like that is what the [RFC recommends](https://www.ietf.org/rfc/rfc6763.html#appendix-C). However for Mapeo that could be considered leaking of potentially sensitive information: anyone on the network with the ability to browse mdns records would be able to see the names of devices. 

I suggest that we use random identifiers that change on every restart for the instance name for mdns records. This avoids leaking any identifiable information to the local network. We do need a name, however, for displaying the list of local devices that could be invited to a project. If we send this name over the Mapeo RPC channel on first connect, then (1) the name is not visible to the casual mDNS browser, and only visible to a sophisticated hack that establishes Noise streams, and (2) a device could choose to not share it's name with other devices that are not part of its project, in security-sensitive situations.

We can do this because the list of devices that we show to the user when listing devices to invite is not the discovered mDNS services, but rather the list of devices that we have been able to establish a noise stream connection with.

### Tasks

- [ ] Use random ids as the instance name for service broadcast (can use this rather than the current txt method for de-duping)
- [ ] Send the device name over the RPC channel on first connect

gmaclennan:
Tracking work to finish up this PR in https://github.com/digidem/mapeo-core-next/issues/244. I've assigned to myself for now.

gmaclennan:
This is now blocking work that needs to be built on top of this. The current issue is failing tests when a large number of peers try to discover each other.

Suggestion: We skip the failing test, and merge this with the bug, along with a new issue to track it. That will enable us to build on top of this (it works in simple scenarios) and we can come back to the bug later, and avoids this PR getting behind work on `main` as well.

@achou11 thoughts on this approach?

achou11:
> This is now blocking work that needs to be built on top of this. The current issue is failing tests when a large number of peers try to discover each other.
> 
> Suggestion: We skip the failing test, and merge this with the bug, along with a new issue to track it. That will enable us to build on top of this (it works in simple scenarios) and we can come back to the bug later, and avoids this PR getting behind work on `main` as well.
> 
> @achou11 thoughts on this approach?

sounds good to me 👍

PR #180, opened 2023-08-16
chore: replace Datastore IndexWriter opt with batch opt

### LINKED ISSUE #179: Update DataStore to accept `batch` option instead of IndexWriter
Related to #167 

Currently the `DataStore` accepts a single `IndexWriter` instance that it [calls the `batch` function](https://github.com/digidem/mapeo-core-next/blob/9f16817b5d861cfde460337cd8155c12a46321af/src/datastore/index.js#L102) on. However, because we want to index project info differently from the other kinds of data, we need a separate `IndexWriter` that specifically handles project info indexing across projects. Instead of passing a second instance to `DataStore`, we should invert control and replace the `IndexWriter` option with a more generic `batch` function so that the caller of the DataStore contructor has more control over how entries get indexed.

achou11:
Closes #179 

Meta question: is this a (breaking?) `feat` or `chore` task??

### DIFF
--- a/src/datastore/index.js
+++ b/src/datastore/index.js
@@ -63,7 +63,7 @@ export class DataStore extends TypedEmitter {
     const cores = coreManager.getCores(namespace).map((cr) => cr.core)
     this.#coreIndexer = new MultiCoreIndexer(cores, {
       storage,
-      batch: (entries) => this.#batchEntries(entries),
+      batch: (entries) => this.#handleEntries(entries),
     })
 
     // Forward events from coreIndexer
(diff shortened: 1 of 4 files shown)


src/datastore/index.js
```diff
@@ -94,12 +94,12 @@ export class DataStore extends TypedEmitter {
    *
    * @param {MultiCoreIndexer.Entry<'binary'>[]} entries
    */
-  async #batch(entries) {
+  async #batchEntries(entries) {
```
achou11, line 97: not sure about the naming, open to suggestions

gmaclennan, line 97:
```suggestion
  async #handleEntries(entries) {
```

gmaclennan (approved):
Looks good to me. Made naming suggestion, but all looks great, good job.

PR #181, opened 2023-08-16
chore: consolidate encryption key columns in projectKeys table

### LINKED ISSUE #178: Move all project keys into single column, add project key
## Description

Thinking about the `projectKeys` table more, I don't think it makes sense to split the keys into separate columns, since we can't query on them anyway because they are encrypted.

We also need to persist the project key somewhere, in addition to the project secret key.

## Tasks

- [ ] Switch keys columns to single `encryptedKeys` column
- [ ] Add `projectKey` to encrypted message with all the keys

achou11:
Closes #178

### DIFF
--- a/src/schema/client.js
+++ b/src/schema/client.js
@@ -3,25 +3,12 @@
 // device
 import { blob, sqliteTable, text } from 'drizzle-orm/sqlite-core'
 import { dereferencedDocSchemas as schemas } from '@mapeo/schema'
-import { NAMESPACES } from '../core-manager/index.js'
 import { jsonSchemaToDrizzleColumns as toColumns } from './schema-to-drizzle.js'
 import { backlinkTable } from './utils.js'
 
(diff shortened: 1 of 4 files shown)


src/schema/client.js
```diff
@@ -3,25 +3,12 @@
...
-  for (const namespace of NAMESPACES) {
-    const columnName = namespace + 'EncryptionKey'
-    columns[columnName] = blob(columnName)
-  }
-
-  return sqliteTable('projectKeys', columns)
-}
+export const projectKeysTable = sqliteTable('projectKeys', {
+  projectId: text('projectId').notNull().primaryKey(),
+  encryptionKeys: blob('encryptionKeys').notNull(),
```
achou11, line 13: Making this not nullable makes sense right?

gmaclennan: yep. There will always be at least a projectKey

PR #184, opened 2023-08-16
feat: $blobs.getUrl and $blobs.create methods

### LINKED ISSUE #104: add blob store api to mapeo project class

sethvincent:
closes #104 

There are a couple todos that need to be addressed that I'm not sure about. See comment below.

### DIFF
--- a/src/blob-api.js
+++ b/src/blob-api.js
@@ -28,9 +28,9 @@ export class BlobApi {
    * @returns {Promise<string>}
    */
   async getUrl(blobId) {
-    const { type, variant, name } = blobId
+    const { driveId, type, variant, name } = blobId
     const port = await getPort(this.blobServer.server)
-    return `http://127.0.0.1:${port}/${this.projectId}/${this.blobStore.writerDriveId}/${type}/${variant}/${name}`
+    return `http://127.0.0.1:${port}/${this.projectId}/${driveId}/${type}/${variant}/${name}`
   }
 
   /**


src/mapeo-project.js
```diff
@@ -113,6 +115,25 @@ export class MapeoProject {
...
+    this.$blobs = {
+      /**
+       * 
+       * @param {import('./types.js').BlobId} blobId 
+       * @returns {String}
+       */
+      getUrl: (blobId) => {
+        const { driveId, type, variant, name } = blobId
+        // TODO: where is the hostname set?
+        // TODO: expose the projectId
```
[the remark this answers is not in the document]
achou11: > TODO: where is the hostname set?

My guess is that we'd want to pass an opt to the `MapeoProject` constructor to specify the hostname. In general, I think it makes sense for the hostname to be generated by whatever is creating the `BlobServer` instance e.g. at the app level. Currently the `BlobServer` doesn't [handle multiprojects](https://github.com/digidem/mapeo-core-next/blob/52e6f12b9a16473ce6e3fcc85e716570cbfa2385/src/blob-server/index.js#L11) but that's the goal for it so that only one instance of a server is created for the entire application. 

> TODO: expose the projectId

think it has to be exposed based on [how the `BlobServer` is implemented](https://github.com/digidem/mapeo-core-next/blob/52e6f12b9a16473ce6e3fcc85e716570cbfa2385/src/blob-server/fastify-plugin.js#L60)

src/blob-store/index.js
```diff
@@ -274,6 +277,55 @@ class PretendCorestore {
...
+      })
+    }
+  }
+
+  /**
+   * 
+   * @param {string} filepath 
+   * @returns {Omit<BlobId, 'driveId'>}
+   */
+  splitPath (filepath) {
```
achou11, line 325: Is there any need for this to be a public method?

```suggestion
  #splitPath (filepath) {
```

src/blob-store/index.js
```diff
@@ -274,6 +277,55 @@ class PretendCorestore {
     }
   }
 
+  /**
+   * Write blobs for provided variants of a file
+   * @param {{ original: string, preview?: string, thumbnail?: string }} filepaths 
+   * @param {{ mimeType?: string }} metadata
+   * @returns {Promise<{ original: Omit<BlobId, 'driveId'>, preview?: Omit<BlobId, 'driveId'>, thumbnail?: Omit<BlobId, 'driveId'> }>}
+   */
+  async create(filepaths, metadata) {
```
achou11, line 286: [Client docs](https://github.com/digidem/mapeo-core-docs/blob/main/client-docs/README.md#blobcreate) specify that `metadata` is an optional arg:

```suggestion
   * @param {{ mimeType?: string }} [metadata={}]
   * @returns {Promise<{ original: Omit<BlobId, 'driveId'>, preview?: Omit<BlobId, 'driveId'>, thumbnail?: Omit<BlobId, 'driveId'> }>}
   */
  async create(filepaths, metadata = {}) {
```

achou11 (commented):
feeling a little brain-drained so hopefully my comment about the hostname question makes some sense 😅

src/mapeo-project.js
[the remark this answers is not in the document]
gmaclennan: For port we'll need to expose a method on blobServer like `getPort()` - should be async and return once the server has started (it will need to close every time the app goes into the background). We can use that to ensure that getUrl only resolves when the blob server is actually running.

gmaclennan (changes requested):
Progress looks good, I made some comments.

How about keeping these methods (create() and getUrl()) in a separate class `BlobApi` that takes the blobStore as a constructor param? It feels like we're overloading the blobStore API here with something that has a different purpose.

src/blob-store/index.js
achou11: as in `@param {{ mimeType: string }} metadata`?

src/mapeo-project.js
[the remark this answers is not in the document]
gmaclennan: You should be able to do:

```ts
async function getPort(server: import('node:http').Server) {
  const addr = server.address()
  const port = addr && typeof addr === 'object' ? addr.port : null
  if (typeof port = 'number') return port
  await once(server, 'listening')
  return getPort(server)
}
```

You can get `server` from [`fastify.server()`](https://fastify.dev/docs/latest/Reference/Server#server)

No strong opinions on whether we pass down the fastify instance, http.server instance, or just the getPort function.

src/blob-api.js
```diff
@@ -0,0 +1,99 @@
...
+   */
+  async create(filepaths, metadata) {
+    const { original, preview, thumbnail } = filepaths
+    const { mimeType } = metadata
+    const blobType = getType(mimeType)
+    const hash = b4a.alloc(sodium.crypto_generichash_BYTES)
+    sodium.crypto_generichash(hash, b4a.from(original))
+
+    const originalBlobId = await writeFile({
+      name: hash,
```
gmaclennan, line 42:
```suggestion
      name: hash.toString('hex'),
```

gmaclennan (changes requested):
Good work on this. It needs tests though before merging, and I left a couple of comments including a suggestion about `getPort` and an undefined method `this.createWriteStream()`

src/blob-api.js
```diff
@@ -0,0 +1,124 @@
...
+            type: blobType,
+          },
+          metadata
+        )
+      : null
+    const thumbnailBlobId = thumbnail
+      ? await this.writeFile(
+          thumbnail,
+          {
+            name: `${name}_${basename(thumbnail)}`,
```
gmaclennan, line 74: Names of all variants should be the same

src/blob-api.js
```diff
@@ -0,0 +1,124 @@
...
+
+    const blobIds =
+      /** @type {{ original: Omit<BlobId, 'driveId'>, preview?: Omit<BlobId, 'driveId'>, thumbnail?: Omit<BlobId, 'driveId'> }} */ ({
+        original: originalBlobId,
+      })
+
+    if (previewBlobId) blobIds.preview = previewBlobId
+    if (thumbnailBlobId) blobIds.thumbnail = thumbnailBlobId
+
+    return blobIds
```
gmaclennan, line 90: See comment above about return type

src/blob-api.js
```diff
@@ -0,0 +1,124 @@
...
+    const { mimeType } = metadata
+    const blobType = getType(mimeType)
+    const hash = b4a.alloc(8)
+    sodium.randombytes_buf(hash)
+    const name = hash.toString('hex')
+
+    const originalBlobId = await this.writeFile(
+      original,
+      {
+        name: `${name}_${basename(original)}`,
```
gmaclennan, line 53: ok to just use the random name here

src/mapeo-project.js
```diff
@@ -113,6 +119,15 @@ export class MapeoProject {
...
       }),
     }
+
+    this.#blobStore = new BlobStore({
+      coreManager: this.#coreManager,
+    })
+
+    this.$blobs = new BlobApi({
+      projectId: this.#projectId,
+      blobStore: this.#blobStore,
```
gmaclennan, line 129: Missing blobServer param - can be added as a constructor param for MapeoProject. ping @achou we'll need to create a BlobServer instance in MapeoClient.

tests/helpers/blob-server.js
gmaclennan, line 1: not sure if this was intentional

tests/blob-api.js
```diff
@@ -0,0 +1,122 @@
...
+
+  t.is(typeof port, 'number')
+  t.is(port, 3456)
+
+  t.teardown(async () => {
+    await server.close()
+  })
+})
+
+test('get port after listening event with unset port', async (t) => {
```
gmaclennan, line 31: I think this is covered in the getUrl test below without needing to unit test getPort()?

tests/blob-api.js
```diff
@@ -0,0 +1,122 @@
+import { join, basename } from 'node:path'
+import { fileURLToPath } from 'url'
+import test from 'brittle'
+import { BlobApi } from '../src/blob-api.js'
+import { createBlobServer, getPort } from '../src/blob-server/index.js'
+import { createBlobStore } from './helpers/blob-store.js'
+import { timeoutException } from './helpers/index.js'
+
+test('get port after listening event with explicit port', async (t) => {
```
gmaclennan, line 9: I think this is tested in the getUrl test below, which avoids needing to test the `getPort`, which is an implementation detail.

gmaclennan (changes requested):
Looking good. Tests are failing though - need fixed before a merge. Plus a change needed in return type of `$blob.create`

src/blob-api.js
```diff
@@ -0,0 +1,124 @@
...
+    const { driveId, type, variant, name } = blobId
+    const port = await getPort(this.blobServer.server)
+    return `http://127.0.0.1:${port}/${this.projectId}/${driveId}/${type}/${variant}/${name}`
+  }
+
+  /**
+   * Write blobs for provided variants of a file
+   * @param {{ original: string, preview?: string, thumbnail?: string }} filepaths
+   * @param {{ mimeType: string }} metadata
+   * @returns {Promise<{ original: Omit<BlobId, 'driveId'>, preview?: Omit<BlobId, 'driveId'>, thumbnail?: Omit<BlobId, 'driveId'> }>}
```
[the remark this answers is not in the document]
gmaclennan: Ah very good point. It's actually a property on the writeStream e.g. `blobStore.createWriteStream().driveId` although I just realised that was not documented (my bad) and it's an awkward API too. Since the driveId for any written blobs will never change for a particular instance of the blobStore, I think it makes sense to just expose it as `blobStore.writerDriveId`, and use it from there. I've created [an issue](https://github.com/digidem/mapeo-core-next/issues/218) and [made a fix](https://github.com/digidem/mapeo-core-next/pull/219)

gmaclennan (changes requested):
Great work on getting this wrapped up, looks good to merge now once the driveId question is resolved. I'll merge #219 today and then you can use `blobStore.writerDriveId` instead of `metadata.driveId`. Once that is done we can merge.

PR #185, opened 2023-08-16
Lint/format setup

tomasciccola:
chore: remove eslintrc.js and move it to package.json
Also, add prettier plugin for eslint and update prettier to v3, so it plays nice with eslint


tomasciccola:
Ooops, beat me to it. Main difference I see is that in that PR `prettier` and `eslint` are managed separately? 
In this one `prettier` is managed as a plugin for `eslint`. I don't know the tradeoffs, this is the one that worked for me :'|

achou11:
> Main difference I see is that in that PR prettier and eslint are managed separately?

Yeah that's just how we've set it up for most of our other projects. That approach more closely reflects the separation between formatting and code linting, but can't elaborate on any significant advantages (major from an executables perspective the separation is helpful?)

gmaclennan:
Making prettier a plugin for eslint means that formatting becomes an error, which removes the utility of running eslint as part of tests. For those who use eslint as a plugin in their editor (e.g. me) it means that you get constant errors reported as you write code (prior to formatting). However eslint errors on their own (with the recommended defaults) are very useful for catching typos etc. as you write code, so they are very helpful to leave on.

In summary using eslint to format code also causes it to show formatting as an error, which removes the utility of linting as a tool to catch coding errors.

gmaclennan:
Closing in favour of #183

PR #186, opened 2023-08-16
chore: explicitly set "mode" opt for encryptionKeys column creation

achou11:
Drizzle's type inference/checking when using the table in queries is practically non-existent without specifying this (even though it defaults to this under the hood). Probably better to specify anyways.

Probably wasn't necessary to regenerate the drizzle files but did so anyways 🤷


achou11:
looks like the issue lies here: 

https://github.com/drizzle-team/drizzle-orm/blob/515afb05df6b0ecc69dacf0e73239828dd727772/drizzle-orm/src/sqlite-core/columns/blob.ts#L138-L140

The TS type fallback is the JSON builder type, which doesn't match the actual implementation (where the fallback is the Buffer builder type)

gmaclennan (approved):

achou11:
Reported bug in Drizzle: https://github.com/drizzle-team/drizzle-orm/issues/1064

Issue #171, opened 2023-08-14
Generate `projectId` with @mapeo/crypto?


achou11:
Is it the project key, the discovery ID, or something else (e.g. a different hash of the project key)?

Comment from Gregor in https://github.com/digidem/mapeo-core-next/pull/169#issuecomment-1677741083:

> We need to define what a projectId is - should it be a hex representation of the project public key, or a hex of the discovery key, or a base32 of the project key (this is what [hypercore uses as coreId](https://github.com/holepunchto/hypercore/blob/c154907fb940f4e1cf35e9d31f901ee50d23fe55/index.js#L371)), or should it be a hash of the project key with something else, so that knowing the project ID does not give you the information for discovery?

cc @sethvincent

gmaclennan:
I think the project ID should be a hash of the project key, but different to the discovery ID (which is a [hash of a hypercore key with the word 'hypercore'](https://github.com/holepunchto/hypercore-crypto/blob/4f9a72c296ef58ff461f97382434ad0e143412d3/index.js#L107)).

The discovery ID is "public" in a sense, since it only allows you to discover other peers, not actually connect to them. However we might want to share the project ID publicly (e.g. in reports generated in Mapeo), and we don't want an attacker to be able to use that key to discover who is online for a particular project.

I suggest using a hash with the word `mapeo` for the project ID e.g.

```js
const MAPEO = Buffer.from('mapeo')

function getProjectId (projectKey) {
  const digest = b4a.allocUnsafe(32)
  sodium.crypto_generichash(digest, MAPEO, projectKey)
  return digest
}
```

achou11:
this makes sense to me! created an issue in mapeo crypto to track: https://github.com/digidem/mapeo-crypto/issues/8

gmaclennan:
Fixed in digidem/mapeo-crypto#13

Issue #168, opened 2023-08-14
Implement `createProject()` for Mapeo client class


achou11:
High level implementation looks like:

1. create project keypair (which KeyManager should do)
2. create 4 encryption keys (1 for each namespace, random 32 bytes)
3. persist keys to sqlite table
4. create MapeoProject instance
5. write project name and any metadata to project instance using `project.$updateProjectSettings(...)`
6. return project id

PR #187, opened 2023-08-16
add project settings functionality to MapeoProject

### LINKED ISSUE #167: Add projectInfo indexing to MapeoProject
High level implementation:

-  accept `IndexWriter` instance as constructor parameter to `MapeoProject`
- `projectInfoIndexWriter.batch()` to be called with any projectInfo records found in hypercores, after validating that `docId` matches `projectId`.

achou11:
Closes #167 

Notes:

- implements project records indexing on the config store
- introduces the `$setProjectSettings` and `$getProjectSettings` methods (which also make it easier to test the indexing)

TODO:

- [x] (better) tests
- [x] update to `@mapeo/schema` https://github.com/digidem/mapeo-schema/issues/121
- [x] update to `@mapeo/schema` https://github.com/digidem/mapeo-schema/issues/132
- [ ] (non-blocking) update to `@mapeo/crypto` https://github.com/digidem/mapeo-core-next/issues/171


achou11:
running into a couple of implementation questions that I probably need some help with:

1. Because the project table lives in a separate database from the other mapeo schema data types, I need a Drizzle instance that uses a SQLite instance that points to the client db. The one created in `MapeoProject` only creates one pointing to the project db right now. At the time of writing, I currently add a "config" object that specifies the db instance, which feels off. With this, every project that gets created will be sharing using a shared "client" db instance, which on paper makes sense but also has me somewhat skeptical... 

2. What's responsible for actually creating the project i.e. writing the very first "project" record? Seems like that would be the responsibility of whatever holds the project `DataType` via `dataType.create(...)`, which is the `MapeoProject` at the time of writing. However, since `DataType.create` is async, we can't call it in the constructor, meaning that we'd need to expose a public method that the creator of the instance would then call. For example, the wrapper client class `createProject` method would look something like:

```ts
class MapeoManager {
  createProject(value) {
    const project = new MapeoProject(...)
    await project.$create(value)
    return project
  }
}
```

   Does that seem reasonable, or do I need to approach the design of this a little differently? Or maybe I'm missing an important detail...

achou11:
After discussion with Gregor:

1. My approach is valid. Not pretty but probably no noticeably better way of doing it for now

2. I was onto something here and turns out we need https://github.com/digidem/mapeo-core-next/issues/190 in order to address the issue. That method will be used in the `MapeoProject.$setProjectSettings` to provide upsert behavior. So the wrapper client would create the `MapeoProject` and then calls `$setProjectSettings`, with the project setting "creation" being taken care of in the MapeoProject implementation

src/mapeo-project.js
```diff
@@ -124,4 +166,34 @@ export class MapeoProject {
...
+    }
+
+    return this.#dataTypes.project[kCreateWithDocId](this.#projectId, {
+      ...settings,
+      schemaName: 'project',
+    })
+  }
+
+  /**
+   * @returns {Promise<import('@mapeo/schema').Project>}
```
achou11, line 194: I have this returning the entire project document for now, but guessing we'd want less information to surface in general.

src/mapeo-project.js
```diff
@@ -124,4 +166,34 @@ export class MapeoProject {
   get field() {
     return this.#dataTypes.field
   }
+
+  /**
+   * @param {Omit<import('@mapeo/schema').ProjectValue, 'schemaName'>} settings
```
achou11, line 171: kind of forgot what `defaultPresets` represents - should it also be omitted here?

src/datastore/index.js
```diff
@@ -25,7 +25,7 @@ import pDefer from 'p-defer'
 
 const NAMESPACE_SCHEMAS = /** @type {const} */ ({
   data: ['observation'],
-  config: ['preset', 'field'],
+  config: ['preset', 'field', 'project'],
   auth: [],
```
achou11, line 29: I think there should be a comment here explaining what this is used for. took me a bit to figure out I needed to update it to add the project schema

src/mapeo-project.js
```diff
@@ -34,8 +43,12 @@ export class MapeoProject {
    * @param {Buffer} opts.projectKey 32-byte public key of the project creator core
    * @param {Buffer} [opts.projectSecretKey] 32-byte secret key of the project creator core
    * @param {Partial<Record<import('./core-manager/index.js').Namespace, Buffer>>} [opts.encryptionKeys] Encryption keys for each namespace
+   * @param {ProjectSettingsConfig} opts.projectSettingsConfig
    */
-  constructor({ storagePath, ...coreManagerOpts }) {
+  constructor({ storagePath, projectSettingsConfig, ...coreManagerOpts }) {
+    // TODO: Update to use @mapeo/crypto when ready (https://github.com/digidem/mapeo-core-next/issues/171)
+    this.#projectId = coreManagerOpts.projectKey.toString('hex')
```
achou11, line 50: just highlighting this as something that will eventually be updated

src/mapeo-project.js
```diff
@@ -34,8 +43,12 @@ export class MapeoProject {
    * @param {Buffer} opts.projectKey 32-byte public key of the project creator core
    * @param {Buffer} [opts.projectSecretKey] 32-byte secret key of the project creator core
    * @param {Partial<Record<import('./core-manager/index.js').Namespace, Buffer>>} [opts.encryptionKeys] Encryption keys for each namespace
+   * @param {ProjectSettingsConfig} opts.projectSettingsConfig
```
gmaclennan, line 46: This name doesn't feel right. We might well use this indexWriter and the db for other things in the future. I suggest:

```js
/**
 * @param {Database} opts.sharedDb
 */
```

src/mapeo-project.js
```diff
@@ -86,7 +99,31 @@ export class MapeoProject {
       config: new DataStore({
         coreManager: this.#coreManager,
         namespace: 'config',
-        batch: (entries) => indexWriter.batch(entries),
+        batch: async (entries) => {
```
gmaclennan, line 102:
```suggestion
        batch: (entries) => this.#handleConfigEntries(entries)
```
Just in the name of keeping the constructor method easier to read, since it's getting quite long.

gmaclennan (changes requested):
Great work on this, there was a lot of tricky stuff to work through. Mainly all good, but a few things to fix.

src/mapeo-project.js
achou11: Is it possible to modify the schema given https://github.com/digidem/mapeo-schema/issues/104?

test-e2e/project-settings.js
```diff
@@ -0,0 +1,204 @@
...
+    label: 'my label',
+  },
+  {
+    schemaName: 'project',
+    name: 'myProject updated',
+    defaultPresets: {},
+  },
+]
+
+test('Project settings indexer only indexes project record types', async (t) => {
```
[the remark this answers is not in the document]
achou11: yeah let's discuss!

src/mapeo-project.js
gmaclennan: Yes, we allow fields to be `undefined`. We just don't allow them to be `null`

achou11: issue created: https://github.com/digidem/mapeo-schema/issues/132

achou11:
Tests will fail on latest push because it assumes that https://github.com/digidem/mapeo-schema/issues/132 is implemented

src/mapeo-project.js
```diff
@@ -86,7 +96,31 @@ export class MapeoProject {
...
+              projectSettingsEntries.push(entry)
+            } else {
+              otherEntries.push(entry)
+            }
+          }
+
+          await Promise.all([
+            indexWriter.batch(otherEntries),
+            projectSettingsConfig.indexWriter.batch(projectSettingsEntries),
+          ])
```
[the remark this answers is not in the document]
achou11: addressed via https://github.com/digidem/mapeo-core-next/pull/187/commits/6763de9a9d74d72ad78c5bffe011e636eed81a96

src/mapeo-project.js
```diff
@@ -124,4 +163,29 @@ export class MapeoProject {
   get field() {
     return this.#dataTypes.field
   }
+
+  /**
+   * @param {import('./types.js').ProjectSettings} settings
+   * @param {string} [versionId]
```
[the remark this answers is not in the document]
achou11: addressed via https://github.com/digidem/mapeo-core-next/pull/187/commits/2cc9d44ba14b587b5a4f52b118d6d2cd11ad29a5

src/mapeo-project.js
achou11: > I think this should be Partial<Omit<ProjectValue, 'schemaName'>>

addressed via https://github.com/digidem/mapeo-core-next/pull/187/commits/2cc9d44ba14b587b5a4f52b118d6d2cd11ad29a5

tsconfig.json
```diff
@@ -22,7 +22,7 @@
       ]
     }
   },
-  "files": ["tests/schema.js", "test-e2e/project-crud.js"],
+  "files": ["tests/schema.js", "test-e2e/project-crud.js", "test-e2e/project-settings.js", "test-e2e/utils.js"],
   "include": ["src/**/*", "types/**/*.d.ts"],
```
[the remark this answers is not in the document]
achou11: addressed via https://github.com/digidem/mapeo-core-next/pull/187/commits/d5ed306e4ae4eade26e426d74f49fca953112e54

src/mapeo-project.js
achou11: addressed via https://github.com/digidem/mapeo-core-next/pull/187/commits/6c5822ad6c6654925f8629649c7c7d97c7661dab

src/mapeo-project.js
achou11: addressed via https://github.com/digidem/mapeo-core-next/pull/187/commits/bd311a2de61c3fc1d42a03e34ad1f3538846f232

src/mapeo-project.js
```diff
@@ -124,4 +180,36 @@ export class MapeoProject {
...
+      ...settings,
+      schemaName: 'project',
+    })
+  }
+
+  /**
+   * @returns {Promise<Omit<import('@mapeo/schema').Project, 'schemaName'>>}
+   */
+  async $getProjectSettings() {
+    return this.#dataTypes.project.getByDocId(this.#projectId)
```
gmaclennan, line 213: Catch and return `{}`

src/mapeo-project.js
```diff
@@ -124,4 +180,36 @@ export class MapeoProject {
...
   }
+
+  /**
+   * @param {Partial<Omit<import('@mapeo/schema').ProjectValue, 'schemaName'>>} settings
+   * @returns {Promise<import('@mapeo/schema').Project>}
+   */
+  async $setProjectSettings(settings) {
+    const { project } = this.#dataTypes
+
+    const existing = await project.getByDocId(this.#projectId).catch(() => {
```
gmaclennan, line 191: Round about way of avoiding a try catch but it works 👍

gmaclennan (commented):
Code looks good. I haven't thoroughly reviewed the tests, maybe we can go through those together?

src/mapeo-project.js
achou11: any objections to returning `null` here instead?

achou11: addressed via https://github.com/digidem/mapeo-core-next/pull/187/commits/a6a468f2dfc6c688975d9ebead97b8364b0cfbdf. went ahead and returned `null` instead since it makes more sense to me 😄

achou11: hm starting to see why you suggested `{}`. doing the `null` checks after calling this may get annoying. still think i prefer it to `{}` and the code you'd have to write to account for it, but open to alternative reasoning

achou11:
After discussion with Gregor, the tests introduced in this PR will be updated to use the work in #168. They should be passing at this point with the latest commit but in case someone reviews this and thinks the tests need more work, would rather update them in that other PR to make them properly more e2e

src/mapeo-project.js
achou11: addressed via https://github.com/digidem/mapeo-core-next/pull/187/commits/fe56a1a52d29eb48742649adb418e9ead3a8bf06

src/mapeo-project.js
```diff
@@ -124,4 +166,34 @@ export class MapeoProject {
   get field() {
     return this.#dataTypes.field
   }
+
+  /**
+   * @param {Omit<import('@mapeo/schema').ProjectValue, 'schemaName'>} settings
+   * @param {string} [versionId]
+   * @returns {Promise<import('@mapeo/schema').Project>}
+   */
+  async $setProjectSettings(settings, versionId) {
```
achou11: does it make sense for this to return the whole updated document, or just a subset? maybe something similar to `$getProjectSettings`?

achou11: decided to return just a subset (see `EditableSettings` type)

gmaclennan (approved):
Looks good, I'll commit my naming suggestion, then go ahead and merge.

Issue #196, opened 2023-08-21
Add `config` core to encryption keys message in Invites


gmaclennan:
## Tasks

- [ ]

gmaclennan:
Fixed in #202

PR #199, opened 2023-08-21
implement wrapper client containing createProject and getProject methods

### LINKED ISSUE #168: Implement `createProject()` for Mapeo client class

achou11:
Closes #168 

Note: this is currently stacked on #187 as it relies on changes in that PR

TODO:

- [x] update project settings e2e tests to use MapeoManager class for setup instead of `setupSharedResources`. Potentially can just merge those into the manager-basic test file
- [X] update to `@mapeo/schema` https://github.com/digidem/mapeo-schema/issues/132
- [ ] (non-blocking) update to `@mapeo/crypto` https://github.com/digidem/mapeo-crypto/issues/14


src/mapeo-manager.js
```diff
@@ -0,0 +1,194 @@
+import { randomBytes } from 'crypto'
+import path from 'path'
+import { serialize, deserialize } from 'v8'
```
achou11, line 3: this seems to be the preferred way of (de)serializing a buffer for on-disk persistence. https://nodejs.org/docs/latest-v18.x/api/v8.html#v8serializevalue

This avoids json stringifying the encryption keys object and then creating a buffer from it, which changes the actual structure of the object for fields with buffer values i.e. 

```js
JSON.stringify({ foo: Buffer.from('foo') })
``` 

turns into

```json
{"foo":{"type":"Buffer","data":[102,111,111]}}
```

which requires extra work when reading from the db and deserializing

src/mapeo-manager.js
```diff
@@ -0,0 +1,194 @@
...
+
+    // 6. Return project id
+    return projectId
+  }
+
+  /**
+   * @param {string} projectId
+   * @returns {MapeoProject}
+   */
+  getProject(projectId) {
```
achou11, line 122: should this be async? nothing called in this requires it to be so

src/mapeo-manager.js
```diff
@@ -0,0 +1,194 @@
...
+
+    return project
+  }
+}
+
+/**
+ * @param {{[key: string]: unknown}} keys
+ * @returns {asserts keys is ProjectKeys}
+ */
+function validateProjectKeys(keys) {
```
achou11, line 169: open to suggestions for a better way to do this, if possible. basically want to validate the deserialized encryption keys that are read to make sure that they have the correct structure and associated types. Was thinking maybe using Typebox but not sure how it plays with buffer types, since that isn't an official JSON type

src/mapeo-manager.js
```diff
@@ -0,0 +1,194 @@
...
+  /**
+   * Create a new project.
+   * @param {import('type-fest').Simplify<Partial<Omit<import("@mapeo/schema").ProjectValue, 'schemaName'>>>} settings
+   * @returns {Promise<string>}
+   */
+  async createProject(settings) {
+    // 1. Create project keypair
+    const projectKeypair = {
+      publicKey: randomBytes(32),
+      secretKey: randomBytes(64),
```
achou11, line 55: highlighting this because I'm not actually sure how the keys are supposed to be generated

src/mapeo-manager.js
```diff
@@ -0,0 +1,194 @@
...
+      encryptionKeys: namespaceEncryptionKeys,
+      keyManager: this.#keyManager,
+      projectKey: projectKeypair.publicKey,
+      projectSecretKey: projectKeypair.secretKey,
+    })
+
+    // 5. Write project name and any other relevant metadata to project instance
+    await project.$setProjectSettings({
+      // TODO: What's the fallback name?
+      name: settings.name || '',
```
achou11, line 110: highlighting this question

gmaclennan:
Just took a quick look, not a full review. Sorry I'd forgotten about the keys being buffers and the encoding question. There are several options for encoding to a buffer, and v8 serialize is one, but I worry that depending on v8 will come back and bite us at some point in the future. We actually have a pbf encoder for the encryption keys in rpc/messages.js. You should be able to do `Invite_EncryptionKeys.decode()` and it should be type-safe. I suggest using that, although we should also probably move the `EncryptionKeys` message out of the `Invite` namespace.

achou11:
> Just took a quick look, not a full review. Sorry I'd forgotten about the keys being buffers and the encoding question. There are several options for encoding to a buffer, and v8 serialize is one, but I worry that depending on v8 will come back and bite us at some point in the future. We actually have a pbf encoder for the encryption keys in rpc/messages.js. You should be able to do `Invite_EncryptionKeys.decode()` and it should be type-safe. I suggest using that, although we should also probably move the `EncryptionKeys` message out of the `Invite` namespace.

Yeah that was something I was thinking about as I started using it. Was way easier to use than anything I was aware of, so I stuck with it.

As a note if `Invite_EncryptionKeys` gets extracted, looks like it doesn't handle the following keys for this use case (at least from a types perspective but probably also at runtime): `projectKey`, `projectSecretKey`, and `config` (last one is going to be addressed via https://github.com/digidem/mapeo-core-next/issues/196)

src/mapeo-manager.js
gmaclennan: This makes me think that we should actually export a method from @mapeo/crypto specifically for encrypt/decrypt of this e.g. `decryptProjectKeys()`. I think this is ok for now, but can you open an issue to address this.

src/mapeo-manager.js
gmaclennan: [`KeyManager.generateProjectKeypair()`](https://github.com/digidem/mapeo-crypto#keymanagergenerateprojectkeypair) (note: static method). You created [the issue](https://github.com/digidem/mapeo-crypto/issues/4) for it 🙂

src/mapeo-manager.js
```diff
@@ -0,0 +1,194 @@
...
+  async createProject(settings) {
+    // 1. Create project keypair
+    const projectKeypair = {
+      publicKey: randomBytes(32),
+      secretKey: randomBytes(64),
+    }
+
+    // 2. Create namespace encryption keys
+    /** @type {Record<import('./core-manager/core-index.js').Namespace, Buffer>} */
+    const namespaceEncryptionKeys = {
```
gmaclennan, line 60:
```suggestion
    const encryptionKeys = {
```
Just a suggestion to reduce typing (see below) 🙂

src/mapeo-manager.js
```diff
@@ -0,0 +1,194 @@
...
+    // 3. Save keys to client db in projectKeys table
+    /** @type {ProjectKeys} */
+    const keys = {
+      projectKey: projectKeypair.publicKey,
+      projectSecretKey: projectKeypair.secretKey,
+      authEncryptionKey: namespaceEncryptionKeys.auth,
+      blobEncryptionKey: namespaceEncryptionKeys.blob,
+      blobIndexEncryptionKey: namespaceEncryptionKeys.blobIndex,
+      configEncryptionKey: namespaceEncryptionKeys.config,
+      dataEncryptionKey: namespaceEncryptionKeys.data,
```
gmaclennan, line 77:
```suggestion
      encryptionKeys
```

src/mapeo-manager.js
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+    // TODO: Update to use @mapeo/crypto when ready (https://github.com/digidem/mapeo-core-next/issues/171)
+    const projectId = projectKeypair.publicKey.toString('hex')
+
+    this.#db
+      .insert(projectKeysTable)
+      .values({
+        projectId,
+        encryptionKeys: this.#keyManager.encryptLocalMessage(
```
gmaclennan, line 87: Naming suggestion. These aren't actually all encryption keys
```suggestion
        keysCipher: this.#keyManager.encryptLocalMessage(
```

src/mapeo-manager.js
```diff
@@ -0,0 +1,194 @@
...
+    })
+    this.#storagePath = storagePath
+  }
+
+  /**
+   * Create a new project.
+   * @param {import('type-fest').Simplify<Partial<Omit<import("@mapeo/schema").ProjectValue, 'schemaName'>>>} settings
+   * @returns {Promise<string>}
+   */
+  async createProject(settings) {
```
gmaclennan, line 51: I suggest doing a `Pick<>` here. Right now the UX is only going to allow choosing a project name at creation time, but that should be optional.

src/mapeo-manager.js
gmaclennan: Leave as undefined if no name. I think the schema probably needs updated to make this optional.

src/mapeo-manager.js
```diff
@@ -0,0 +1,194 @@
...
+      keyManager: this.#keyManager,
+      projectKey: projectKeypair.publicKey,
+      projectSecretKey: projectKeypair.secretKey,
+    })
+
+    // 5. Write project name and any other relevant metadata to project instance
+    await project.$setProjectSettings({
+      // TODO: What's the fallback name?
+      name: settings.name || '',
+      defaultPresets: settings.defaultPresets || {},
```
gmaclennan, line 111: Let's not set this at creation time. We'll build a default into the code for when this is undefined

src/mapeo-manager.js
gmaclennan: In general I've made all methods that access the db async. I did this in case we need to move db queries onto a worker thread in the future - this will make it easier. Plus it means that tests we write here will also work for testing the client implementation (which proxies methods over rpc-reflector).

src/mapeo-manager.js
```diff
@@ -0,0 +1,194 @@
...
+  getProject(projectId) {
+    const result = this.#db
+      .select({
+        encryptionKeys: projectKeysTable.encryptionKeys,
+      })
+      .from(projectKeysTable)
+      .where(eq(projectKeysTable.projectId, projectId))
+      .get()
+
+    if (!result.encryptionKeys) {
```
gmaclennan, line 131: Need to test `result` is not undefined - the sqlite type is broken in drizzle but should be fixed in new version.

Maybe make encryptionKeys a required field? And rename to `keysCipher`

src/mapeo-manager.js
gmaclennan: ok to skip validation here, since there is no unknown input between writes to the DB and reads - same reason drizzle doesn't do validation internally on reads and writes.

src/types.ts
```diff
@@ -168,3 +169,11 @@ export type ProtocolStream = NoiseStream & { userData: Protomux }
...
   [K in keyof T]: [K, T[K]]
 }[keyof T][]
+
+export type ProjectKeys = {
+  projectKey: Buffer
+  projectSecretKey?: Buffer
+  authEncryptionKey: Buffer
+} & {
+  [n in `${Exclude<(typeof NAMESPACES)[number], 'auth'>}EncryptionKey`]?: Buffer
+}
```
gmaclennan, line 179: See renaming comment above - you can reduce a lot of typing with:

```suggestion
import { SetRequired } from 'type-fest'

export type ProjectKeys = {
  projectKey: Buffer
  projectSecretKey?: Buffer
  encryptionKeys: SetRequired<{
    [typeof NAMESPACES)[number]]?: Buffer
  }, 'auth'>
}
```

gmaclennan (changes requested):
Good work on this. Made some suggestions. Maybe we can discuss testing strategy this week?

src/mapeo-manager.js
gmaclennan: ok great, then ok to just do `if (!result) { .. }`

gmaclennan: Fix should be in [drizzle@0.28.2](https://github.com/drizzle-team/drizzle-orm/releases/tag/0.28.2)

src/mapeo-manager.js
achou11: done! https://github.com/digidem/mapeo-crypto/issues/10

achou11: addressed via https://github.com/digidem/mapeo-core-next/pull/199/commits/6d14bacb6f178af5102c8d855efffbb0d3b4c085

src/mapeo-manager.js
achou11: addressed via https://github.com/digidem/mapeo-core-next/pull/199/commits/6d14bacb6f178af5102c8d855efffbb0d3b4c085

src/mapeo-manager.js
achou11: addressed via https://github.com/digidem/mapeo-core-next/pull/199/commits/6d14bacb6f178af5102c8d855efffbb0d3b4c085

src/mapeo-manager.js
achou11: addressed via 115f6bedb1bf8f330c597402982cbac219187b6d

src/mapeo-manager.js
achou11: addressed via f68d04926c8483559047b65c1d306ea83f4e7296

src/mapeo-manager.js
achou11: addressed via 6d14bacb6f178af5102c8d855efffbb0d3b4c085

src/mapeo-manager.js
achou11: addressed via 6d14bacb6f178af5102c8d855efffbb0d3b4c085

src/mapeo-manager.js
achou11: addressed via 6d14bacb6f178af5102c8d855efffbb0d3b4c085

src/mapeo-manager.js
achou11: addressed via 9bc05084882ba8f7ebea7a87f39a85dfbb464752

src/mapeo-manager.js
achou11: addressed via f68d04926c8483559047b65c1d306ea83f4e7296

src/types.ts
achou11: addressed via 6d14bacb6f178af5102c8d855efffbb0d3b4c085

package.json
```diff
@@ -104,7 +104,7 @@
     "better-sqlite3": "^8.3.0",
     "compact-encoding": "^2.12.0",
     "corestore": "^6.8.4",
-    "drizzle-orm": "^0.28.1",
+    "drizzle-orm": "0.28.2",
```
achou11, line 107: currently pinned this because type exports are broken in 0.28.3: https://github.com/drizzle-team/drizzle-orm/issues/1088

src/mapeo-manager.js
```diff
@@ -0,0 +1,139 @@
...
+
+    // TODO: Update to use @mapeo/crypto when ready (https://github.com/digidem/mapeo-core-next/issues/171)
+    const projectId = projectKeypair.publicKey.toString('hex')
+
+    this.#db
+      .insert(projectKeysTable)
+      .values({
+        projectId,
+        keysCipher: this.#keyManager.encryptLocalMessage(
+          Buffer.from(ProjectKeys.encode(keys).finish()),
```
achou11, line 78: is this the correct way of doing this? seems a little verbose but needed for TS reasons (`finish` returns `Uint8Array` but `encryptLocalMessage` accepts `Buffer`)

gmaclennan: Ah I should update the key manager method to accept uint8array. Fine to TS ignore for now. At some point need to go through the code and use b4a.

src/mapeo-manager.js
```diff
@@ -0,0 +1,137 @@
...
+      sharedIndexWriter: this.#projectSettingsIndexWriter,
+    })
+
+    // 5. Write project name and any other relevant metadata to project instance
+    await project.$setProjectSettings({
+      name: settings.name,
+    })
+
+    // 6. Return project id
+    return projectId
```
gmaclennan, line 101: If not returning the project here then we should close it - although no close method exists yet! Must be careful of having two or more instances with the same files open

gmaclennan: Can be a follow up issue

gmaclennan (commented):
Good work. We need to wait for tests to be fixed?

achou11:
> We need to wait for tests to be fixed?

Regrettably implemented changes that are now reliant on the project info schema being updated to allow optionals. probably would've made more sense to do as a follow-up. Couple of options to make this unblocked:

1. skip the tests added in this PR that fail because of ^
2. revert changes related to the assumption above

any preferences?

achou11:
also tests aren't running on this PR since it's stacked on a different one, but tests will fail for the same reason

src/mapeo-manager.js
achou11: created: https://github.com/digidem/mapeo-core-next/issues/207

src/mapeo-manager.js
achou11: Issue created: https://github.com/digidem/mapeo-crypto/issues/14

achou11: simplified based on your suggestion and added a todo comment via https://github.com/digidem/mapeo-core-next/pull/199/commits/b3634e6b3b45e0de5ffbb7bf6ce3f82e0c177caf

achou11:
@gmaclennan this should be in a good state to review now (tests aren't running since it's stacked though, but they pass locally for me 😄 )

src/datatype/index.js
```diff
@@ -121,7 +121,8 @@ export class DataType {
    * @param {string} docId
    */
   async getByDocId(docId) {
-    return deNullify(this.#sql.getByDocId.get({ docId }))
+    const result = this.#sql.getByDocId.get({ docId })
+    return result ? deNullify(result) : result
```
gmaclennan, line 125: This will possibly return `undefined`. I think we want to throw for notFound, and always return a value. Can you open a new issue to address this outside this PR. Ok with this change to keep typescript happy for now, I would have thought this would cause the type tests to fail - we should look at why they aren't

src/mapeo-manager.js
```diff
@@ -0,0 +1,148 @@
...
+    const dbPath =
+      storagePath !== undefined
+        ? path.join(storagePath, CLIENT_SQLITE_FILE_NAME)
+        : ':memory:'
+
+    const sqlite = new Database(dbPath)
+    this.#db = drizzle(sqlite)
+    migrate(this.#db, { migrationsFolder: './drizzle/client' })
+
+    this.#keyManager = new KeyManager(KeyManager.generateRootKey())
```
gmaclennan, line 37: `rootKey` should be a constructor parameter. Otherwise we have no way of recreating this instance for tests (or actual use)

src/mapeo-manager.js
```diff
@@ -0,0 +1,148 @@
...
+    const projectId = projectKeypair.publicKey.toString('hex')
+
+    this.#db
+      .insert(projectKeysTable)
+      .values({
+        projectId,
+        keysCipher: this.#keyManager.encryptLocalMessage(
+          // TODO: Type error should go away once https://github.com/digidem/mapeo-crypto/issues/14 is resolved
+          // @ts-expect-error
+          ProjectKeys.encode(keys).finish(),
```
gmaclennan, line 83:
```suggestion
          Buffer.from(ProjectKeys.encode(keys).finish().buffer),
```
see https://github.com/digidem/mapeo-crypto/issues/14. Using the `.buffer` [avoids making a copy](https://nodejs.org/api/buffer.html#buffer_static_method_buffer_from_arraybuffer_byteoffset_length)

src/mapeo-manager.js
```diff
@@ -0,0 +1,148 @@
...
+      .select({
+        keysCipher: projectKeysTable.keysCipher,
+      })
+      .from(projectKeysTable)
+      .where(eq(projectKeysTable.projectId, projectId))
+      .get()
+
+    if (!result) {
+      throw new Error(
+        `Could not find keys cipher for project with ID ${projectId}`
```
gmaclennan, line 130:
```suggestion
        `NotFound: project ID ${projectId} not found`
```

Less obscure error description

gmaclennan (changes requested):
Great work. Good to merge after committing suggestions

src/mapeo-manager.js
achou11: using `.buffer` leaves me with this 

image

achou11: or am i supposed to still do `Buffer.from` here? i.e. `Buffer.from(ProjectKeys.encode(keys).finish().buffer)`

src/datatype/index.js
gmaclennan: oh yeah I'd been meaning to update that CI config

src/mapeo-manager.js
gmaclennan: Yes, I think if I understand the question, use as my suggested commit above (haven't checked it works)

src/mapeo-manager.js
gmaclennan: I think required is best.

src/mapeo-manager.js
achou11: addressed via https://github.com/digidem/mapeo-core-next/pull/199/commits/793707cb049f0c6e3b20899e948911a10ae05c6f

src/mapeo-manager.js
achou11: yeah forgot to call Buffer.from in my initial question - didn't actually read the suggestion when implementing the change (not a morning person 😄 )

achou11: addressed via https://github.com/digidem/mapeo-core-next/pull/199/commits/a59a97a4e70a7cabb466920bac0810e0cb728730

src/mapeo-manager.js
achou11: addressed via https://github.com/digidem/mapeo-core-next/pull/199/commits/f4d60c5327c468407190304c4cda13d28ff0fbd5

achou11:
Did a rebase on main to unstack this but it was kind of tricky (probably should've used merge instead). Seems like it only made a couple of minor mistakes (fixed now) but in case anything in the diff looks weird 😅

gmaclennan (approved):

PR #213, opened 2023-08-23
Feat: swarm discovery

### LINKED ISSUE #164: Implement `HyperswarmDiscovery`
> (or a different class name)

- Join topics for all projects on the device or that are set to be synced with remotely
- Ensure only remote peers connect through hyperswarm. Context: https://github.com/digidem/mapeo-core-next/issues/93
- Provide methods for starting and stopping hyperswarm to allow users to allow or disallow remote connections
- Ensure we deduplicate peers that try to join from both mdns and hyperswarm
- Provide suspend and resume methods #13

tomasciccola:
this should close #164

### DIFF
--- a/src/discovery/hyperswarm.js
+++ b/src/discovery/hyperswarm.js
@@ -1,0 +1,35 @@
+import { TypedEmitter } from 'tiny-typed-emitter'
+import Hyperswarm from 'hyperswarm'
+
+export class HyperSwarmDiscovery extends TypedEmitter {
+  /** @type {String} */
+  #projectDiscoveryId
+  #hyperswarm
+  constructor() {
+    super()
+  }
+  /**
+   * @param {String} projectDiscoveryId
+   * @returns {Promise<void>}
+   */
+  async join(projectDiscoveryId) {
+    this.#projectDiscoveryId = projectDiscoveryId
+    if (!this.#hyperswarm) {
+      this.#hyperswarm = new Hyperswarm()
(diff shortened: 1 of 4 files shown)


src/discovery/hyperswarm.js
```diff
@@ -0,0 +1,52 @@
...
+      })
+    }
+  }
+
+  leave() {}
+
+  async stop() {
+    const hyperswarm = this.#hyperswarm
+    this.#hyperswarm = null
+    this.removeAllListeners('connection')
```
gmaclennan, line 49: We shouldn't do this - it's ok to keep the same connection listeners attached to `this` across a `stop()` then a `join()`, and generally it's the responsibility of code that attaches listeners to remove them.

However we should remove the listener from the hyperswarm e.g. `hyperswarm.removeAllListeners()`

gmaclennan:
It's because we want to control whether there is any network traffic, and whether we are listening, on a project-by-project basis. If we had a single hyper swarm instance I think the code would be complicated for managing when from each project setDiscovery(false) or setSync(false) is called. So having a instance for each project (that only joins one topic) makes the implementation easier. Also in reality with our projected UI a user would only ever be syncing one project at a time anyway, so even though there will be an instance per project, only one will be listening at any time anyway

achou11:
Can we close this? cc @EvanHahn

EvanHahn:
I think so. @tomasciccola, do you mind if we close?

tomasciccola:
not at all!

PR #208, opened 2023-08-22
implement listProjects method for MapeoManager class

achou11:
Closes #206 

~NOTE: Stacked on #199~ 

Also:

- ~adds a `$getProjectInfo` to the `MapeoProject` class, which is used to retrieve information that's used for the resulting `MapeoManager.listProjects()` return value. `$getProjectInfo` returns the `ProjectInfo` type, which represents meta information about a project (different from its settings). It is currently defined as~ EDIT: No longer relevant

```ts
type ProjectInfo = {
  createdAt: Date
  updatedAt: Date
  projectId: string
}
```

PR #215, opened 2023-08-23
implement addProject method for MapeoManager class

### LINKED ISSUE #209: Implement `addProject` method for wrapper client class
## Description

Part of https://github.com/digidem/mapeo-core-next/issues/130

Should approximately follow what's spec'd here: https://github.com/digidem/mapeo-core-docs/blob/main/client-docs/README.md#clientaddproject

achou11:
Closes #209 

~NOTE: Stacked on #208~

Lots of implementation questions:

- ~my understanding is that adding the project should update the `ProjectKeys` table in the client db if it's not there. However, how do we get the `auth` encryption key if that's the case? (which is required in the `encryptionKeys` column value)~ EDIT: addressed
- ~I'm guessing that we don't want to call `project.$setProjectSettings()` after creating the project instance, like we do in `manager.createProject()`. If that's the case, does it still make sense to return the project instance in this `addProject()` method? doing so seems to cause some issues unless I'm missing an important step:~ EDIT: addressed

TODO:

- [x] Add tests. Not really sure what to test here...Was thinking that there's gotta be some distinction between creating a project and adding a project created by someone else i.e. you don't necessarily have the project secret key + encryption keys for the project (in the latter).
- [ ] (non-blocking) wait for https://github.com/digidem/mapeo-core-next/pull/223 to fix type issue with `Invite`

### DIFF
--- a/src/mapeo-manager.js
+++ b/src/mapeo-manager.js
@@ -48,6 +48,34 @@ export class MapeoManager {
   }
 
   /**
+   * @param {Buffer} keysCipher
+   * @param {string} projectId
+   * @returns {ProjectKeys}
+   */
+  #decodeProjectKeysCipher(keysCipher, projectId) {
+    return ProjectKeys.decode(
+      this.#keyManager.decryptLocalMessage(keysCipher, projectId)
+    )
+  }
+
+  /**
+   * @param {string} projectId
+   * @param {ProjectKeys} keys
+   */
+  #saveProjectKeys(projectId, keys) {
+    this.#db
+      .insert(projectKeysTable)
+      .values({
+        projectId,
+        keysCipher: this.#keyManager.encryptLocalMessage(
+          Buffer.from(ProjectKeys.encode(keys).finish().buffer),
+          projectId
+        ),
+      })
+      .run()
+  }
+
+  /**
    * Create a new project.
    * @param {import('type-fest').Simplify<Partial<Pick<ProjectValue, 'name'>>>} [settings]
    * @returns {Promise<string>}
@@ -77,16 +105,7 @@ export class MapeoManager {
     // TODO: Update to use @mapeo/crypto when ready (https://github.com/digidem/mapeo-core-next/issues/171)
     const projectId = projectKeypair.publicKey.toString('hex')
 
-    this.#db
-      .insert(projectKeysTable)
-      .values({
-        projectId,
-        keysCipher: this.#keyManager.encryptLocalMessage(
-          Buffer.from(ProjectKeys.encode(keys).finish().buffer),
-          projectId
-        ),
-      })
-      .run()
+    this.#saveProjectKeys(projectId, keys)
 
     // 4. Create MapeoProject instance
     const project = new MapeoProject({
@@ -131,8 +150,9 @@ export class MapeoManager {
       throw new Error(`NotFound: project ID ${projectId} not found`)
     }
 
-    const projectKeys = ProjectKeys.decode(
-      this.#keyManager.decryptLocalMessage(result.keysCipher, projectId)
+    const projectKeys = this.#decodeProjectKeysCipher(
+      result.keysCipher,
+      projectId
     )
 
     const project = new MapeoProject({
@@ -160,5 +180,54 @@ export class MapeoManager {
       .from(projectTable)
       .all()
       .map((value) => deNullify(value))
+  }
+
+  /**
+   * @param {string} projectId
+   * @returns {Promise<MapeoProject>}
+   */
+  async addProject(projectId) {
+    const activeProject = this.#activeProjects.get(projectId)
+
+    if (activeProject) return activeProject
+
+    const existingKeysResult = this.#db
+      .select({
+        keysCipher: projectKeysTable.keysCipher,
+      })
+      .from(projectKeysTable)
+      .where(eq(projectKeysTable.projectId, projectId))
+      .get()
+
+    /** @type {ProjectKeys} */
+    let projectKeys
+
+    if (existingKeysResult) {
+      projectKeys = this.#decodeProjectKeysCipher(
+        existingKeysResult.keysCipher,
+        projectId
+      )
+    } else {
+      projectKeys = {
+        projectKey: Buffer.from(projectId, 'hex'),
+        encryptionKeys: {
+          // TODO: How do we get the auth key in this case?
+          auth: randomBytes(32),
+        },
+      }
+      this.#saveProjectKeys(projectId, projectKeys)
+    }
+
+    const project = new MapeoProject({
+      ...projectKeys,
+      storagePath: this.#storagePath,
+      keyManager: this.#keyManager,
+      sharedDb: this.#db,
+      sharedIndexWriter: this.#projectSettingsIndexWriter,
+    })
+
+    this.#activeProjects.set(projectId, project)
+
+    return project
   }
 }
--- a/src/mapeo-manager.js
+++ b/src/mapeo-manager.js
@@ -205,15 +205,14 @@ export class MapeoManager {
       throw new Error(`Project with ID ${projectId} already exists`)
     }
 
-    const projectKeys = {
-      projectKey: Buffer.from(projectId, 'hex'),
+    this.#saveProjectKeys(projectId, {
+      projectKey,
       encryptionKeys,
-    }
-
-    this.#saveProjectKeys(projectId, projectKeys)
+    })
 
     const project = new MapeoProject({
-      ...projectKeys,
+      projectKey,
+      encryptionKeys,
       storagePath: this.#storagePath,
       keyManager: this.#keyManager,
       sharedDb: this.#db,
--- a/src/mapeo-manager.js
+++ b/src/mapeo-manager.js
@@ -137,11 +137,27 @@ export class MapeoManager {
    * @returns {Promise<MapeoProject>}
    */
   async getProject(projectId) {
-    const existing = this.#activeProjects.get(projectId)
+    // 1. Check for existing active project
+    const activeProject = this.#activeProjects.get(projectId)
 
-    if (existing) return existing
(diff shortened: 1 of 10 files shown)


src/mapeo-manager.js
```diff
@@ -214,4 +214,40 @@ export class MapeoManager {
...
   }
+
+  /**
+   * @param {string} projectId
+   * @returns {Promise<MapeoProject>}
+   */
+  async addProject(projectId) {
+    const existing = this.#activeProjects.get(projectId)
+
+    if (existing) return existing
```
achou11, line 225: wondering about this return type a bit. should it return the project instance in general or some boolean about success, or void?

secondly, depending on the answer to this first question, does it make sense to return the existing project instance or do something else like throw an error or noop?

gmaclennan (commented):
Somewhere along the way the API docs got out of date.

`addProject()` is what we would use when receiving an invite, so the param should be the same type as an invite e.g.

```ts
addProject(invite: {
  projectKey: Buffer,
  encryptionKeys: {
    auth: Buffer;
    data?: Buffer | undefined;
    config?: Buffer | undefined;
    blobIndex?: Buffer | undefined;
    blob?: Buffer | undefined;
  },
  projectInfo: { name?: string | undefined }
}
```

Not sure what to do with projectInfo - once we sync we will get the project info from the replicated cores, but in the meantime it would be good to use the name passed here. Suggestions?

gmaclennan:
If the project you are trying to add already exists then I think maybe throw? And maybe return the `projectId` the same as createProject?

gmaclennan:
Also the proto definitions for Invite need updated - encryptionKeys should be required, since we always need an auth core encryption key. I'll do that now.

achou11:
> Somewhere along the way the API docs got out of date.
>
> addProject() is what we would use when receiving an invite, so the param should be the same type as an invite 

Ah this make sense. Figured that something about the method params would need to be updated. I have an issue in the core docs repo (https://github.com/digidem/mapeo-core-docs/issues/14) to track changes that need to be made there as a result of the implementation changes we do here, in case that's helpful (I already added it as an item there to address later)

> Not sure what to do with projectInfo - once we sync we will get the project info from the replicated cores, but in the meantime it would be good to use the name passed here. Suggestions?

not sure what suggestions you're looking for, but seems like it would make sense to use whatever's passed there to update the settings of a project via `project.$setProjectSettings`

achou11:
> Update listProjects implementation to read from the projectKeys table as the source of truth for what projects exist, and cross reference with queried project table to get desired information for return value (no sql join necessary for now)

One question I have from attempting to implement this: let's say we have the following situation:

1. I _create_ a project
2. I _add_ a project (but haven't synced anything)
3. I list all projects

How many projects are returned in step 3? In step 2, I update the `projectKeys` table but not the `project` table with the added project. When I read the `project` table in step 3, there's no information about the added project and thus I can't derive any `createdAt` or `updatedAt` values for the return result. My assumption is that we only want to list projects that have definitive `createdAt` and `updatedAt` dates, meaning that the added project would _*not*_ be listed as a result. Is that desirable? Probably ties in with some UX questions

achou11:
Upon some additional thinking, I convincing myself that "just-added" projects shouldn't be listed in `listProjects`. Reasoning is that including it would entice the end-user to try to use it via a `MapeoProject` instance, which won't work in that state since it doesn't fully exist in the local config core / project database.

achou11:
more braindumping as I try to reason about what makes sense here. At the app level, something like this will happen:

1. I get invited to a project
2. I accept invite, which calls `manager.addProject` under the hood
3. I wait for peer who invited me to acknowledge me accepting their invite, which they then officially add me to the project on their end and enable the ability to sync the full project details over
4. I wait for project details to be synced (enough?)
5. After some point, I have what I need to properly interact with the project

Between 2 and 3, does it make sense for me to be able to see any information about that project, even though I haven't been officially acknowledged? basically, after step 2, if I went to some UI that listed the projects that I'm a part of, does the project I just accepted an invite for really count, or do I need to wait for acknowledge + syncing to fall under that classification? Maybe it counts, but under some "pending" state (playing into the `addedAt` variant I mentioned before for listing projects)

achou11:
Alright discussion with Gregor surfaced the following:

- we need to add a download step for the auth + config cores in the `addProject` implementation before persisting to the project keys database, this way we'll have a usable project.

- in the meantime, we should update the return type of `listProjects` to make `createdAt` and `updatedAt` optional fields, which applies for added but not-yet-synced projects

- we need to update the `MapeoProject` class in the following ways (out of scope for this PR):
  - pass in a MDNS discovery instance. this instance maintains a list of active connections
  - add a method on the `MapeoProject` instance to download the auth + config cores (uses the MDNS instance passed) 
    - this method would eventually be used by the `addProject` method on the client

test-e2e/manager-basic.js
```diff
@@ -13,17 +14,49 @@ test('Managing multiple projects', async (t) => {
...
-  t.is(allProjects.length, createdProjectIds.length)
-  t.ok(
-    allProjects.every((p) => createdProjectIds.includes(p.projectId)),
-    'all created projects are listed'
+  const listedProjects = await manager.listProjects()
+
+  t.is(listedProjects.length, 2)
+
+  t.is(
+    listedProjects[0].projectId,
```
gmaclennan, line 32: Why would this always be the first entry in this array? Should it be? Or should we just test that one of the entries has this projectId? If we are sorting by date then we can't guarantee this will always be first, if they are both written in the same millisecond.

src/schema/client.js
```diff
@@ -4,11 +4,12 @@
...
 import { jsonSchemaToDrizzleColumns as toColumns } from './schema-to-drizzle.js'
-import { backlinkTable } from './utils.js'
+import { backlinkTable, customJson } from './utils.js'
 
 export const projectTable = sqliteTable('project', toColumns(schemas.project))
 export const projectBacklinkTable = backlinkTable(projectTable)
 export const projectKeysTable = sqliteTable('projectKeys', {
   projectId: text('projectId').notNull().primaryKey(),
   keysCipher: blob('keysCipher', { mode: 'buffer' }).notNull(),
+  projectInfo: customJson('projectInfo').default(JSON.stringify({})),
```
gmaclennan, line 14:
```suggestion
  projectInfo: customJson('projectInfo').default({}),
```

src/schema/utils.js
```diff
@@ -30,3 +35,16 @@ export function backlinkTable(tableSchema) {
...
+    return 'text'
+  },
+  fromDriver(value) {
+    // @ts-ignore
+    return JSON.parse(value)
+  },
+  toDriver(value) {
+    return JSON.stringify(value)
+  },
+})
```
gmaclennan, line 50:
```suggestion
export const customJson =
  /** @type {ReturnType<typeof customType<{ data: import('../generated/rpc.js').Invite_ProjectInfo }>>} */ (
    customType({
      dataType() {
        return 'text'
      },
      fromDriver(value) {
        // @ts-ignore
        return JSON.parse(value)
      },
      toDriver(value) {
        return JSON.stringify(value)
      },
    })
  )
```

gmaclennan (approved):
Looks good. A question on the tests, and a couple of suggestions to improve types

src/schema/client.js
achou11: seems like there's a bug in Drizzle where this ideally would work but doesn't. i tried this in https://github.com/digidem/mapeo-core-next/pull/215/commits/bfe7c0f6f7cabf697d805781a3caaa27904e7096 but if you see the sql it generates, it doesn't do what you'd expect. have to keep the stringify here for now

test-e2e/manager-basic.js
achou11: there's no sorting that's happening at the moment. think we can decide on that in a separate PR. will update the test to just look for the id rather than assuming ordering 👍

src/schema/utils.js
achou11: this is used for other schema things so would have to work with a generic instead

https://github.com/digidem/mapeo-core-next/blob/209/client-add-project/src/schema/schema-to-drizzle.js#L59

https://github.com/digidem/mapeo-core-next/blob/209/client-add-project/src/schema/schema-to-drizzle.js#L81

test-e2e/manager-basic.js
achou11: addressed via https://github.com/digidem/mapeo-core-next/pull/215/commits/3fd174dfd1be6318719c444a062146fa5d862e12

src/schema/utils.js
achou11: can't figure out how to do this with generics so going to avoid diving into a rabbit hole and keep my ugly runtime checks for now 😄

src/schema/client.js
```diff
@@ -4,11 +4,12 @@
 import { blob, sqliteTable, text } from 'drizzle-orm/sqlite-core'
 import { dereferencedDocSchemas as schemas } from '@mapeo/schema'
 import { jsonSchemaToDrizzleColumns as toColumns } from './schema-to-drizzle.js'
-import { backlinkTable } from './utils.js'
+import { backlinkTable, customJson } from './utils.js'
```
gmaclennan, line 7:
```suggestion
import { backlinkTable, customJson } from './utils.js'

const projectInfoColumn = /** @type {ReturnType<typeof customType<{ data: import('../generated/rpc.js').Invite_ProjectInfo }>>} */ (customJson)
```

achou11: lovely, thanks!

achou11: addressed via https://github.com/digidem/mapeo-core-next/pull/215/commits/f83d914545462d7804bdff0b4df62951328f48eb

src/schema/utils.js
achou11: addressed via https://github.com/digidem/mapeo-core-next/pull/215#discussion_r1309172868

PR #223, opened 2023-08-24
fix: invite.encryptionKeys should be required

gmaclennan:
This unfortunately requires more than just an update to the proto definition because ts-proto [does not support making messages required](https://github.com/stephenh/ts-proto#current-status-of-optional-values). This means we need to check that ourselves and throw / ignore invites if missing.

There also seem to be a few type errors in src/rpc.js that need fixing. Going to keep this PR open to address those issues.


achou11:
the type error that stands out to me is the one regarding conflicting type definitions of the transformer and the protobuf interface that ts-proto outputs, basically what's described in https://github.com/stephenh/ts-proto/issues/881. The reasoning in that issue makes sense if you're strictly working with TS files, but in our case we're not so attempting to create a JSDoc typedef out of convenience doesn't work unless:

1. you rename the import of the transformer e.g. `import { Invite as InviteTransformer } from '../generated/rpc.js'`
2. avoid creating a typedef and inlining the `import(...)` bit each time you want to reference it in jsdoc
3. name the created typedef something different

a quick glance at the ts-proto options didn't surface anything immediately helpful, but maybe with fresher eyes there will be something that works. otherwise, gotta pick our poison with one of the above options I guess 😄 

Another thing I was thinking is that maybe we should create another directory `e.g. src/transformers` where we can use that as the source of imports for anything related to the proto stuff. for example, the rpc one would be similar to the below and any code that needs to work withe proto stuff would import from here instead of `src/generated`:

```ts
import {
  Invite as InviteTransformer,
  InviteResponse as InviteResponseTransformer,
  InviteResponse_Decision as InviteResponse_DecisionTransformer,
} from '../generated/rpc.js'

/** @typedef {import('../generated/rpc.js').Invite} Invite */

export {
  InviteTransformer,
  InviteResponseTransformer,
  InviteResponse_DecisionTransformer,
}
```

yes, it's boilerplate-y and somewhat verbose, but this approach would also provide a location for us to override/customize the encode/decode functions, which we may need to do based on Gregor's assessment of the original issue

tests/rpc.js
```diff
@@ -103,7 +126,7 @@ test('Send invite and already on project', async (t) => {
   replicate(r1, r2)
 })
 
-test('Send invite with encryption key', async (t) => {
+test('Send invite with encryption keys', async (t) => {
```
achou11, line 129: perhaps this test should be renamed, since it used to assume that the providing some encryption keys was optional

src/transformers/rpc.js
```diff
@@ -0,0 +1,56 @@
...
+// and we can't add the assert annotation directly here because this needs to return a value
+/**
+ * @param {import('../generated/rpc.js').IInvite} message
+ * @param {Parameters<typeof Invite['encode']>[1]} [writer]
+ * @returns {ReturnType<typeof Invite['encode']>}
+ */
+function inviteEncode(message, writer) {
+  inviteValidate(message)
+  return Invite.encode(message, writer)
+}
```
achou11, line 32: guess the assert is more useful if this function were called in contexts where the input is not controlled by us, which isn't the case at the moment. still, an inconvenient behavior of TS

src/transformers/rpc.js
achou11, line 1: the editor experience with this isn't as ideal as I'd want it to be (the redefined interfaces aren't always expanded to the full definition), so just recognizing that with this comment 😄

scripts/build-messages.js
```diff
@@ -21,3 +22,64 @@ console.log(command2)
...
+
+await fixTsProtoOutput(outPath)
+
+// ts-proto exports the proto message declaration for both the transformer and the interface
+// which causes problems for TS when it comes to resolving type imports.
+// This fixes the output by:
+// 1. Renaming the exported interfaces so that they're prefixed with `I`
+// 2. Updating generated TS source files to re-import relevant transformers (since it gets lost in the renaming step)
+// https://github.com/digidem/mapeo-core-next/pull/223#issuecomment-1709125859
+async function fixTsProtoOutput(directoryPath) {
```
achou11, line 34: happy to explain this in more detail if it's not clear enough from the comments

achou11 (commented):
Not sure if we wanted also add changes to handle/ignore validation errors from the encode/decode calls so I haven't updated relevant callsites yet.

achou11:
@gmaclennan would you mind taking a look at the recent changes I added? To summarize:

- generally fixed the types issues that you alluded to. involved using https://github.com/dsherret/ts-morph after ts-proto does its thing to resolve output issues that aren't fixable with ts-proto

- adds a "transformers" directory that's meant to be a home for adjusting and/or re-exporting generated exports from ts-proto. only does `rpc` for now since that's relevant for the original issue

gmaclennan:
Replaced by #260

PR #229, opened 2023-08-28
feat: CoreOwnership class w getOwner & getCoreKey

gmaclennan:
Stacked on #214

- also writeOwnership

Required the addition of an internal `select()` method for dataType,
which directly accesses drizzle.select().from(table)


src/core-ownership.js
```diff
@@ -1,13 +1,83 @@
...
+  /**
+   *
+   * @param {import('./types.js').KeyPair} identityKeypair
+   * @param {Record<typeof NAMESPACES[number], import('./types.js').KeyPair>} coreKeypairs
+   */
+  async writeOwnership(identityKeypair, coreKeypairs) {
+    /** @type {import('./types.js').CoreOwnershipWithSignaturesValue} */
+    const docValue = {
+      schemaName: 'coreOwnership',
+      ...mapObject(coreKeypairs, (key, value) => {
```
achou11, line 65: no suggestion here but was interested writing a limited version of this in case it would be useful to vendor since we're not using any of `map-obj`'s more advanced options in this case.

Seems to work (enough 😄 )

```ts
/**
 * @template {string} SourceKey
 * @template {Record<SourceKey, any>} Source
 * @template {string} TargetKey
 * @template TargetValue
 *
 * @param {Source} source
 * @param {(key: keyof Source, value: Source[keyof Source]) => [TargetKey, TargetValue]} mapper
 *
 * @returns {Record<TargetKey, TargetValue>}
 */
function mapObj(source, mapper) {
  const result = /** @type {Record<TargetKey, TargetValue>} */ ({})

  for (const [key, value] of Object.entries(source)) {
    const [newKey, newValue] = mapper(/** @type {keyof Source} */ (key), value)
    result[newKey] = newValue
  }

  return result
}
```

achou11 (approved):

src/core-ownership.js
gmaclennan: It's a lot to vendor - would need to vendor all the type tests as well as the regular tests. I think sindre's modules are fairly reliable, I think it's ok to add as a dependency. I recently discovered `map-obj` and there are places where it would be better to use than what I have written in the past, and be more type-safe.

PR #230, opened 2023-08-28
feat: coreOwnership integration [3/3]

gmaclennan:
Stacked on #229

Add an instance of CoreOwnership to the MapeoProject instance and
write a core ownership record when the project is first created


src/core-manager/index.js
```diff
@@ -135,12 +136,25 @@ export class CoreManager extends TypedEmitter {
         this.#handleExtensionMessage(data, peer)
       },
     })
+
+    this.#ready = Promise.all(
+      [...this.#coreIndex].map(({ core }) => core.ready())
+    ).catch(() => {})
```
achou11, line 142: what happens if we try to use a core that failed to become ready later on? will we get a clear enough error during that time? wondering if it would make sense to do some handling here to catch earlier

achou11 (approved):
have a non-blocking question but otherwise lgtm

src/core-manager/index.js
gmaclennan: If a core failed to become ready then `coreManager.ready()` would reject.
If you try to use a core that failed to become ready then it would throw an error.
I hope the error we would get would be clear enough

src/mapeo-project.js
```diff
@@ -220,3 +279,29 @@ function extractEditableProjectSettings(projectDoc) {
...
+  for (const namespace of NAMESPACES) {
+    keypairs[namespace] =
+      namespace === 'auth' && projectSecretKey
+        ? { publicKey: projectKey, secretKey: projectSecretKey }
+        : keyManager.getHypercoreKeypair(namespace, projectKey)
+  }
+  return /** @type {Record<import('./core-manager/core-index.js').Namespace, import('./types.js').KeyPair>} */ (
+    keypairs
+  )
+}
```
achou11, line 307: not sure if any better from a type perspective but for any function that does casting internally, I typically prefer to add an an explicit `returns` annotation in the function definition:
 
```suggestion
/**
 * Return a map of namespace -> core keypair
 *
 * For the project owner the keypair for the 'auth' namespace is the projectKey
 * and projectSecretKey. In all other cases keypairs are derived from the
 * project key
 *
 * @param {object} opts
 * @param {Buffer} opts.projectKey
 * @param {Buffer} [opts.projectSecretKey]
 * @param {import('@mapeo/crypto').KeyManager} opts.keyManager
 * @returns {Record<import('./core-manager/core-index.js').Namespace, import('./types.js').KeyPair>}
 */
function getCoreKeypairs({ projectKey, projectSecretKey, keyManager }) {
  const keypairs =
    /** @type {Record<import('./core-manager/core-index.js').Namespace, import('./types.js').KeyPair>} */ ({})

  for (const namespace of NAMESPACES) {
    keypairs[namespace] =
      namespace === 'auth' && projectSecretKey
        ? { publicKey: projectKey, secretKey: projectSecretKey }
        : keyManager.getHypercoreKeypair(namespace, projectKey)
  }

  return keypairs
}
```

Suggestion diff isn't clear but basically:

- add return type in JSDoc definition
- add type cast on `keyPairs` variable declaration
- remove type cast from return value

achou11 (approved):
1 non-blocking suggestion but otherwise 👍

PR #231, opened 2023-08-28
feat: add capabilities

gmaclennan:
Implements #189

More complete tests are blocked by the ability to invite and sync devices, tracking that in #238 

This PR adds an internal `this.#capabilities` instance to the MapeoProject instance. It's initial use will be for assigning roles (part of the invite process) and for checking capabilities for sync when syncing with other devices. Checking capabilities for reading/writing mapeo docs is post-MVP. Also need to expose in the API, tracking that in #239

Post-MVP we will implement the ability have custom roles and capabilities, but hard-coding is ok for now.

Issue #225, opened 2023-08-24
Add `$member.invite()` method on MapeoProject


achou11:
Towards to #133

Approximate documentation lives here: https://github.com/digidem/mapeo-core-docs/blob/main/client-docs/README.md#memberinvite

Tasks: 

- [x] Implement MemberAPI class with invite method
- [x] Integrate into MapeoProject as `$member` namespace

PR #232, opened 2023-08-28
initial implementation of invite method for MemberApi

achou11:
Towards #225 

- Introduces the `MemberApi` class
- Implements the `invite` method

TODO:

- [X] What is `role` here and how does it work with MapeoRPC? Our original documentation suggests that it's some human-readable string, but given the recent work on capabilities, wondering if that changes to a role id or something like that. EDIT: answered in https://github.com/digidem/mapeo-core-next/pull/232#discussion_r1310386163
- [ ] Does this class need to keep track of pending invites? I know MapeoRPC does that internally but how about at this level?
- [ ] What happens when you send multiple invites to the same device id?
- [x] integrate capabilities to write the capability when invite accepted #231

### DIFF
--- a/src/member-api.js
+++ b/src/member-api.js
@@ -1,6 +1,8 @@
 import { TypedEmitter } from 'tiny-typed-emitter'
+import { InviteResponse_Decision } from './generated/rpc.js'
 
 export class MemberApi extends TypedEmitter {
+  #capabilities
   #encryptionKeys
   #getProjectInfo
   #projectKey
@@ -8,29 +10,37 @@ export class MemberApi extends TypedEmitter {
 
   /**
    * @param {Object} opts
+   * @param {import('./capabilities.js').Capabilities} opts.capabilities
    * @param {import('./generated/keys.js').EncryptionKeys} opts.encryptionKeys
    * @param {() => Promise<import('./generated/rpc.js').Invite_ProjectInfo>} opts.getProjectInfo
    * @param {Buffer} opts.projectKey
    * @param {import('./rpc/index.js').MapeoRPC} opts.rpc
    */
-  constructor({ encryptionKeys, getProjectInfo, projectKey, rpc }) {
+  constructor({
+    capabilities,
+    encryptionKeys,
+    getProjectInfo,
+    projectKey,
+    rpc,
+  }) {
     super()
     this.#encryptionKeys = encryptionKeys
     this.#getProjectInfo = getProjectInfo
     this.#projectKey = projectKey
     this.#rpc = rpc
+    this.#capabilities = capabilities
   }
 
   /**
    * @param {string} deviceId
    *
    * @param {Object} opts
-   * @param {string} opts.roleId
+   * @param {import('./capabilities.js').RoleId} opts.roleId
    * @param {number} [opts.timeout]
    *
    * @returns {Promise<import('./generated/rpc.js').InviteResponse_Decision>}
    */
-  async invite(deviceId, { timeout }) {
+  async invite(deviceId, { roleId, timeout }) {
     const projectInfo = await this.#getProjectInfo()
 
     const response = await this.#rpc.invite(deviceId, {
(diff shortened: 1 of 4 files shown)


tests/member-api.js
```diff
@@ -0,0 +1,39 @@
+import { test } from 'brittle'
+import { randomBytes } from 'crypto'
+import { KeyManager } from '@mapeo/crypto'
+
+import { MapeoRPC } from '../src/rpc/index.js'
+import { MemberApi } from '../src/member-api.js'
+import { replicate } from './helpers/rpc.js'
+
+test('Invite sends expected project-related details', async (t) => {
```
achou11, line 9: this test kind of overlaps with some of the rpc unit tests. okay for now since I just want something here until i figure out what else to test

src/member-api.js
```diff
@@ -0,0 +1,44 @@
...
+    this.#getProjectInfo = getProjectInfo
+    this.#projectKey = projectKey
+    this.#rpc = rpc
+  }
+
+  /**
+   * @param {string} deviceId
+   *
+   * @param {Object} opts
+   * @param {string} opts.role
```
achou11, line 28: what is this and how should it be used with MapeoRPC?

achou11: how does the device being invited know what role they will be assigned prior to responding? Shouldn't the call to `rpc.invite()` also accept some param about the role to send to the target peer?

My understanding is that from a UX perspective, someone with an invite should see something like "Bob has invited you to be _ROLE_ in _PROJECT NAME_"

achou11: param name addressed via https://github.com/digidem/mapeo-core-next/pull/232/commits/88a3ef4cfeefc90adabf4ec6c38ab1d3fc952cd6

tests/member-api.js
```diff
@@ -0,0 +1,39 @@
...
+  r1.on('peers', (peers) => {
+    memberApi.invite(peers[0].id, {
+      role: 'member',
+    })
+  })
+
+  r2.on('invite', async (_peerId, invite) => {
+    t.alike(invite.projectKey, projectKey)
+    t.alike(invite.encryptionKeys, encryptionKeys)
+    t.alike(invite.projectInfo, projectInfo)
```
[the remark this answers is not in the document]
achou11: addressed via https://github.com/digidem/mapeo-core-next/pull/232/commits/fcb1a085330c1a7d9c628322f35e06d80ef9c80d

src/member-api.js
achou11: role assignment addressed via https://github.com/digidem/mapeo-core-next/pull/232/commits/e26ff99d4f47f3a77cd002df75ef2b31d655e987

tests/member-api.js
```diff
@@ -0,0 +1,143 @@
...
+    r2.inviteResponse(peerId, {
+      projectKey: invite.projectKey,
+      decision: InviteResponse_Decision.ACCEPT,
+    })
+  })
+
+  replicate(r1, r2)
+})
+
+test('invite() does not assign role to invited device if invite is not accepted', async (t) => {
```
achou11, line 96: maybe testing implementation details too much?

tests/member-api.js
```diff
@@ -0,0 +1,143 @@
...
+    r2.inviteResponse(peerId, {
+      projectKey: invite.projectKey,
+      decision: InviteResponse_Decision.ACCEPT,
+    })
+  })
+
+  replicate(r1, r2)
+})
+
+test('invite() assigns role to invited device after invite accepted', async (t) => {
```
achou11, line 50: maybe testing implementation details too much?

achou11 (commented):
added the role assignment bit. wanted to have some tests to cover that part but they may be diving into implementation details too much. perhaps better to write an e2e for this?

gmaclennan (approved):
Looks good to me, good work on this. I think tests are ok - we can eventually replace with e2e tests but at this stage these tests are helpful for making sure the internals of this are working as expected as we build upon it.

PR #235, opened 2023-08-29
chore: adjust storage options for MapeoManager and MapeoProject

### LINKED ISSUE #226: Fix storage options for MapeoManager and MapeoProject
## Description

Currently there are a few issues with how we define storage paths:

- Because we pass a storagePath instead of a storage function, there is no way in tests to set up re-usable memory storage.
- Currently project storage is not namespaced, so the project database will be overwritten by each project
- The filepool is set-up for each project instance. There should be one file pool for all storage across all projects

Switching to accepting a storage path or a storage function (similar to what Hypercore does) would allow us to pass `() => new RAM()` as storage, and persist RAM storage where needed for tests. It would also remove all the boilerplate in the code for setting up memory storage if storagePath is undefined

We want storage for each project to be stored in a separate folder. The folder name should be the projectPublicId. This means we will need to wrap the storage function that is passed to MapeoProject to append the projectId to the name - see [how corestore does this.](https://github.com/holepunchto/corestore/blob/a68c20c8aed0660540c5824e7f8930f4a8844af5/index.js#L251)

We would still need a path/folder for sqlite database files. It probably makes sense to pass a `dbPath` to MapeoProject and `dbFolder` to MapeoClient. `client.db`. We could store all databases in the same folder, if we name the project database files with the projectPublicId. This is probably be easier than trying to place the db file in folders alongside the hypercore storage / random-access-storage.

## Tasks

- [ ] Add `dbFolder` required option to MapeoManager
- [ ] Make `opts.storage` a required option, and allow a storage creation function (`name => new RandomAccessStorage(name)`)
- [ ] If a path is passed, use a [default random-access-file storage function](https://github.com/digidem/mapeo-core-next/blob/main/src/mapeo-project.js#L74-L78)
- [ ] Wrap the storage function passed to MapeoProject so that cores for each project are stored in separate folders named with the project public ID.
- [ ] Inside MapeoProject, the storage function will still need to be wrapped to separate index storage from core storage.
- [ ] Create a single instance of RandomAccessFilePool in MapeoManager and use that for all storages
- [ ] Remove boilerplate for handling undefined opts.storage and setting up memory storage in the module code
- [ ] Update tests to setup their own memory storage
- [ ] Add a `dbPath` required option to MapeoProject
- [ ] If `dbFolder` is `:memory:`, then pass `:memory:` as opts.dbPath to MapeoProject
- [ ] Otherwise `clientDbPath = path.join(opts.dbFolder, 'client.db')` and `new MapeoProject({ dbPath: path.join(dbFolder, projectId) })`

achou11:
Closes #226 

Tried my best to follow the checklist but there were some uncertainties, mostly around handling the `storage` opt in `MapeoProject`


src/mapeo-project.js
```diff
@@ -35,57 +28,58 @@ export class MapeoProject {
...
-      storagePath !== undefined
+      typeof storage === 'string'
         ? (name) =>
             new RandomAccessFile(
-              path.join(storagePath, INDEXER_STORAGE_FOLDER_NAME, name),
+              path.join(dbPath, storage, INDEXER_STORAGE_FOLDER_NAME, name),
               { pool: filePool }
             )
-        : () => new RAM()
+        : (name) => storage(path.join(INDEXER_STORAGE_FOLDER_NAME, name))
```
achou11, line 82: > If a path is passed, use a [default random-access-file storage function](https://github.com/digidem/mapeo-core-next/blob/main/src/mapeo-project.js#L74-L78)

>  Inside MapeoProject, the storage function will still need to be wrapped to separate index storage from core storage.

Made the guess that these points applied here, but could be mistaken

src/mapeo-project.js
```diff
@@ -35,57 +28,58 @@ export class MapeoProject {
...
-   * @param {string} [opts.storagePath] Folder for all data storage (hypercores and sqlite db). Folder must exist. If not defined, everything is stored in-memory
+   * @param {string} opts.dbPath Folder for all data storage (hypercores and sqlite db). Folder must exist. Use ':memory:' to store everything in-memory
    * @param {import('@mapeo/crypto').KeyManager} opts.keyManager mapeo/crypto KeyManager instance
    * @param {Buffer} opts.projectKey 32-byte public key of the project creator core
    * @param {Buffer} [opts.projectSecretKey] 32-byte secret key of the project creator core
    * @param {Partial<Record<import('./core-manager/index.js').Namespace, Buffer>>} [opts.encryptionKeys] Encryption keys for each namespace
    * @param {import('drizzle-orm/better-sqlite3').BetterSQLite3Database} opts.sharedDb
    * @param {IndexWriter} opts.sharedIndexWriter
+   * @param {import('./types.js').StorageParam} opts.storage Folder to store all hypercore data
+   * @param {import('./core-manager/random-access-file-pool.js').RandomAccessFilePool} [opts.filePool] FilePool instance to use if opts.storage is a path
```
achou11, line 39: > Create a single instance of RandomAccessFilePool in MapeoManager and use that for all storages

is this aligned with what you described here? wasn't explicitly noted to create an opt for MapeoProject but seemed to make sense given the changes required by other points

src/mapeo-manager.js
```diff
@@ -43,8 +53,26 @@ export class MapeoManager {
...
+      dbFolder !== ':memory:'
+        ? new RandomAccessFilePool(MAX_FILE_DESCRIPTORS)
+        : undefined
+
+    this.#createProjectStorageOpts =
+      dbFolder === ':memory:'
+        ? () => ({
+            dbPath: ':memory:',
+            storage: () => new RAM(),
+          })
```
achou11, line 68: have a feeling that this shouldn't be a function that creates a new RAM instance each time a project is created. is the idea to share the ram instance across all projects too?

gmaclennan:
Not quite what I meant, sorry the issue wasn't clear. I pushed https://github.com/digidem/mapeo-core-next/pull/235/commits/28d4f8963366270fdab70cf2869f077f4c0ebf75 to show what I meant. I think this is done now, can you review?

gmaclennan:
Should merge and release https://github.com/digidem/multi-core-indexer/pull/17 before merging, so that we can update the snapshot and fix the storage locations

gmaclennan (changes requested):
I made some changes and added a test to catch breaking changes to storage locations. There is a bug fix here for something unrelated (encoding project keys). There should be a failing test for that, in a separate PR.

gmaclennan (approved):

achou11:
your changes lgtm! thanks for making the fixes - was hoping to do this to take something off your plate but ended up with you doing most of the work anyways 😄

PR #242, opened 2023-08-31
return content hash from blobApi.create

### LINKED ISSUE #220: Store blob hashes in attachment records
## Description

A hash of blobs attached to observations and other documents would allow an extra check for tampering when an observation is shared outside of Mapeo: the signed record written to the data core store would have attachments which referenced the hash, so you can know that an image/video/audio file that was shared has not been changed since the observation was created by verifying the hash.

Putting this as MVP, because it affects the data structure and it will be easier to always assume that attachments have hashes, rather than deal with some without in the future.

## Implementation proposal

Using the `BlobId.name` as the hash is one option, but it would either require an extra step when writing to the blob store (hash first, then write), or would require us to write our own implementation of [`hyperdrive.createWriteStream`](https://github.com/holepunchto/hyperdrive/blob/81fe0cdfa72fb7620a2b5a454e61484ed72ca32e/index.js#L434) to hash during write and use that as the name.

An easier-to-implement option is to return the hash from `blobApi.create()` and store that as part of the attachmentId.

For some blobs, e.g. photos, we store several variants - lower resolution versions. The hash we store should be of the original version.

We [should use a sha256 hash](https://security.stackexchange.com/questions/198631/which-hashing-algorithm-shoud-i-use-for-a-safe-file-checksum).

## Tasks

- [ ] Update the schema for observation attachments to include a hash
- [ ] Return the hash from `blobApi.create()`

sethvincent:
Updates blob-api.js and relevant tests to return a content hash from `blobApi.create`.

closes #220.

There are a couple ts-ignore comments with todos that could be done in this pr, though maybe shouldn't be blockers. Especially the pipeline-related one.

### DIFF
--- a/src/blob-api.js
+++ b/src/blob-api.js
@@ -1,5 +1,6 @@
 import fs from 'node:fs'
 import { pipeline } from 'node:stream/promises'
+import { createHash } from 'node:crypto'
 import sodium from 'sodium-universal'
 import b4a from 'b4a'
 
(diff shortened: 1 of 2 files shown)


achou11 (approved):
noice!

PR #243, opened 2023-08-31
blobStore.createReadStream should not wait

### LINKED ISSUE #80: blobStore.createReadStream() should not "wait"
## Description

Currently the blobStore.createReadStream() method will wait for a blob to become available if it is not downloaded.

## Tasks

- [ ] Add `wait` and `timeout` options to blobStore.createReadStream()

sethvincent:
This adds `wait` and `timeout` options to `blobStore.createReadStream`.

It adds those options to `blobStore.createEntryReadStream` as well.

closes #80

### DIFF
--- a/src/blob-store/index.js
+++ b/src/blob-store/index.js
@@ -119,7 +119,10 @@ export class BlobStore {
    * @param {boolean} [options.wait=false] Set to `true` to wait for a blob to download, otherwise will throw if blob is not available locally
    * @param {number} [options.timeout] Optional timeout to wait for a blob to download
    */
-  createReadStream({ type, variant, name, driveId }, options = { wait: false }) {
+  createReadStream(
+    { type, variant, name, driveId },
+    options = { wait: false }
+  ) {
     // TODO: Error thrown from this be an emit error on the returned stream?
     const drive = this.#getDrive(driveId)
     const path = makePath({ type, variant, name })
@@ -216,7 +219,7 @@ export class BlobStore {
    * @param {BlobId} blobId
    * @param {object} [options]
    * @param {boolean} [options.diff=false] Enable to return an object with a `block` property with number of bytes removed
-   * @return {Promise<any>}
+   * @return {Promise<{ blocks: number } | null>}
    */
   async clear({ type, variant, name, driveId }, options = {}) {
     const path = makePath({ type, variant, name })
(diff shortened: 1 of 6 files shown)


src/blob-store/index.js
```diff
@@ -204,6 +211,19 @@ export class BlobStore {
     const entry = await drive.entry(path, options)
     return entry
   }
+
+  /**
+   * @param {BlobId} blobId
+   * @param {object} [options]
+   * @param {boolean} [options.diff=false] Enable to return an object with a `block` property with number of bytes removed
+   * @return {Promise<any>}
```
achou11, line 219: any specific reason for using `any` here? may be able to omit this line entirely and rely on inference in this case, since `drive.clear` goes about 3 levels of deps deep 😅

tests/blob-server.js
```diff
@@ -215,6 +217,52 @@ test('GET photo returns 404 when trying to get non-replicated blob', async (t) =
...
+      method: 'GET',
+      url: buildRouteUrl({ ...blobId, projectId, driveId: blobStore.writerDriveId }),
+    })
+
+    t.is(res.statusCode, 404)
+  }
+
+  const driveId = await blobStore.put(blobId, expected)
+
+  await blobStore.put(blobId, expected)
```
achou11, line 246: is this additional call to `put()` needed?

tests/blob-server.js
```diff
@@ -215,6 +217,52 @@ test('GET photo returns 404 when trying to get non-replicated blob', async (t) =
   t.is(res.statusCode, 404)
 })
 
+test('GET photo returns 404 when trying to get non-existent blob', async (t) => {
```
achou11, line 220: maybe out of scope but would be helpful to ensure that the reasons for the 404s align with our expectations. right now this test only checks the status code but maybe in the short term, check the status message as well?

more futureproof solution would be to define some custom fastify errors to use

src/blob-store/index.js
achou11: ah I see that you added the type for the `clear()` method in this PR, so guessing the `any` was a placeholder?

achou11 (approved):
couple of non-blocking questions but overall looks good!

achou11:
also noticing that some of the files haven't been formatted using `npm run format`. we should probably set up a pre-commit step for this but in the meantime, mind running that on the relevant files?

tests/blob-server.js
gmaclennan: Just to point out that since this server is serving images, error messages do not make a lot of sense, because we are not processing the response in JS code (the response is handled on the platform specific image component). For now I think 404 is ok.

PR #247, opened 2023-09-05
update MapeoManager to return and handle project public IDs

### LINKED ISSUE #234: Use @mapeo/crypto to derive project id from project key
## Description

Follow-up to #171

~More of a chore task that applies throughout the codebase, but with the introduction of `projectKeyToPublicId` in `@mapeo/crypto`, we should update callsites that derive a project id from the project key to use it (currently most do `.toString('hex')` I think)~

To minimize exposure of potentially useful keys to the client, we want to update the project id that is returned from the MapeoManager + MapeoProject to be a hash of the project key (as opposed to just a different encoding).

## Tasks

- [x] Update `@mapeo/crypto` to 1.0.0-alpha.8 and use exported `projectKeyToPublicId` to create project public id
- [x] Update indexing of project records to save the public project id as a column in the client database

achou11:
Closes #234 

Notes:
- Updates `@mapeo/crypto` to 1.0.0-alpha.8
- Adds two opaque types: `ProjectId` and `ProjectPublicId`.
  - `ProjectId` refers to the "internal" project id, whereas `ProjectPublicId` is what's used for client consumption
- Adds a `projectPublicId` column to the `projectKeys` client db table
  - Tried a couple of things, but wasn't able to get Drizzle to understand that the `projectPublicId` column should store and return the opaque `ProjectPublicId` type. Pretty sure there's a way but my initial attempts failed me. For now, there are a couple of type casts in the manager code.

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -1,4 +1,5 @@
 import b4a from 'b4a'
+import { projectKeyToPublicId as keyToPublicId } from '@mapeo/crypto'
 
 /**
  * @param {String|Buffer} id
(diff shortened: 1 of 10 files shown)


src/utils.js
```diff
@@ -90,3 +91,33 @@ export function valueOf(doc) {
...
+    keyToPublicId(projectKey)
+  )
+}
+
+/**
+ * @param {import('./types.js').ProjectId} projectId
+ * @returns {Buffer} 24-byte nonce (same length as sodium.crypto_aead_xchacha20poly1305_ietf_NPUBBYTES)
+ */
+export function projectIdToNonce(projectId) {
+  return Buffer.from(projectId, 'hex').subarray(0, 24)
```
achou11, line 122: didn't use the sodium constant because it seems that `sodium-javascript` doesn't export it

gmaclennan (approved):
Looks good to me 👍

Issue #249, opened 2023-09-06
Initial deviceInfo implementation


gmaclennan:
## Description

This is a subtask of [Add device info records #182](https://github.com/digidem/mapeo-core-next/issues/182). 

As per https://github.com/digidem/mapeo-core-next/issues/240, we want a user to be able to set a device name at a device level (e.g. for all projects), but the data also needs to be stored in each individual project (so that other clients can read it).

We initially only want a device to only be able to write its own device name/info, and for that information to be shared across all projects. In the future we might want to allow administrators to name devices, and for a device to have different names on different projects.

## Tasks for initial implementation

- [x] Add a deviceInfo internal dataType
- [x] Add `client.setDeviceInfo()` and `client.getDeviceInfo()` methods
- [x] Create a client SQLite table for persisting local device info.

## Tasks for full implementation (tied to project.$member work)

- [ ] Implement `project.$member.update()` but only allow a device to update it's own deviceInfo (the role that is set by this method has different permissions, so needs to be evaluated separately)
- [ ] From MapeoManager, call `project.$member.update()` with the current device name when joining a project
- [ ] From the `client.setDeviceInfo()` method, call `project.$member.update()` for all current projects on the device

PR #250, opened 2023-09-06
feat: Add set & get deviceInfo & datatype

gmaclennan:
This fixes initial implementation #249. It adds an internal datatype for `deviceInfo` on the project instance, so that it can be used by the `$member` namespace for reading and writing member info. This part has no tests yet because it's best to test the `$member` methods once they are available.

This also adds `client.setDeviceInfo()` and `client.getDeviceInfo()` methods, along with a new table in the client db for storing this info.

I wasn't too sure about how to store this in SQLite. We will only ever have one deviceInfo record, so options are:

1. Store a single row with either:
    i. A column for each property
    ii A single JSON column with all info
2. Write a row for each property of deviceInfo, like how mbtiles metadata is stored.

I went with (2), but having done it I think (1) with a JSON column might be the better / simpler option.

### DIFF
--- a/drizzle/client/0000_charming_klaw.sql
+++ b/drizzle/client/0000_charming_klaw.sql
@@ -1,4 +1,4 @@
-CREATE TABLE `deviceInfo` (
+CREATE TABLE `localDeviceInfo` (
 	`deviceId` text NOT NULL,
 	`deviceInfo` text NOT NULL
 );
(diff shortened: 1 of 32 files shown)


achou11 (commented):
method implementations look good. can re-review once the sqlite changes are finalized

gmaclennan:
> methods implementations look good. can re-review once the sqlite changes are finalized

Ok changes made, it's ready to review again.

achou11 (approved):

Issue #124, opened 2023-07-17
Pre-emptively send all project haves via extension


gmaclennan:
We want to know what data (if any) needs to be synced before we start syncing, however in order to do that currently we would need to start replicating all project cores.

We don't want to add all project cores to the replication before the user has consented to sync, because this would allow the "other side" to start pulling data (because Hypercore currently does not expose a method to add a hypercore to a replication stream with "uploading" disabled).

This ticket is to add an extension to pre-emptively send the "haves" for all project cores via the project owner core (always the first core to connect and start replicating). This would allow us to give the user feedback about whether anything needs to be synced _before_ they decide to start syncing.

PR #254, opened 2023-09-06
feat: send core "haves" bitfield on first connect

gmaclennan:
This implements a solution to https://github.com/digidem/mapeo-core-next/issues/124 so that we can know how much data there is to sync before starting replicating the hypercores.

The code here implements a new custom hypercore extension to send core bitfields over the creator core replication stream, enabling the other peer to build up remote bitfields for each core for the connected peer. This can then be compared with a device's own bitfields to know what data needs to be sent.


gmaclennan:
This ended up being a bit more work than originally estimated. What this does:

1. The bitfields for all cores on a device are shared with a peer over the replication stream for the project creator core. The project creator core replication stream is always the first, and is connected before other cores are replicated to the stream.
2. The bitfields are written to `RemoteBitfield` instances, which is copied directly from the [Hypercore remoteBitfield](https://github.com/holepunchto/hypercore/blob/main/lib/remote-bitfield.js) (just added types).
3. These bitfields are stored by core discoveryId and by peer public key id.

The bitfield are compressed using [run-length-encoding](https://en.wikipedia.org/wiki/Run-length_encoding). The implementation is mainly copied from [`bitfield-rle`](https://github.com/mafintosh/bitfield-rle/blob/master/index.js), but modified to work with `Uint32Array` bitfields (which is what Hypercore uses). The code for encoding Uint32Arrays that works with different "endian-ness" was [copied from `compact-encoding`](https://github.com/holepunchto/compact-encoding/blob/5ee527f7d39f7153005b953c807ce3b530178610/index.js#L244).

I am leaving the use of these bitfields to a follow-up PR. We will use them for evaluating sync progress / the need for sync before actually replicating all cores between peers. We can either:

1. Copy these bitfields to [`peer.remoteBitfield`](https://github.com/holepunchto/hypercore/blob/4385cc779a014d5d24185a94980c1e21e1cd8e1c/lib/replicator.js#L308) when the peer connects, via [`peer.remoteBitfield.insert()`](https://github.com/holepunchto/hypercore/blob/4385cc779a014d5d24185a94980c1e21e1cd8e1c/lib/remote-bitfield.js#L249).
2. Replace `peer.remoteBitfield` directly with our "pre-updated" remote bitfield
3. Use both bitfields (our "pre-updated" bitfield and `peer.remoteBitfield`) when calculating progress

I think (1) is the preferable option, we just need to make sure it happens before the bitfield changes for the peer.

There is a lot in this PR. Some notes to help what to focus on:

1. `src/core-manager/remote-bitfield.js` is pretty much a direct copy of the [code in Hypercore](https://github.com/holepunchto/hypercore/blob/4385cc779a014d5d24185a94980c1e21e1cd8e1c/lib/remote-bitfield.js).
2. `src/core-manager/bitfield-rle.js` and `tests/bitfield-rle.js` are modified from [`bitfield-rle`](https://github.com/mafintosh/bitfield-rle/blob/master/index.js)

As part of this PR I also updated how extensions are handled, by handing off decoding/encoding to hypercore (rather than encoding/decoding in our own code). There is also a small fix to handling the existing extensions (for sharing core keys) that was uncovered as part of this work.

src/core-manager/compat.js
```diff
@@ -0,0 +1,17 @@
+import { createRequire } from 'module'
+
+const require = createRequire(import.meta.url)
+
+// Export the appropriate version of `quickbit-universal` as the plain import
+// may resolve to an older version in some environments
```
achou11, line 6: purely out of curiosity, what's this about?

achou11 (approved):
not much to comment on. excited about what this will enable!

src/core-manager/compat.js
gmaclennan: Don't actually know. This is copied from hypercore code

PR #261, opened 2023-09-12
WIP: add ipc wrappers

achou11:
Currently not working as I think the client implementation has some issues:

The test currently shows the following error after running this line in the test:

```ts
const project = await client.getProject(projectId)
```

```sh
node:internal/per_context/domexception:53
    ErrorCaptureStackTrace(this);
    ^
DOMException [DataCloneError]: function () { [native code] } could not be cloned.
    at new DOMException (node:internal/per_context/domexception:53:5)
    at SubChannel.postMessage (file:///Users/andrewchou/GitHub/digidem/mapeo-core-next/src/ipc-wrapper/sub-channel.js:38:23)
    at send (/Users/andrewchou/GitHub/digidem/mapeo-core-next/node_modules/rpc-reflector/client.js:78:15)
    at /Users/andrewchou/GitHub/digidem/mapeo-core-next/node_modules/rpc-reflector/client.js:262:11
    at new Promise (<anonymous>)
    at Object.apply (/Users/andrewchou/GitHub/digidem/mapeo-core-next/node_modules/rpc-reflector/client.js:260:31)
```


achou11:
@gmaclennan requesting review as in "I need help with figuring out the bugs" 😄

src/ipc-wrapper/server.js
```diff
@@ -0,0 +1,39 @@
+// @ts-check
+import { createServer } from 'rpc-reflector'
+import { SubChannel } from './sub-channel.js'
+
+/**
+ * @param {import('../mapeo-manager.js').MapeoManager} manager
+ * @param {import('rpc-reflector/server.js').MessagePortLike} messagePort
```
achou11, line 7: should this be accepting "plain" MessagePort instances too, similar to what rpc-reflector does? to account for usage in environments like the browser or electron, which adhere to what's described here: https://developer.mozilla.org/en-US/docs/Web/API/MessagePort

Node's worker_threads.MessagePort implementation does not match (interface + behavior), which is why things are slightly annoying.

improving interop would require some changes to event listeners similar to what's done here:

https://github.com/digidem/rpc-reflector/blob/3cbcc1d379b1bc7311cddded0d55fbc11319e3c4/client.js#L90-L92

also updating the SubChannel implementation

achou11:
running into what seems like race condition in the second test related to `client.getProject()`.  think there's still more work to be done in terms of properly closing rpc things, so maybe that would fix it

src/ipc-wrapper/server.js
```diff
@@ -0,0 +1,55 @@
...
+  const existingProjectServers = new Map()
+
+  const managerChannel = new SubChannel(messagePort, '@@manager')
+
+  const managerServer = createServer(manager, managerChannel)
+
+  messagePort.on('message', handleMessage)
+
+  return {
+    close() {
```
achou11, line 21: does this need to emit some "closing" event so that projects on the client can do additional handling when it comes to closing?

src/ipc-wrapper/server.js
```diff
@@ -0,0 +1,55 @@
...
+  async function handleMessage(payload) {
+    const id = payload?.id
+
+    if (typeof id !== 'string' || id === '@@manager') return
+
+    if (existingProjectServers.has(id)) return
+
+    const projectChannel = new SubChannel(messagePort, id)
+
+    const project = await manager.getProject(
```
gmaclennan, line 45: Here is where your race condition is (always look for `await` in your code flow to find them). At this stage a second message for this project could come during the `await`.

One way you can solve this: add queuing to the SubChannel implementation (I think Tomás has done this for the implementation in the mapeo mobile new example), and only start this channel after you have done `createServer()` below. You will also need something like `existingSubChannels` which is set before this await, and check for that at the start of this `handleMessage` (so you won't get here twice - the queued messages in SubChannel will handle that)

achou11: ahhh makes sense, thanks for explaining!

achou11: addressed via https://github.com/digidem/mapeo-core-next/pull/261/commits/9f6a7a5749f1b4c47ec4a1ffc44a18d29447bcc9

src/ipc-wrapper/sub-channel.js
```diff
@@ -0,0 +1,118 @@
...
+/**
+ * Node's built-in types for MessagePort are misleading so we opt for this limited type definition
+ * that fits our usage and works in both Node and browser contexts
+ * @typedef {EventTarget & { postMessage: (message: any) => void}} MessagePortLike
+ */
+
+/**
+ * @extends {TypedEmitter<Events>}
+ */
+export class SubChannel extends TypedEmitter {
```
achou11, line 21: wondering if i should avoid using TypedEmitter for this so that this can run in a browser environment as well (it's used for both server and client).

I havea local implementation that works using [EventTarget](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget) based on this helpful [post](https://dev.to/thepassle/events-are-the-shit-b3i). 

one of the main differences is the spec for the [`dispatchEvent()` method](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/dispatchEvent):

> The dispatchEvent() method of the [EventTarget](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget) sends an [Event](https://developer.mozilla.org/en-US/docs/Web/API/Event) to the object, (synchronously) invoking the affected event listeners in the appropriate order. The normal event processing rules (including the capturing and optional bubbling phase) also apply to events dispatched manually with dispatchEvent().

Most notably the part about _synchronously_ invoking listeners. Not sure if the Node implementation adheres to this or if it still behaves similar to an event emitter, but wondering if this specific detail could cause problems in our case or not.

achou11: i guess the simpler option would be using https://github.com/primus/eventemitter3, which is what rpc-reflector uses.

it's cool that the EventTarget approach works for our use case though!

achou11: addressed via https://github.com/digidem/mapeo-core-next/pull/261/commits/e067dc37cf03f4b76671324edab73052cf4f5499 (using eventemitter3)

src/ipc-wrapper/client.js
```diff
@@ -0,0 +1,72 @@
...
+        return createProjectClient
+      } else {
+        return Reflect.get(target, prop, receiver)
+      }
+
+      /**
+       * @param {import('../types.js').ProjectPublicId} projectPublicId
+       * @returns {Promise<import('rpc-reflector/client.js').ClientApi<import('../mapeo-project.js').MapeoProject>>}
+       */
+      function createProjectClient(projectPublicId) {
```
gmaclennan, line 42: Move this out of the scope of `get()`, into the `createMapeoClient()` scope.

test-e2e/ipc-wrapper.js
```diff
@@ -0,0 +1,94 @@
+import { test } from 'brittle'
```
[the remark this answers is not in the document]
gmaclennan: Add test for `client.getProject('non-existent-project-id')`

gmaclennan (changes requested):
Great work on this, this is tricky code and you've solved it really well. I realised reading through this that we need to handle the case of a project not being found, or an error when doing `manager.getProject()`. I have given a suggestion about how to do this. We don't need a whole `SubChannel` for this, we just need to pass a message somehow to the client.

src/ipc-wrapper/client.js
```diff
@@ -0,0 +1,72 @@
...
+        const existingClient = existingProjectClients.get(projectPublicId)
+
+        if (existingClient) return Promise.resolve(existingClient.instance)
+
```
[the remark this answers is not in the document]
gmaclennan: Actually recommend using [`p-timeout`](https://github.com/sindresorhus/p-timeout) here in case something happens and we never get a message back.

src/ipc-wrapper/client.js
```diff
@@ -0,0 +1,72 @@
...
+      channel.close()
+    },
+  })
+
+  const managerChannel = new SubChannel(messagePort, MANAGER_CHANNEL_ID)
+
+  /** @type {import('rpc-reflector').ClientApi<import('../mapeo-manager.js').MapeoManager>} */
+  const managerClient = createClient(managerChannel)
+
+  managerChannel.start()
```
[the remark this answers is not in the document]
achou11: addressed via https://github.com/digidem/mapeo-core-next/pull/261/commits/71a8398c522e298001b3de239ed2d81e92ebf8f1

src/ipc-wrapper/client.js
achou11: addressed via https://github.com/digidem/mapeo-core-next/pull/261/commits/fd58298a81bb2dae703bcd11ec76213b41d19863

src/ipc-wrapper/client.js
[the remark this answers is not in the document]
achou11: addressed via https://github.com/digidem/mapeo-core-next/pull/261/commits/b7e98da58adab73cf23704734ea2cd7fa8891243

not entirely sure if it was just me, but this turned out to be much more complicated than I thought. ended up adding a back-and-forth initiated by the client on a new subchannel in order for the server to actually handle the getProject as intended. think it leads to a clearer understanding of what the server is doing though, which is helpful

src/ipc-wrapper/sub-channel.js
```diff
@@ -0,0 +1,115 @@
...
+
+      switch (this.#state) {
+        case 'idle': {
+          this.#queued.push(value)
+          break
+        }
+        case 'active': {
+          const { id, message } = value
+
+          if (this.#id !== id) return
```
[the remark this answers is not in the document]
achou11: addressed via https://github.com/digidem/mapeo-core-next/pull/261/commits/612f06869c18049e1e34916cf9cf7e307e02549c

test-e2e/ipc-wrapper.js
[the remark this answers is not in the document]
achou11: addressed via https://github.com/digidem/mapeo-core-next/pull/261/commits/b7e98da58adab73cf23704734ea2cd7fa8891243 and https://github.com/digidem/mapeo-core-next/pull/261/commits/687c2bcb2097d28f0b51501ad31554d0502e8655

src/ipc-wrapper/server.js
```diff
@@ -0,0 +1,78 @@
...
+ * @param {import('./sub-channel.js').MessagePortLike} messagePort
+ */
+export function createMapeoServer(manager, messagePort) {
+  /** @type {Map<string, { close: () => void, channel: SubChannel }>} */
+  const existingProjectServers = new Map()
+
+  const mapeoIpcChannel = new SubChannel(messagePort, MAPEO_IPC_ID)
+  mapeoIpcChannel.start()
+
+  mapeoIpcChannel.on('message', handleGetProject)
```
achou11, line 16: i changed the implementation such that it's only doing work when the client specifically requests it i.e. we call `client.getProject()`, which sends a subchannel message from the client that is received here. 

before, we were listening on the whole messagePort and then filtering for relevant messages. this approach handles "indirect" requests of a project client i.e. a client calls a method on an active project client, but I'm not sure that's a situation we actually run into (since we always keep a map of existing project clients and don't do any automatic cleanup).

wondering if the edge case here is still relevant or if this new approach suffices

src/ipc-wrapper/sub-channel.js
```diff
@@ -0,0 +1,117 @@
+// @ts-check
+import { EventEmitter } from 'eventemitter3'
+import { extractMessageEventData } from './utils.js'
+
+// Ideally unique ID used for identifying "global" Mapeo IPC messages
+export const MAPEO_IPC_ID = '@@mapeo'
```
achou11, line 6: open to name changes

src/ipc-wrapper/client.js
```diff
@@ -0,0 +1,113 @@
...
+   * @param {import('../types.js').ProjectPublicId} projectPublicId
+   * @returns {Promise<import('rpc-reflector/client.js').ClientApi<import('../mapeo-project.js').MapeoProject>>}
+   */
+  async function createProjectClient(projectPublicId) {
+    const existingClient = existingProjectClients.get(projectPublicId)
+
+    if (existingClient) return existingClient.instance
+
+    // Wait for the server to confirm that it could retrieve the desired project
+    await pTimeout(
```
gmaclennan, line 55: Unfortunately another race condition here. I believe this test will fail:

```js
const projectId = await client.createProject()
const [projectInstance1, projectInstance2] = await Promise.all([
  client.getProject(projectId),
  client.getProject(projectId)
])
t.is(projectInstance1, projectInstance2)
```

One way to solve this it with a deferred promise (I use `p-defer`). 

```js
  if (existingClient) return existingClient.deferredInstance
  const deferred = pDefer()
  existingProjectClients.set(projectPublicId, deferred.promise)

  //...

  deferred.resolve(projectClient)
```

Will also need to reject the deferred promise in the catch, check with this test:

```js
const results = await Promise.allSettled([
  client.getProject('aaaa'),
  client.getProject('aaaa')
])
t.alike(results.map(({ status }) => status), ['rejected', 'rejected'])
```

gmaclennan (changes requested):
Good progress on this, still some changes needed, see comments in code.

src/ipc-wrapper/server.js
```diff
@@ -0,0 +1,113 @@
...
+    const projectChannel = new SubChannel(messagePort, id)
+    existingProjectChannels.set(id, projectChannel)
+
+    let project
+    try {
+      project = await manager.getProject(
+        /** @type {import('../types.js').ProjectPublicId} */ (id)
+      )
+    } catch (err) {
+      // TODO: how to respond to client so that method errors?
```
achou11, line 76: for posterity, idea from Gregor is to hook into RPC internals and call `postMessage()` with the encoded RPC message that represents an error that's sent due to method erroring. can be done as follow-up

achou11 (commented):
open to name changes in regarding the introduced `MapeoRpc...` variables. Just named them that way since it aligned with the `MAPEO_RPC_ID` constant that was introduced.

src/ipc-wrapper/server.js
achou11: reverted 👍

test-e2e/ipc-wrapper.js
```diff
@@ -0,0 +1,170 @@
...
+
+  t.alike(
+    results.map(({ status }) => status),
+    ['rejected', 'rejected']
+  )
+
+  return cleanup()
+})
+
+test('Concurrent calls that succeed', async (t) => {
```
achou11, line 92: what's a better name for this test?

gmaclennan (approved):
Excellent work on this @achou11 - this is a very tricky piece of code and I think it's pretty solid right now. There is a small suggestion to commit before merging, and two issues to create to deal with later.

src/ipc-wrapper/sub-channel.js
```diff
@@ -0,0 +1,117 @@
...
+  get id() {
+    return this.#id
+  }
+
+  /**
+   * Send messages with the subchannel's ID
+   * @param {any} message
+   */
+  postMessage(message) {
+    this.#messagePort.postMessage({ id: this.#id, message })
```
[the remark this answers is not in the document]
achou11: Created issue here: https://github.com/digidem/mapeo-core-next/issues/277

src/ipc-wrapper/client.js
```diff
@@ -0,0 +1,96 @@
...
+    projectClientPromises.set(projectPublicId, deferred.promise)
+
+    try {
+      await mapeoRpcClient.assertProjectExists(projectPublicId)
+    } catch (err) {
+      deferred.reject(err)
+      throw err
+    }
+
+    const projectChannel = new SubChannel(messagePort, projectPublicId)
```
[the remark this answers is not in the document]
achou11: created issue here: https://github.com/digidem/mapeo-core-next/issues/276

PR #271, opened 2023-09-15
fix: change version IDs to use the discovery keys

### LINKED ISSUE #265: Change version IDs to use the discovery keys instead of core keys
## Description

Core keys should be considered "sensitive" data. They are the authentication mechanism to replicate a hypercore. We have additional layers of security with the project key, encryption of hypercores, and the capabilities check before any cores are replicated, but it seems like we should still treat hypercore keys carefully.

Currently we use the hypercore key as part of the version ID. The risk is that we export data including the version ID without realizing its sensitive nature.

## Tasks

- [x] https://github.com/digidem/mapeo-schema/issues/141
- [ ] Lookup previous versions by discoveryKey in DataType
- [ ] Use the discovery key when decoding in the Indexer

gmaclennan:
Fixes #265

Depends on https://github.com/digidem/mapeo-schema/pull/145 and
updating mapeo-schema

Contains fixes for @mapeo/schema@3.0.0-next.9
(renamed project -> projectSettings)

### DIFF
--- a/drizzle/client/0000_tidy_micromacro.sql
+++ b/drizzle/client/0000_tidy_micromacro.sql
@@ -3,10 +3,6 @@ CREATE TABLE `localDeviceInfo` (
 	`deviceInfo` text NOT NULL
 );
 --> statement-breakpoint
-CREATE TABLE `project_backlink` (
-	`versionId` text PRIMARY KEY NOT NULL
-);
---> statement-breakpoint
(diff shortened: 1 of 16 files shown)


achou11:
@gmaclennan is it okay to close this in favor of #290 ?

gmaclennan:
Close in favour of #290

PR #274, opened 2023-09-19
feat: addCreatedBy

### LINKED ISSUE #153: Add `createdAt` and `createdBy`
## Problem

Clients need to know which device a record was created by, because for some devices (currently mobile, could be based on role later) can only edit records that they created.

Clients need to know when a record was created in order to sort by created date.

`createdAt` and `createdBy` should not be editable.

## Possible soultions

1. _Store `createdAt` and `createdBy` in every record_

    This is the simplest to implement. Every time a document is updated then these two fields are duplicated to the new version. However there is nothing to stop a "rogue" device changing these fields. However this might be ok, if we are assuming that trusted users are only interacting with the database through the Mapeo Core API, which can enforce that these never change.

2. _Derive `createdAt` and `createBy` from "root" record_

    The indexer could track back to the "root" record (e.g. the version with `links: []` - the record from the `create()` action). `createdBy` can be derived from the hypercore key where this record is written. `createdAt` just needs to be trusted. This means a "rogue" client cannot change these fields by updating a document, but it adds more complexity to the indexer. If the "root" record is unavailable (e.g. if the original has not synced for whatever reason) then these fields would be unknown / undefined.

3. _Encode `createdAt` and `createdBy` to the `docId`_

    Reading about [snowflake IDs](https://en.m.wikipedia.org/wiki/Snowflake_ID) made me think of this. This has the advantage of (2) without the cost of indexing complexity and the problem of missing "root" records. (we still essentially implicitly trust that a `docId` is not changed by an update - but that would effectively create a new document anyway). We currently use 256-bit random numbers for our `docId`, which is overkill really. 72-bits of entropy with 1 million document IDs would have a [1 in 9.4 billion chance of collision](https://hash-web-e45df.web.app/en/), which seems like plenty - I think worth adding an extra byte over 64-bit, which would have a 1 in 37 million chance of collision with 1 million IDs.

    Some ways we might implement our IDs:

    - Encode a hash of the creator device public key as the device ID, so that the ID can only be used to find the creator device ID if they know the public key. Could hash with the random part of the ID, but would be expensive to derive the device ID for every record (because when reading would need to calculate a hash for each record). Device public keys are 256-bits, and the hash would be the same. Suggest just using the first 64-bits of the hashed device ID.
    - Encode 72-bits of random data
    - Encode a 48-bit timestamp as milliseconds since UNIX epoc (gives us until 10,889 AD)
    - Encode the random 72-bit part first (e.g. docID to be `${random72bits}${deviceID64bits}${timestamp48bits}` so that when a user shares a [doc ID externally by only sharing the first few bytes](https://github.com/digidem/mapeo-core/issues/104) then we can lookup the ID quickly from SQLite index tables via searching for IDs that start with this value.
    - We don't need IDs to be sortable by creation date - can decode that to a separate column when indexing.

I'm leaning towards option (3), but option (1) might be ok, but we can't change it in the future, so maybe worth encoding this info to docIds anyway?

## Related

- https://github.com/digidem/mapeo-schema/issues/83
- https://github.com/digidem/mapeo-sqlite-indexer/issues/7
- https://github.com/digidem/mapeo-core-next/issues/136

tomasciccola:
This should close #153 

Additionally, this PR takes care of refactoring `project` -> `projectSettings` after `@mapeo/schema` update


achou11:
>  On the create fn in src/datatype/index.js:119 how should I create the createdBy field?

@gmaclennan may have a proper answer for this but my guess is...

the dataStore instance will handle that internally in the `write` call so you'd pass a document without that field in the datatype create, but then in the dataStore write method, you update this line:

https://github.com/digidem/mapeo-core-next/blob/ca3b6c73d5bbd6650124d27f14410b1a188f54ef/src/datastore/index.js#L136

In terms of actually deriving the discovery id, guessing it'd be something like this but not entirely sure:

```ts
const discoveryId = crypto
  .discoveryKey(this.#writerCore.key)
  .toString('hex')
  
const block = encode({
  ...doc,
  createdBy: discoveryId,
})
```

tomasciccola:
> ```ts
> const discoveryId = crypto
>   .discoveryKey(this.#writerCore.key)
>   .toString('hex')
> ```

is `crypto` from `hypercore-crypto` ?

achou11:
> is crypto from hypercore-crypto ?

yes, should've updated that name 😅

tomasciccola:
Checks are failing on the `build:types` stage. Most type errors are related to `project`. Are we letting it fail? other individual stages seems to fail too.

tomasciccola:
Oh, right. We changed `project` to `projectSettings`. I'll probably take time to change that in a separate PR

tomasciccola:
It seems that we're getting a runtime error in the `data-type.js` tests complaining that the column `createdBy` doesn't exist. I tried to follow the trail but I haven't found something that rings a bell. 
I don't know if there's some build step related to the drizzle ORM that I should take into account

tomasciccola:
I ran the npm scripts to regenerate the project and client tables and now it seems to be working.

gmaclennan (approved):
Looks good, good work.

gmaclennan (dismissed):
Sorry just realized there is no test for this. Can you add a test for createdBy before merging.

tomasciccola:
> Sorry just realized there is no test for this. Can you add a test for createdBy before merging.

I added checkf of `createdBy` to the e2e test of mapeo-manager, and CRUD which meant also modifying some of its source code. Does it make sense what I did?

gmaclennan:
> I added checkf of `createdBy` to the e2e test of mapeo-manager, and CRUD which meant also modifying some of its source code. Does it make sense what I did?

We should have a test that checks that createdBy matches the discovery key of the writer core of the device that created it, and that createdBy stays the same even when a different device edits. (in the current test for it not changing, the edit is by the same device/instance). This can be a unit test if needed, because I don't think we expose the writer core key, although we could maybe expose it by a private symbol which is what we have done for some e2e tests.

tests/data-type.js
```diff
@@ -49,3 +55,125 @@ test('private createWithDocId() method', async (t) => {
...
+  /** @type {import('@mapeo/schema').ObservationValue} */
+  const newObsFixture = {
+    schemaName: 'observation',
+    refs: [{ id: randomBytes(32).toString('hex') }],
+    tags: {},
+    attachments: [],
+    metadata: {},
+  }
+  dt2.update(obs.versionId, newObsFixture)
+
```
tomasciccola, line 132: I'm missing here a way so that the peer 1 gets the updated observation from peer 2 - which I _think_ its necessary for this test? - 
I'm probably missing something @achou11 ?

achou11: not sure if this test is needed. the primary things that should be tested are:

1. A created document should have a `createdBy` field that matches its creator's writer core's discovery key id. 

2.  A created document that is replicated to another core...

    1. holds the same `createdBy` as when it was created.
    2. holds the same `createdBy` when making an update to that document.
    
I think the first two tests you added are addressing these cases, respectively (`test validity of `createdBy` field` and `test validity of `createdBy` field from another peer`).

I think the case you're describing would be more necessary if we had to work with an `updatedBy` field.

achou11: in short, I think you can include the "update" test in the `test validity of createdBy field from another peer` test, where the following occurs in that test:

1. Create doc from peer 1. Check `createdBy`
2. Replicate doc from peer 1 to peer 2.
3. Get doc in peer 2 and check `createdBy`
4. Update doc in peer 2 and checkc `createdBy`

tomasciccola: Ok, in the last commit I added that missing check on the second test (that `createdBy` is doesn't change when updating the document). 
I think that this is maybe ready to merge?

src/mapeo-manager.js
```diff
@@ -246,7 +246,7 @@ export class MapeoManager {
   }
 
   /**
-   * @returns {Promise<Array<Pick<ProjectValue, 'name'> & { projectId: ProjectPublicId, createdAt?: string, updatedAt?: string }>>}
+   * @returns {Promise<Array<Pick<ProjectValue, 'name'> & { projectId: ProjectPublicId, createdAt?: string, updatedAt?: string, createdBy?: string }>>}
```
achou11, line 249: not sure if `listProjects()` needs to include `createdBy`. how useful is it for a frontend client to have the info?

achou11 (approved):
Overall looks good! The one thing I would consider before merging is whether the update to `MapeoManager.listProjects()` is necessary.

src/mapeo-manager.js
achou11: Think for now I would revert this (and the corresponding changes to the tests) and it's it turns out to be useful, do as a follow-up PR

tomasciccola: reverted on last commit, and now merging

Issue #278, opened 2023-09-21
Publish IPC wrapper as separate module


achou11:
## Description

To make integration of Mapeo Core easier apps, we should move the IPC wrapper implementation from https://github.com/digidem/mapeo-core-next/tree/aae39e8815b8fe813d8667a87be7a78cf27a1800/src/ipc-wrapper into a separate published npm module.

## Tasks

- [X] Create separate module https://www.npmjs.com/package/@mapeo/ipc
- [ ] Remove implementation from core (see #306)

PR #280, opened 2023-09-25
chore: remove usage of opaque ids types related to project ids

achou11:
Towards #278

I had introduced opaque types for project id and project public id, but made the mistake of using those for the public-facing API, which makes their usage more tedious and doesn't bring any major benefit for now. 

Opaque types are useful if usage is kept internally, but maybe that's something to revisit later on.

Doing this towards publishing the IPC wrapper as a separate module, which works with APIs that used to rely on these opaque types.


gmaclennan (approved):
I'm still on the fence about opaque types - I can definitely see the appeal. But happy to remove these for now. Part of the problem is the syntax when using JSDoc.

achou11:
yeah hoping there's some sort of write-up about using them and the best way to utilize them for libraries.

PR #284, opened 2023-09-26
feat: update main exports

achou11:
Towards #278 

Waiting for #283 to be merged since it also includes some relevant changes

Summary of changes (so far):

- [X] adds `MapeoManager` to exports
- [X] removes `MapeoProjects` from exports

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,1 +1,2 @@
 export { MapeoProject } from './mapeo-project.js'
+export { MapeoManager } from './mapeo-manager.js'
--- a/src/index.js
+++ b/src/index.js
@@ -1,2 +1,1 @@
-export { MapeoProject } from './mapeo-project.js'
 export { MapeoManager } from './mapeo-manager.js'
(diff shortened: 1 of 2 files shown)


achou11:
@gmaclennan anything else that you think should be updated in this PR?

gmaclennan:
> @gmaclennan anything else that you think should be updated in this PR?

I don't think so. I already did all the work for publishing typescript types, and have done a test publish as a module. It is currently published as `@mapeo/core@9.0.0-alpha.0`, and only the `alpha` npm dist tag points to it - the `latest` dist tag points to v8.6.2 (we should keep it like that for now).  To install do `npm i @mapeo/core@alpha` or use the specific version `npm i @mapeo/core@9.0.0-alpha.0`

gmaclennan:
Maybe remove the MapeoProject export now? Not sure it is needed?

achou11:
> Maybe remove the MapeoProject export now? Not sure it is needed?

Feel like it'd be useful to keep if we're talking about general usage of this library, but if we're not concerned about that right now then that's okay too

gmaclennan (approved):

PR #286, opened 2023-09-27
feat: expose capabilities in MapeoProject and its member api

### LINKED ISSUE #239: Expose "ownCapabilities"
## Description

We need a way to get a device's own capabilities. This will be used to show the correct UX depending on what a device's capabilities are (e.g. if a device is able to assign roles, then it is able to invite other devices).

Not sure what the best API for this is? `project.getOwnCapabilities()`?

achou11:
Closes #239 

Exposes capabilities in the following ways:

- adds a `$getOwnCapabilities()` method to `MapeoProject`
- adds a `capabilities` field to the return value of `project.$member.getById()` and `project.$member.getMany()`

TODO:

- [ ] Feel like some of the tests belong in e2e but not sure if that's blocked on the invites stuff
  - the e2e capabilities tests should ideally be able to use `project.$member` to retrieve the capabilities. currently relies on https://github.com/digidem/mapeo-core-next/issues/289 and updating the `MemberApi.invite()` implementation
  - the e2e manager invite tests should check that the inviter can retrieve the member's device info using the member api after the invite is accepted
- [X] still wondering what the best way to get your own capabilities for a project is. basically, should there be a `project.$getOwnCapabilities()` method?
  - [X] Update tests to account for this method

### DIFF
--- a/src/mapeo-manager.js
+++ b/src/mapeo-manager.js
@@ -16,6 +16,7 @@ import {
 import { ProjectKeys } from './generated/keys.js'
 import {
   deNullify,
+  getDeviceId,
   projectIdToNonce,
   projectKeyToId,
   projectKeyToPublicId,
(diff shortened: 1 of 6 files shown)


gmaclennan (approved):
Looks good, I think we can sort out the e2e tests once we have some of the other things in place.

Reviewing this I realized that we need to tweak a couple of other things, which I have opened issues for:

- #299 
- #298

Issue #287, opened 2023-09-28
Add new `CoreReplicationState` class


gmaclennan:
## Description

Reviewing how we handle replication state, I think it needs some restructuring to get the information we need for the UX designs.

1. We need to know the replication state _before_ we actually have the core, based on the shared "haves"
2. Information we need to download is relatively easy, but uploads are slightly more confusing - if we are connected to 10 peers and they each want the same block from us, do we show that as 10x the data to send as if we were sending to 1 peer? Because of the p2p sync, it should not take nearly 10x as long. But how to estimate?
3. It will likely be better if the "raw" state we keep is bitfields, and then we can run a function to calculate state based on those bitfields. This can be optimized for speed and tested, and tweaked to provide the right data
4. We need to support "wants" being bitfields, rather than just "everything you have", so that we can show upload progress for sparse blob sync. We need to be able to set these "want" bitfields dynamically, when we get the information

Plan

- Create CoreReplicationState with discoveryId
- `attachCore` method - can only be called once, must match discoveryId
- `addHave(peerId, bitfield)` - adds a remoteBitfield instance for the given peer
- When core is attached, when peers connect, can update this bitfield per peer
- "wants" by peer, by default no bitfield - want everything
- `addWant(peerId, { start, length })` - create want bitfield if it doesn't exist, add wants to it.
- `state` event is bitfields:
```ts
type Bitfield<T extends 'remote' | 'local'> = T extends 'remote' ? RemoteBitfield : LocalBitfield
type PeerReplicationState<T extends 'remote' | 'local'> = {
  have: Bitfield<T>,
  want?: Bitfield<T>,
}
type CoreReplicationState = {
  length: number
  localState: PeerReplicationState<'local'>
  peerStates: Record<string, PeerReplicationState<'remove'>>
}
``` 

Then a function that takes `CoreReplicationState` and returns a usable state with numbers that indicate how much there is to do.

```ts
type CoreReplicationState = {
  length: number // length of core
  have: number // how many blocks we have
  want: number // how many blocks we want that at least one peer has
  wanted: number // how many blocks are wanted by at least one peer (only count each block once, e.g. one block wanted by ten peers is `1`) 
}
```

Note: `missing = length - have - want`

## Tasks

- [ ]

PR #288, opened 2023-09-28
feat: CoreReplicationState

gmaclennan:
Add a new CoreReplicationState implementation, see
https://github.com/digidem/mapeo-core-next/issues/287

### DIFF
--- a/src/core-manager/core-replication-state.js
+++ b/src/core-manager/core-replication-state.js
@@ -5,16 +5,34 @@ import RemoteBitfield from './remote-bitfield.js'
 /**
  * @typedef {RemoteBitfield} Bitfield
  */
-
 /**
  * @typedef {string} PeerId
  */
-
 /**
- * @typedef {Object} CoreState
+ * @typedef {Object} InternalState
  * @property {number | undefined} length Core length, e.g. how many blocks in the core (including blocks that are not downloaded)
  * @property {PeerState} localState
  * @property {Map<PeerId, PeerState>} remoteStates
+ */
+/**
+ * @typedef {object} PeerSimpleState
+ * @property {number} have blocks the peer has locally
+ * @property {number} want blocks the peer wants, and at least one peer has
+ * @property {number} wanted blocks the peer has that at least one peer wants
+ * @property {number} missing blocks the peer wants but no peer has
+ */
+/**
+ * @typedef {PeerSimpleState & { connected: boolean }} RemotePeerSimpleState
+ */
+/**
+ * @typedef {object} DerivedState
+ * @property {number} coreLength known (sparse) length of the core
+ * @property {PeerSimpleState} localState local state
+ * @property {Record<PeerId, RemotePeerSimpleState>} remoteStates map of state of all known peers
+ */
+/**
+ * @typedef {object} CoreReplicationEvents
+ * @property {() => void} update
  */
 
 class PeerState {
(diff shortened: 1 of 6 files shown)


gmaclennan:
@achou11 I still need to write tests, but otherwise this is ready for review. Hopefully this is clearer than the [current implementation](https://github.com/digidem/mapeo-core-next/blob/1e0df1c2380f9231a6f8a6fc9bf74204322ccbb8/src/sync/replication-state.js#L108). It considers state in relation to all connected peers, not a separate state with each peer as done previously. I think for the UI, the main two numbers are `localState.want` and `localState.wanted` - when they are both `0` then there is nothing to sync. When we add UI for peers, we will use the `want` and `wanted` for each peer. I don't think we have a use yet have a use yet for `have` and `missing`, but including them since we already do the calculations for them. The `deriveState` could probably be written more efficiently, using bitwise operations maybe, or using the bitfield indexes to `findFirst(value)` to avoid iterating through every block in the core, but I think it's ok to start with something inefficient but readable, and optimize in the future if we identify this as a bottleneck. I've designed this as "pull based" - deriving the state is expensive (because of iterating through the bitfields) so instead it emits an `update` event (passing no arguments) when we know state has changed, and the consumer can then call `getState()`. I plan to write another class that combines state for all cores (in a separate PR), that will listen for `update` events on each core, and throttle its events, and each time this combined state class emits state, it will call `getState()` on any core replication states that it knows have been updated.

gmaclennan:
I added some benchmarks and made some improvements to the core that reduced the benchmark time from 4,150ms to 2,180ms. It's still not super fast. I can think of two ways to improve things:

1. Iterate through each bitfield in 32-bit chunks, and use JS bitwise operations to calculate haves and wants
2. Use the bitfield indexing such as `findFirst` to avoid reading every bit in a bitfield

I think this is fast enough for the MVP, and the options above might take a while, so we can attempt later.

gmaclennan:
> I added some benchmarks and made some improvements to the core that reduced the benchmark time from 4,150ms to 2,180ms. It's still not super fast. I can think of two ways to improve things:
> 
> 1. Iterate through each bitfield in 32-bit chunks, and use JS bitwise operations to calculate haves and wants
> 2. Use the bitfield indexing such as `findFirst` to avoid reading every bit in a bitfield
> 
> I think this is fast enough for the MVP, and the options above might take a while, so we can attempt later.

I ended up adding bitwise operations (after adding tests, since the code is harder to reason about), which increases performance by about 16x. The benchmarks now run in 94ms - I increased the iterations to 10,000 to get better accuracy.

src/core-manager/core-replication-state.js
```diff
@@ -0,0 +1,395 @@
...
+    // @ts-ignore - we know discoveryKey exists here
+    const discoveryId = keyToId(core.discoveryKey)
+    if (discoveryId !== this.#discoveryId) {
+      throw new Error('discoveryId does not match')
+    }
+    if (!this.#core) return
+
+    this.#core = core
+    // @ts-ignore - internal property
+    this.#localState.setHavesBitfield(core?.core?.bitfield)
```
achou11, line 112: Small suggestion since I've been bit by this before, but mind changing this placement of the ignore on the intended call?

```suggestion
    this.#localState.setHavesBitfield(
      // @ts-ignore - internal property
      core?.core?.bitfield
    )
```

src/core-manager/core-replication-state.js
```diff
@@ -0,0 +1,395 @@
...
+  n = (n & 0x33333333) + ((n >> 2) & 0x33333333)
+  return (((n + (n >> 4)) & 0xf0f0f0f) * 0x1010101) >> 24
+}
+
+/**
+ *
+ * @param {RemoteBitfield} bitfield
+ * @param {number} index
+ */
+function getBitfieldWord(bitfield, index) {
```
achou11, line 387: mind briefly elaborating on what "word" means in this context?

src/core-manager/core-replication-state.js
```diff
@@ -0,0 +1,395 @@
...
+  }
+  return derivedState
+}
+
+/** @param {number} n */
+function bitCount(n) {
+  n = n - ((n >> 1) & 0x55555555)
+  n = (n & 0x33333333) + ((n >> 2) & 0x33333333)
+  return (((n + (n >> 4)) & 0xf0f0f0f) * 0x1010101) >> 24
+}
```
achou11, line 380: wish i could even pretend to know what this is doing 😄

achou11 (commented):
overall seems promising! the explanation makes sense to me, although parts of the implementation are a little over my head. very nice work!

src/core-manager/core-replication-state.js
gmaclennan: very good idea - I've been bitten by this before too

src/core-manager/core-replication-state.js
gmaclennan: It means "32-bit integer", since the bitfield is a `UInt32Array` - this reads a single entry from that array. I'll add a code comment.

achou11 (approved):
can't really comment much on the deeply technical parts of the implementation, but the abstractions, methodology, and tests seem good to me.

hoping you're prepared to revisit this code down the line 😄

PR #290, opened 2023-09-28
chore: update @mapeo/schema to 3.0.0-next.10

### LINKED ISSUE #265: Change version IDs to use the discovery keys instead of core keys

### LINKED ISSUE #272: Change driveIds to use discovery keys instead of core keys
## Description

Same as #265 but for driveIds referenced in observation attachments blob IDs.

## Tasks

- [ ]

achou11:
Closes #265 
Closes #272

Extracts the relevant changes from #271 

This version of `@mapeo/schema` introduced the following:

- [X] adding `deleted` field to common schema
- [X] renaming coreKey to coreDiscoveryKey
- [X] renaming driveId to driveDiscoveryId (see #294)
  - this one doesn't have many direct changes from schema, but requires us to derive the drive ids differently

### DIFF
--- a/drizzle/client/0000_foamy_madame_hydra.sql
+++ b/drizzle/client/0000_foamy_madame_hydra.sql
@@ -22,6 +22,7 @@ CREATE TABLE `projectSettings` (
 	`createdBy` text NOT NULL,
 	`updatedAt` text NOT NULL,
 	`links` text NOT NULL,
+	`deleted` integer NOT NULL,
 	`name` text,
 	`defaultPresets` text,
 	`forks` text NOT NULL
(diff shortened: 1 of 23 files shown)


tomasciccola (approved):
LGTM!

tomasciccola (approved):

PR #294, opened 2023-10-02
use drive discovery key for drive ids

### LINKED ISSUE #272: Change driveIds to use discovery keys instead of core keys

achou11:
Closes #272 

Stacked on #290 

This PR:

- updates BlobStore to use discovery keys instead of public keys for drive ids
- renames the `driveId` field of the `BlobId` type to `driveDiscoveryId`, to align with the issue's desire to update the observation attachment reference.

I initially thought we may want to rename `driveId` used in the blobs-related code to `driveDiscoveryId` but then realized that we may not want that. If we want to revert the naming, just a matter of reverting the last commit on this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,44 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+## [9.0.0-alpha.1](https://github.com/digidem/mapeo-core/compare/v9.0.0-alpha.0...v9.0.0-alpha.1) (2023-10-03)
+
+
+### Features
+
+* $blobs.getUrl and $blobs.create methods ([#184](https://github.com/digidem/mapeo-core/issues/184)) ([fb33178](https://github.com/digidem/mapeo-core/commit/fb33178e61a0b96add31809a227e59f954ba3d1d))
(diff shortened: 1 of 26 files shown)


gmaclennan (approved):
Looks good to me. Not sure about the naming question. Internally, it makes sense that it is called `driveDiscoveryId`, since that describes what it is, but for the API consumer, I think `driveId` makes sense, because for the API consumer it is just "ID that identities the drive the blob is stored on", and we don't expose any other ID to the API consumer.

In summary, I think maybe call it `driveId` where the API consumer sees it, even though that makes it a little confusing when reading the internals - just make sure to add comments.

achou11:
yeah all of that came to mind as I was implementing.

One question I have then is should `BlobApi.create()` contain a field called `driveId` or `driveDiscoveryId`? Your suggestion seems to suggest the former, but my understanding is that the latter would be more convenient because you could directly use the return value as the value for an observation attachment (which uses `driveDiscoveryId` based on the schema change)

achou11:
Decided to basically revert the renaming entirely, without attempting to strategically update naming for internals. Just became confusing and I think for now the comments are helpful enough. Can revisit if it really does become super confusing

PR #297, opened 2023-10-02
fix: write own device info when creating and adding projects

### LINKED ISSUE #289: Save relevant device info when initially creating/adding a project
## Description

When creating a project instance via `MapeoManager` using `createProject()` or `addProject()`, the project instance should attempt to write to its deviceInfo store with information from `MapeoManager.getDeviceInfo()`. A couple of ideas for approaching:

1. Pass the device info as a constructor param for `MapeoProject`, which is then used when instantiating

2. Expose a method on `MapeoProject` to add that info and then call it from the consumer after the instance is created. Maybe in the `$member` namespace i.e. `project.$member.add(deviceId, info)`. e.g.

```ts
// From MapeoManager...
// `addProject` would be similar

async createProject() {
  // ... 

  const project = new MapeoProject({...})
  
  const selfDeviceInfo = await this.#getDeviceInfo()
  
  project.$member.add(this.#deviceId, selfDeviceInfo)
  
  // ...
}
```

One potential advantage for 2 is that it could also be used by `MemberApi.invite()`, which lives in the `MapeoProject`, We currently don't save the device info if the invite is accepted (currently only update the capabilities). Believe we still need #245 but once that's in place, we can update the implementation like so:

```diff
if (response === InviteResponse_Decision.ACCEPT) {
+ await this.#add(deviceId, deviceInfo) // device info would come from rpc invite response
  await this.#capabilities.assignRole(deviceId, roleId)
}
```

## Tasks

- [ ]

achou11:
Closes #289 

exposes a `$setOwnDeviceInfo()` method on `MapeoProject`, which is used by the `MapeoManager` when calling `createProject()`, `acceptProject()`, and `setDeviceInfo()`


src/mapeo-project.js
```diff
@@ -61,6 +63,7 @@ export class MapeoProject {
    * @param {IndexWriter} opts.sharedIndexWriter
    * @param {import('./types.js').CoreStorage} opts.coreStorage Folder to store all hypercore data
    * @param {import('./rpc/index.js').MapeoRPC} opts.rpc
+   * @param {() => Promise<Partial<Pick<import('@mapeo/schema').DeviceInfoValue, 'name'>>>} opts.getLocalDeviceInfo
```
achou11, line 66: not super happy with the naming. open to suggestions

src/mapeo-project.js
```diff
@@ -48,6 +49,7 @@ export class MapeoProject {
   #coreOwnership
   #capabilities
   #ownershipWriteDone
+  #ownDeviceInfoWriteDone
```
achou11, line 52: open to suggestions for renaming

gmaclennan (commented):
I think this should be done the "other way around", e.g. it is the responsibility of the MapeoManager class when it creates or adds a project to call `project.setDeviceInfo()`.

As currently written, if we release a future version that allows non-device-owners to change their device info, and a device is running this older version, we will get an loop where the device would keep automatically changing it back.

src/mapeo-project.js
```diff
@@ -103,13 +107,20 @@ export class MapeoProject {
       storage: coreManagerStorage,
       sqlite,
     })
+
+    // TODO: Use discovery key
```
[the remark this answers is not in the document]
achou11: yeah i just realized this distinction (which is why tests are breaking) and i'm currently addressing this!

src/mapeo-manager.js
```diff
@@ -219,10 +219,16 @@ export class MapeoManager {
     // 5. Write project name and any other relevant metadata to project instance
     await project.$setProjectSettings(settings)
 
+    // 6. Write device info into project
+    const deviceInfo = await this.getDeviceInfo()
+    if (deviceInfo.name) {
+      await project.$setOwnDeviceInfo({ name: deviceInfo.name })
+    }
```
achou11, line 226: wondering if this is desired behavior or not. keeping this means that one should call `manager.setDeviceInfo()` before adding or creating a project. Otherwise, attempting to use the `$member` namespace causes issues.

src/mapeo-manager.js
```diff
@@ -365,6 +371,14 @@ export class MapeoManager {
       projectInfo,
     })
 
+    // 5. Write device info into project
+    const deviceInfo = await this.getDeviceInfo()
+
+    if (deviceInfo.name) {
+      const project = await this.getProject(projectPublicId)
+      await project.$setOwnDeviceInfo({ name: deviceInfo.name })
+    }
```
achou11, line 380: same concern as previously stated regarding needing to call `manager.setDeviceInfo()` first

tests/member-api.js
achou11, line 1: found that there was a lot of overlap between the tests here and in various other places (both e2e + unit). It was getting harder to update this because of the trickiness with mocking the various contructor parameters, and felt like i was re-implementing a bad version of MapeoManager + MapeoProject. Decided that moving this to e2e to specifically test the `$member` namespace made more sense (see `test-e2e/members.js`)

test-e2e/members.js
```diff
@@ -0,0 +1,144 @@
...
+
+  const members = await project.$member.getMany()
+
+  t.is(members.length, 1)
+
+  t.alike(members[0], me)
+})
+
+// TODO: Fails. Needs MemberApi.invite() to be updated to write device info record of invited peer
+test('getting others after invite', { todo: true }, async (t) => {
```
achou11, line 37: I think this test fails to run due to some additional work needed in the MemberApi implementation, namely saving the saving the invited member's device info when they accept.

gmaclennan: - DeviceInfo and CoreOwnership records need to be written by the invited peer
- The invitor needs to write the Role record
- They need to sync after invite so each gets these records
-

gmaclennan (commented):
I think actually the member namespace needs to be able to handle a device not having a name, and also no code ownership record existing for a device. It's possible that after an invite, the invite or writes the role record (so the device appears as a member) but sync does not complete, so no device info or code ownership record is received from the invited device. The api then needs to have name as optional, and we need to have a fallback in the UX - maybe just show the device ID (we haven't planned out device IDs yet - there was a discussion somewhere about it, I think base-32 of the truncated device key, maybe hashed too. That can be a separate issue.

achou11:
a note that this is waiting on #326 in order to update the MemberApi implementation to handle what is described in https://github.com/digidem/mapeo-core-next/pull/297#pullrequestreview-1658347562

gmaclennan (changes requested):
a couple of non-blocking suggestions, but I think it would be good to put `project.$setOwnDeviceInfo` behind a private symbol before merging, so we don't expose it to the client API

PR #302, opened 2023-10-03
fix: Fix MDNS errors in tests

gmaclennan:
This PR is an attempt to fix flakey MDNS tests.

Initially trying to add latency to both the UDP socket (used for MDNS messages) and to the TCP connections (used by NoiseStream), but unfortunately adding latency seems to reduce the probability of errors, so it's not helping reproduce the error case.


gmaclennan:
closing in favour of #323

PR #306, opened 2023-10-03
chore: remove ipc wrapper implementation

### LINKED ISSUE #278: Publish IPC wrapper as separate module

achou11:
Closes #278 

now lives in https://www.npmjs.com/package/@mapeo/ipc


gmaclennan:
I was thinking about this more and wondering if it makes sense to keep the server implementation here and only have the client in a separate module? Maybe let's test it how it is and see if it causes any issues when using in mapeo mobile

gmaclennan (approved):

achou11:
> I was thinking about this more and wondering if it makes sense to keep the server implementation here and only have the client in a separate module? Maybe let's test it how it is and see if it causes any issues when using in mapeo mobile

yeah that may make sense. will give it some time before deciding on it

PR #308, opened 2023-10-04
chore: rename to CoreSyncState and update API

gmaclennan:
The name was confusing for me, because we also have a "replication state machine". This "CoreSyncState" exists even when cores are not replicating, so I think it makes more sense to use the terms:

- "sync state" = state of synchronization between cores - are they "in sync" or not
- "replication" = actualy exchange of data between cores.


tests/core-replication-state.js
```diff
@@ -377,17 +383,21 @@ function setPeerWants(crs, peerId, bits) {
 
 /**
  * Wait for update event with a timeout
- * @param {CoreReplicationState} crs
+ * @param {CoreSyncState} crs
```
achou11, line 386: maybe `state` or `coreSyncState` would be better, but at the very least update the abbreviation.

```suggestion
 * @param {CoreSyncState} css
```

tests/core-replication-state.js
```diff
@@ -356,15 +362,15 @@ async function downloadCore(core, bits) {
 
 /**
  *
- * @param {CoreReplicationState} crs
+ * @param {CoreSyncState} crs
```
achou11, line 365: same comment about updating the param name

```suggestion
 * @param {CoreSyncState} css
```

achou11 (changes requested):
technically non-blocking but changes worth including before merging. other than that, lgtm

achou11 (approved):

PR #311, opened 2023-10-04
WIP: send device info on first connect

### LINKED ISSUE #245: Send device name over RPC channel
## Description

We don't want to expose device names in multicast messages for MDNS discovery, so instead we will pass the name as a message over the RPC channel on first connect, so that devices could choose not to send their name to untrusted devices.

## Tasks

- [x] Add method for sending and receiving deviceInfo over the RPC channel
- [ ] Send a device's info on first connect

achou11:
Closes #245

As of now:

- only implements sending the device info on initial connection. does not listen for `device-info` events and do anything with those

Questions:

- does this live at the manager level or project instance level?
  - initially thought manager but now i'm wondering if this is project-based, because we would have the option to selectively send the device info based on whether we have local knowledge of the connected peer (which lives in the project dbs)
- what should the manager do when the rpc instance emits a `device-info` event?
  - are we persisting that info somewhere? e.g.  in the `localDeviceInfo` directory
  
- noticing that the `deviceId` on the `device-info` payload is referring to a peer id as defined by hypercore secret stream or whatever, but `deviceId` in MapeoManager/MapeoProject refers to a value derived from the keyManager instance's identity keypair. is this intentional? if so, how are we supposed to use the `deviceId` from the `device-info` event?

### DIFF
--- a/test-e2e/device-info.js
+++ b/test-e2e/device-info.js
@@ -1,7 +1,8 @@
 import { test } from 'brittle'
 import { KeyManager } from '@mapeo/crypto'
-import { MapeoManager } from '../src/mapeo-manager.js'
+import { MapeoManager, kRPC } from '../src/mapeo-manager.js'
 import RAM from 'random-access-memory'
+import { replicate } from '../tests/helpers/rpc.js'
 
 test('write and read deviceInfo', async (t) => {
   const rootKey = KeyManager.generateRootKey()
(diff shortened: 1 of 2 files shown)


test-e2e/device-info.js
```diff
@@ -19,3 +20,35 @@ test('write and read deviceInfo', async (t) => {
...
+
+  await them.setDeviceInfo({ name: 'them' })
+
+  us[kRPC].on('device-info', (deviceInfo) => {
+    t.is(deviceInfo.deviceId, them.deviceId)
+    t.is(deviceInfo.name, 'them')
+  })
+
+  them[kRPC].on('device-info', (deviceInfo) => {
+    t.is(deviceInfo.deviceId, us.deviceId)
```
achou11, line 49: the `deviceId` checks here fail due to the `deviceInfo.deviceId` representing the hex string of the peer's remote public key, which seems to be derived differently than our `MapeoManager.deviceId` property

achou11: for posterity, this was related to the `replicate` rpc helper, which auto generates the keypair used to derive the device id. in practice, it would be aligned with the keypair used by the manager+projects

PR #313, opened 2023-10-04
feat: Add NamespaceSyncState

### LINKED ISSUE #303: Add new `NamespaceSyncState` class
## Description

We need a class that combines states from all the `CoreReplicationState` classes for a namspace. Should create `CoreReplicationState` instances either when a core is added to CoreManager, or when a peer "have" is received. Should throttle state events, because these are expensive to calculate.

## Tasks

- [ ]

gmaclennan:
This combines all core sync states for a namespace. It listens to
CoreManager for new cores and pre-have messages.

Fixes #303

- [x] Depends on #314 (adding namespace to pre-have messages)


gmaclennan:
Writing a lot of unit tests is hard at this stage, due to the setup required. Instead going to add more e2e tests once the sync controller is written.

achou11 (approved):

PR #314, opened 2023-10-04
feat: add namespace to peer pre-have messages

### LINKED ISSUE #12: track which peers initiate the connection
use `isInitiator` on secret-stream instances

as part of this we should add a test for deterministic deduplication of peers.

### LINKED ISSUE #312: Add namespace to peer pre-have messages
## Description

This seems necessary for creating the `NamespaceSyncState` class (#303)

gmaclennan:
Fixes #312

### DIFF
--- a/src/generated/extensions.d.ts
+++ b/src/generated/extensions.d.ts
@@ -12,7 +12,19 @@ export interface HaveExtension {
     discoveryKey: Buffer;
     start: number;
     encodedBitfield: Buffer;
+    namespace: HaveExtension_Namespace;
 }
+export declare const HaveExtension_Namespace: {
+    readonly auth: "auth";
+    readonly config: "config";
+    readonly data: "data";
+    readonly blobIndex: "blobIndex";
+    readonly blob: "blob";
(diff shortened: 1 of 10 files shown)


src/generated/extensions.d.ts
```diff
@@ -12,7 +12,19 @@ export interface HaveExtension {
...
     encodedBitfield: Buffer;
+    namespace: HaveExtension_Namespace;
 }
+export declare const HaveExtension_Namespace: {
+    readonly auth: "auth";
+    readonly config: "config";
+    readonly data: "data";
+    readonly blobIndex: "blobIndex";
+    readonly blob: "blob";
+    readonly UNRECOGNIZED: "UNRECOGNIZED";
```
achou11, line 23: Just for my own understanding, this is added in by the ts-proto step because of the `enumsAsLiterals=true` change? isn't immediately clear to me where this comes from but could kind of guess

achou11 (approved):
non-blocking question but otherwise lgtm!

Issue #292, opened 2023-10-02
create Icon http server


tomasciccola:
Modeled similary to how we serve assets from the BlobStore.

tentative API:

```js
get(`/icon/:iconDocId/:size/:pixelDensity`, () => {
  const icon = await iconDataType.getByDocId(iconDocId)
  const variant = findBestVariantMatch(icon.variants, { size, pixelDensity })
  const { coreDiscoveryId, index } = parseVersionId(variant.blobVersionId)
  const core = coreManager.getByDiscoveryKey(coreDiscoveryId)
  headers.set('mime-type', variant.mimeType)
  return core.get(index)
})
```

For forwards compat, bestVariantMatch should always return something that matches the mimetype (svg or png) so that if we drop all the hi-res PNG variants in a future version and do runtime rendering of SVG to PNG, then we can leave one https://github.com/1x pixel density PNG with size=small.

Issue #291, opened 2023-10-02
Client API: Icons


tomasciccola:
tentaive API:

```js
icon.getUrl({ iconId, size, pixelDensity })
```

achou11:
Closing in favor of #333

PR #315, opened 2023-10-04
Feat: Icon HTTP Server

tomasciccola:
This would solve #292


src/mapeo-project.js
```diff
@@ -206,6 +208,13 @@ export class MapeoProject {
       projectId: this.#projectId,
     })
 
+    this.#iconServer = createIconServer({
+      logger: true,
+      prefix: '/icon/',
+      coreManager: this.#coreManager,
+      iconDataType: this.#dataTypes.icon,
+    })
```
achou11, line 216: are we creating an icon server for each project? my initial thought is that this would be handled similarly to how we're planning to do blobs i.e. one shared server with routes prefixed by the kind of server (e.g. `/blob/` or `/icon/`) and the routes accepting a `projectId` param to work across all projects

src/icon-server/fastify-plugin.js
```diff
@@ -0,0 +1,73 @@
...
+  fastify.register(routes, options)
+}
+
+/** @type {import('fastify').FastifyPluginAsync<
+ * Omit<IconServerPluginOpts, 'prefix'>,
+ * import('fastify').RawServerDefault,
+ * import('@fastify/type-provider-typebox').TypeBoxTypeProvider>} */
+async function routes(fastify, options) {
+  fastify.get(
+    '/:iconDocId/:size/:pixelDensity',
```
achou11, line 41: should this be `':projectId/:iconDocId/:size/:pixelDensity'`? my understanding is that the plan is to have a single server instance that will host these routes.

achou11 (changes requested):
have some questions about the end-usage of this plugin:

- my understanding is that the plugin should work across several projects, similar to the blob plugin.

- in terms of design, my thinking is that the plugin should be able to be used like so (in our case this would probably happen in the MapeoManager implementation):

```ts
const fastify = fastify()

fastify.register(BlobServerPlugin, { prefix: '/blob/', ... })
fastify.register(IconServerPlugin, { prefix: '/icon/', ... })
```

src/icon-server/fastify-plugin.js
```diff
@@ -0,0 +1,73 @@
...
+
+/**
+ * @param {import('@mapeo/schema').IconValue['variants']} variants
+ * @param {object} opts
+ * @param {string} opts.size
+ * @param {number} opts.pixelDensity
+ **/
+function findBestVariantMatch(variants, { size, pixelDensity }) {
+  return variants[0]
+}
```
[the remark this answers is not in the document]
tomasciccola: yeah yeah. Most of it is incomplete, this is just a stub so that we please ts. I didn't realize I made a mistake and didn't make it a draft.

src/icon-server/fastify-plugin.js
```diff
@@ -0,0 +1,73 @@
...
+    '/:iconDocId/:size/:pixelDensity',
+    {
+      schema: {
+        params: PARAMS_JSON_SCHEMA,
+      },
+    },
+    async (req, res) => {
+      const { coreManager, iconDataType } = options
+      const { iconDocId, size, pixelDensity } = req.params
+      const icon = await iconDataType.getByDocId(iconDocId)
```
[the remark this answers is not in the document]
achou11: scratch that: thinking that maybe getByDocId implementation should be updated to throw when no match is found, which means you'd be able to keep it like this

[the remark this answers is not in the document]
tomasciccola: heh, nice catch. `core.get`  can also return undefined. So I should handle both cases and return a 404

src/mapeo-project.js
tomasciccola: This first commit just starts to put everything into place by using the blob-server implementation as a reference. 
For now the blob-server is instantiated inside `mapeo-project` so I was assuming it was handled per project. 
But what you say - I think - makes sense: only have one server per MapeoManager and use `:projectId` as a param.

achou11: > For now the blob-server is instantiated inside mapeo-project so I was assuming it was handled per project.

the `BlobApi` is instantiated in each project, but the server isn't. We have yet to actually integrate any server into mapeo manager/project

tomasciccola: Yeah, I think #291 is made to implement that

tomasciccola: mhm, from what I see in `mapeo-project.js` in main, it seems that the server is instantiated on every project (see [here](https://github.com/digidem/mapeo-core-next/blob/9d465d388e76633fe54c0df46d999cb6142f4948/src/mapeo-project.js#L202)). But maybe I'm missing smth and `createBlobServer` acts as a singleton that instantiates the server if it doesn't exists yet??

achou11: ah no, you're right. totally missed that and was only fixated on the blob api instantiation, my bad 😅

tomasciccola: > ah no, you're right. totally missed that and was only fixated on the blob api instantiation, my bad 😅

I mean, it makes sense to have only one server per `MapeoManager` and not per `MapeoProject`. More so if the `projectId` is a param to the server; if not, it kinda feels redundant to pass the `projectId` ? (since we're always talking about the same instance of `MapeoProject` ).

achou11:
my thinking is that each of the Fastify plugins (so both blobs and icons) should accept an option that is either:

1. The manager instance (like you have it so far)
2. An async function called `getProject()`.

The fundamental thing that each plugin eventually needs is to access project-specific resources, so we need to define a way for the plugin to access a project and its API. For this reason, I _think_ 2 makes a little more sense than 1, but either should work.

In my opinion, the end goal is to be able to set up the server in `MapeoManager` like so:

```ts
this.#fastifyServer = new Fastify({...})

// Register the blob routes
this.#fastifyServer.register(BlobServerPlugin, {
  prefix: '/blob/',
  // One of the following...
  // 1. The manager option
  manager: this,
  
  // 2. The getProject option
  getProject: async (projectId) => {
   return this.#getProject(projectId)
  }
)

// Register the icon routes
this.#fastifyServer.register(IconServerPlugin, {
  prefix: '/icon/',
  // One of the following...
  // 1. The manager option
  manager: this,
  
  // 2. The getProject option
  getProject: async (projectId) => {
   return this.#getProject(projectId)
  }
)
```

At a glance, a shortcoming with this approach is that you may need to expose random instance properties that aren't necessarily useful to a frontend client (remember, we're going to proxy this when interfacing with an IPC layer). I _think_ a  solution to that is using the symbol key accessor pattern that we're using to expose certain methods/fields but for internal usage + testing. For example, if you need the core manager instance, you can set it up like so:

```ts
export const kCoreManager = Symbol('coreManager')

class MapeoProject {
  get [kCoreManager]() {
    return this.#coreManager
  }
}

// Somewhere else where you have the project instance:
import {kCoreManager} from '...' // must explicitly import and use this symbol if you want to access that exposed getter
const coreManager = project[kCoreManager]
```

src/icon-server/fastify-plugin.js
```diff
@@ -1,25 +1,28 @@
...
 export default fp(iconServerPlugin, {
   fastify: '4.x',
   name: 'mapeo-icon-server',
 })
 
-const HEX_REGEX_32_BYTES = '^[0-9a-fA-F]{64}$'
-const HEX_STRING_32_BYTES = T.String({ pattern: HEX_REGEX_32_BYTES })
+// const HEX_REGEX_32_BYTES = '^[0-9a-fA-F]{64}$'
+// const HEX_REGEX_26_BYTES = '^[0-9a-fA-F]{52}$'
+const HEX_STRING_32_BYTES = T.String({})
```
tomasciccola, line 13: I commented this regexes 'cause I was getting validation errors and I couldn't bother with regex today.
It seems that the project key is actually 26 bytes? (printing `projectId` that's returned from `manager.createProject()` yields `52`...)
Anyways, there's probably something I'm missing...

achou11: that number seems familiar for some reason, but the project id that's used for the url _should_ be 32 bytes since we're using this from @mapeo/crypto to create them:

https://github.com/digidem/mapeo-crypto/blob/a3ec70b97318260e2f1593d6abbaeda065cc9407/utils.js#L39

achou11: ah yeah the encoding is different, which produces an id of 26 bytes. The projectId here is really the project's "public" id, which is a one-way hashed version of the project's "internal" id. Think it's a matter of updating this regex to assume length of 52 characters

Needs to be done for the blob server plugin too (but that's out of scope here)

tomasciccola: Mhm, following the trail, the `keys.projectKey` on `mapeo-manager:217` seems to be 32-bytes long. Now, after passing it to `projectKeyToPublicId` (which wraps `keyToPublicId`) the resulting string length is 52 (shouldn't it be 64?).
I see in that fn that you linked that it uses some encoding called z-base-64. I'm not familiar with that, but does that changes something??
I see that I'm on mapeo-crypto-alpha.8 which is the last version, but at the same time on my local version of `mapeo-crypto` that function is named `projectKeyToPublicId` instead of `keyToPublicId` (I've `rm -rf node_modules && npm install` just in case...)

achou11: @tomasciccola yeah not sure if you saw my last comment but the project id we're returning from `createProject` is not a hex-encoded string but rather the z-base-64 one. this is expected, i just forgot that it's a different length and the only change needed here is updating the regex to use 52 instead of 64, since this API would only be accepting "public" ids

achou11: also updating the variable name to something like `PROJECT_PUBLIC_ID_REGEX`, maybe with the addition of a comment to summarize what's been explained here 😄

tomasciccola: yeah, an also include every letter since its not hex (I think only lowercase, but left also upper just to be safe)

gmaclennan:
not essential for this PR, but just a thought on how we expose "getProject". The Fastify "way" is to have a plugin that "decorates" the fastify instance with `fastify.getMapeoProject()`, and making that plugin the dependency of this and the blob http server. It's fine I think how it is implemented here, but if we start moving things into separate files if the code grows then it gets more complicated passing the getProject function to the enclosure (vs. attaching to context as a plugin).

src/icon-server/fastify-plugin.js
```diff
@@ -0,0 +1,62 @@
...
+
+  fastify.get(
+    '/:projectId/:iconDocId/:size/:pixelDensity',
+    { schema: { params: PARAMS_JSON_SCHEMA } },
+    async (req, res) => {
+      const { projectId, iconDocId, size, pixelDensity } = req.params
+      const project = await getProject(projectId)
+      const iconDataType = project[kDataTypes].icon
+      const iconDataStore = project[kDataStores].config
+      const iconApi = new IconApi({ iconDataType, iconDataStore })
```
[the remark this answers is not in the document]
gmaclennan: Should generally avoid creating instances within methods / requests - it has a performance impact.

gmaclennan (changes requested):
Good work. Needs a couple of changes. Is the SQL changed here because of a change in schema in the stacked PR? I don't see any changes to schema in this PR.

src/icon-server/fastify-plugin.js
```diff
@@ -0,0 +1,64 @@
...
+ * Promise<import('../mapeo-project.js').MapeoProject>} GetProject
+ */
+
+/** @type {import('fastify').FastifyPluginAsync<
+ * Omit<IconServerPluginOpts, 'prefix'>,
+ * import('fastify').RawServerDefault,
+ * import('@fastify/type-provider-typebox').TypeBoxTypeProvider>}
+ * */
+async function routes(fastify) {
+  const { getProject } = fastify
```
tomasciccola, line 45: Type error here happens because I should somehow tell typescript that `fastify` param here is decorated with the `getProject` method. 
I tried a couple of things but couldn't make it work...
@gmaclennan ideas??

gmaclennan (changes requested):
all looks good! good work. Hopefully my comment helps resolve the type issue, otherwise just @ts-ignore for now

tomasciccola:
> all looks good! good work. Hopefully my comment helps resolve the type issue, otherwise just @ts-ignore for now

yeah, thanks for that. I saw that fastify documentation, just was trying to avoid using .d.ts and using jsdoc directly...

gmaclennan (approved):
Great job. Lgtm.

achou11 (approved):

src/mapeo-manager.js
```diff
@@ -96,6 +101,20 @@ export class MapeoManager {
...
+      'getProject',
+      /** @param {string} projectId */
+      async (projectId) => {
+        return this.getProject(projectId)
+      }
+    )
+
+    this.#fastifyServer.register(IconServerPlugin, {
+      prefix: '/icons/',
+    })
```
achou11, line 116: didn't realize we were going with a decorator approach, which seems reasonable but maybe not completely necessary imo. with it, looks like we need to declare the dependency on this decorator whenever we register the plugins (https://fastify.dev/docs/latest/Reference/Decorators#decoratename-value-dependencies)

```ts
async function getProjectDecorator (fastify, opts) {
  this.#fastifyServer.decorate(
    'getProject',
    /** @param {string} projectId */
    async (projectId) => {
      return this.getProject(projectId)
    }
  )
}
```

```suggestion
    this.#fastifyServer.register(getProjectDecorator, 'getProject')

    this.#fastifyServer.register(IconServerPlugin, { prefix: '/icons/' }, ['getProject'])
```

types/fastify.d.ts
```diff
@@ -0,0 +1,7 @@
+import { MapeoProject } from '../mapeo-project'
```
achou11, line 1:
```suggestion
import { MapeoProject } from '../src/mapeo-project.js'
```

achou11 (commented):
Couple of small comments on things I noticed (mostly for myself). Keeping in mind that with #365, some other changes are probably going to be needed here, but otherwise generally in a good state

src/mapeo-manager.js
achou11: hm after trying to implement for the blobs plugin, starting to question the benefit of the decorator, especially since you'd have to provide some kind of opt for defining how to get the project anyways. Maybe just keeping it to a `getProject()` opt on the icon/blob plugin is fine as opposed to the decorator?

achou11:
heads up that i did a rebase after #335 was merged into main. entirely possible that I messed some stuff up but hopefully not too badly 😄

achou11 (commented):
Still need to write a couple of unit tests specific to the plugin (e.g. plugin usage and url matching). Thinking about potentially holding off on e2e until this is merged and the media server work is done, but still on the fence about it

gmaclennan (approved):
Looks good to me. lgtm

gmaclennan:
I merged main, since it was a lot of my sync changes on there so thought it easiest for me to resolve the merge conflicts

PR #319, opened 2023-10-04
feat: define sync capability per namespace

### LINKED ISSUE #318: Capabilities should define sync capability per namespace
## Description

For the default capability when syncing with a newly added device (no capability record found), then we want to allow sync of the auth namespace but no other namespaces. Hence we need to the sync capability to be per-namespace. This might also be useful in the future if we want to introduce something like a "config editor" role - someone who can sync and edit the config store, but not the data and blob stores.

## Tasks

- [ ] Update capabilities to make sync a map of namespace to sync permission

gmaclennan:
Fixes #318

### DIFF
--- a/src/capabilities.js
+++ b/src/capabilities.js
@@ -20,7 +20,7 @@ export const BLOCKED_ROLE_ID = '9e6d29263cba36c9'
  * @property {string} name
  * @property {Record<import('@mapeo/schema').MapeoDoc['schemaName'], DocCapability>} docs
  * @property {RoleId[]} roleAssignment
- * @property {'allowed' | 'blocked'} sync
+ * @property {Record<import('./core-manager/core-index.js').Namespace, 'allowed' | 'blocked'>} sync
  */
 
 /**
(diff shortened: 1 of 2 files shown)


src/capabilities.js
```diff
@@ -135,7 +187,9 @@ export class Capabilities {
       if (authCoreId === this.#projectCreatorAuthCoreId) {
         return CREATOR_CAPABILITIES
       } else {
-        return DEFAULT_CAPABILITIES[BLOCKED_ROLE_ID]
+        // When no role assignment exists, e.g. a newly added device which has
+        // not yet synced role records.
+        return NO_ROLE_CAPABILITIES
```
achou11, line 192: looking at this just a few lines above:

```ts
const authCoreId = await this.#coreOwnership.getCoreId(deviceId, 'auth')
```

do you have to wrap that in a try/catch, and return this no role capabilities in the case of the catch? with https://github.com/digidem/mapeo-core-next/pull/317, that call may throw an error if the document isn't found.

Basically, should this block look like this?

```ts
      try {
        const authCoreId = await this.#coreOwnership.getCoreId(deviceId, 'auth')
        if (authCoreId === this.#projectCreatorAuthCoreId) {
          return CREATOR_CAPABILITIES
        } else {
          // When no role assignment exists, e.g. a newly added device which has
          // not yet synced role records.
          return NO_ROLE_CAPABILITIES
        }
      } catch (err) {
        // When no role assignment exists, e.g. a newly added device which has
        // not yet synced role records.
        return NO_ROLE_CAPABILITIES
      }
```

PR #321, opened 2023-10-05
chore: check npm lockfile version does not change

gmaclennan:
running `npm install` with a different version of npm will result in
the lockfileVersion being changes in package-lock.json
This checks for that change in PRs, since we don't want that to happen.


achou11 (approved):
which (minimum) version of npm is this lockfile version associated with? would be nice to make note of this whenever we do a comprehensive update of the development docs

gmaclennan:
I think v9 but unsure

PR #322, opened 2023-10-05
chore: update github action steps

gmaclennan:
Updates to actions/checkout@v4 and actions/setup-node@v2
Removes deprecation warnings

### DIFF
--- a/.github/workflows/node.yml
+++ b/.github/workflows/node.yml
@@ -23,7 +23,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - name: Use Node.js ${{ matrix.node-version }}
-        uses: actions/setup-node@v2
+        uses: actions/setup-node@v3
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/node.yml
```diff
@@ -21,9 +21,9 @@ jobs:
           - '16.x'
           - '18.x'
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v4
       - name: Use Node.js ${{ matrix.node-version }}
-        uses: actions/setup-node@v1
+        uses: actions/setup-node@v2
```
achou11, line 26: any reason to not use v3?

achou11 (approved):
one non-blocking question

.github/workflows/node.yml
gmaclennan: Just an error on my part

PR #323, opened 2023-10-05
fix: fix mdns discovery (finally)

gmaclennan:
The dedupe logic was incorrect, but in most cases the check in
`handleServiceUp` was hiding that. Temporarily removing that check
revealed the bug in a reproducible way.


achou11 (approved):
FINALLY

PR #328, opened 2023-10-05
fix: fix 'write EPIPE' error & failing discovery test

gmaclennan:
Docs on this error:

> EPIPE (Broken pipe): A write on a pipe, socket, or FIFO for which
> there is no process to read the data. Commonly encountered at the net
> and http layers, indicative that the remote side of the stream being
> written to has been closed.

This is happening when we close the connection but there is not yet an
error handler on the socket. It's ok to ignore - so far as the consumer
is concerned we haven't connected yet.

Turns out the other errors seen with discovery were due to not leaving enough delay before checking deduplication of incoming connections. The sodium-native upgrade to v4.0.4 seemed to highlight this because something is slower. It was hard to diagnose because tests were not failing on my (fast) macbook pro.

Also added catching of errors in tests which can happen when connections are deduplicated.

This PR also includes the update to sodium-native@4.0.4, so #338 is no longer needed if this is merged.


achou11 (approved):
looking forward to reduced flakiness 👍

PR #334, opened 2023-10-09
Feat: add DataStore.writeRaw method

### LINKED ISSUE #332: add DataStore `.writeRaw` and `readRaw` methods
## Description
In order to write and read an arbritrary Buffer to the `DataStore` we need a `writeRaw` and `readRaw` methods

## Tasks

- [x] Create `writeRaw` implementation
- [x] Create `readRaw` implementation
- [x] Write unit tests

rough implementation
```js

  /** @param {string} versionId */
  async readRaw(versionId) {
    const { coreDiscoveryKey, index } = parseVersionId(versionId)
    const core = this.#coreManager.getCoreByDiscoveryKey(coreDiscoveryKey)
    if (!core) throw new Error('Invalid versionId')
    const block = await core.get(index, { wait: false })
    if (!block) throw new Error('Not Found')
    return block
  }

 async writeRaw(buf) {
    const { length } = await this.#writerCore.append(buf)
    const index = length - 1
    const coreDiscoveryKey = this.#writerCore.discoveryKey
    if (!coreDiscoveryKey) {
      throw new Error('Writer core is not ready')
    }
    const versionId = getVersionId({ coreDiscoveryKey, index })
    return versionId
  }

```

tomasciccola:
closes #332

### DIFF
--- a/src/datastore/index.js
+++ b/src/datastore/index.js
@@ -1,11 +1,5 @@
 import { TypedEmitter } from 'tiny-typed-emitter'
-import {
-  encode,
-  decode,
-  getVersionId,
-  parseVersionId,
-  getVersionId,
-} from '@mapeo/schema'
+import { encode, decode, getVersionId, parseVersionId } from '@mapeo/schema'
 import MultiCoreIndexer from 'multi-core-indexer'
 import pDefer from 'p-defer'
 import { discoveryKey } from 'hypercore-crypto'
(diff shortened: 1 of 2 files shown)


src/datastore/index.js
```diff
@@ -170,4 +170,26 @@ export class DataStore extends TypedEmitter {
...
+    }
+    const versionId = getVersionId({ coreDiscoveryKey, index })
+    return versionId
+  }
+
+  /** @param {string} versionId */
+  async readRaw(versionId) {
+    const { coreDiscoveryKey, index } = parseVersionId(versionId)
+    const core = this.#coreManager.getCoreByDiscoveryKey(coreDiscoveryKey)
+    if (!core) throw new Error('Invalid versionId')
```
gmaclennan, line 190: This is actually a not found error, because the versionId parsed ok so it is valid.

gmaclennan: This would happen when a peer has not synced the core that has the data.

gmaclennan (approved):
Looks good to me!

PR #335, opened 2023-10-09
Feat: IconApi

### LINKED ISSUE #333: IconApi
We need an `IconApi` that the `IconHTTPServer` talks to, which is different from the `IconClientApi`

rough api
```js
class IconApi {
  constructor({ iconDataType, iconDataStore, coreManager }) {

  }

  /**
   * @param {Omit<IconValue, 'variants'> & { variants: Array<IconVariant & { blob: Buffer }>}} icon
   */
  create(icon) {
    const blobsIds = await Promise.all(icon.variants.map(({ blob }) => {
         // DataStore needs a .writeRaw method to allow writing a Buffer
         return iconDataStore.writeRaw(blob)
    }))
    const variantsWithIds = icon.variants.map((variant, i) => { 
      return { ...variant, blobId: blobIds[i] }
    })
    await this.iconDataType.create({
      schemaName: 'icon',
      name: 'myIcon',
      variants: variantsWithIds,
    })
  }
  getIcon({ iconId, size, pixelDensity }){
    const iconRecord = await this.#dataType.getByDocId(iconId)
    const iconVariant = this.#getBestVariant(iconRecord.variants, {
      size,
      pixelDensity,
    })
    return await this.#dataStore.readRaw(iconVariant.blobVersionId)
  }
}
```

tomasciccola:
stacked on top of  [this PR](https://github.com/digidem/mapeo-core-next/pull/334)

closes #333


src/icon-api.js
```diff
@@ -0,0 +1,72 @@
...
+    this.#dataStore = iconDataStore
+  }
+
+  /**
+   * @param {Omit<import('@mapeo/schema').IconValue, 'variants'> &
+   * {variants: Array<
+   *    import('@mapeo/schema/dist/proto/icon/v1.js').Icon_1_IconVariant
+   *    & {blob: Buffer}>}} icon
+   */
+  async create(icon) {
```
gmaclennan, line 27: maybe add a check that icon.variants.length >= 1

gmaclennan (commented):
Looks good to me, let's add some unit tests before merging.

tomasciccola:
When trying to create tests, I'm getting the following error when creating the instance of `DataType`:
```bash

/home/szgy/data/src/dd/mapeo-core-next/node_modules/better-sqlite3/lib/methods/wrappers.js:5
        return this[cppdb].prepare(sql, this, false);
                           ^
SqliteError: no such table: icon
    at Database.prepare (/home/szgy/data/src/dd/mapeo-core-next/node_modules/better-sqlite3/lib
/methods/wrappers.js:5:21)
    at BetterSQLiteSession.prepareQuery (file:///home/szgy/data/src/dd/mapeo-core-next/node_mod
ules/drizzle-orm/better-sqlite3/index.mjs:17:34)
    at SQLiteSelect.prepare (file:///home/szgy/data/src/dd/mapeo-core-next/node_modules/drizzle
-orm/session-16f863cd.mjs:1112:92)
    at new DataType (file:///home/szgy/data/src/dd/mapeo-core-next/src/datatype/index.js:85:10)
    at file:///home/szgy/data/src/dd/mapeo-core-next/tests/icon-api.js:25:24
    at Test._run (/home/szgy/data/src/dd/mapeo-core-next/node_modules/brittle/index.js:573:13)
    at processTicksAndRejections (node:internal/process/task_queues:96:5) {
  code: 'SQLITE_ERROR'
}
```

I've done `rm -rf drizzle && npm run db:generate:client && npm run db:generate:project` and that didn't solve it. 
When I regenerate the drizzle stuff it seems the `icon` table is created.
 I'm also using the last version of mapeo-schema and the sqlite-indexer which has `icon`
@gmaclennan @achou11 any ideas??

achou11:
@tomasciccola after you create the drizzle instance i.e.

```
const db = drizzle(sqlite)
```

you have to run the migration on it to apply the schema like what we do in `MapeoProject`:

https://github.com/digidem/mapeo-core-next/blob/05b0830d118951fb55ab6b742b242fc0057eecfd/src/mapeo-project.js#L86-L88

tomasciccola:
> @tomasciccola after you create the drizzle instance i.e.
> 
> ```
> const db = drizzle(sqlite)
> ```
> 
> you have to run the migration on it to apply the schema like what we do in `MapeoProject`:
> 
> https://github.com/digidem/mapeo-core-next/blob/05b0830d118951fb55ab6b742b242fc0057eecfd/src/mapeo-project.js#L86-L88

Yeah, thanks I was missing that.

Now I'm getting another error when creating the icon. Pasting the extended commit message:

It seems that when calling `dataStore.write` (wrapped on `iconApi.create`), the method returns the doc by doing `this.getDocById(docId)`. This seems to return nothing (which is weird, if `.write` is succesfull, shoudn't `.getDocById` return something??) 

```bash
file:///home/szgy/data/src/dd/mapeo-core-next/src/datatype/index.js:140
    if (!result) throw new Error('Not found')
                       ^

Error: Not found
    at DataType.getByDocId (file:///home/szgy/data/src/dd/mapeo-core-next/src/datatype/index.js:140:24)
    at DataType.[kCreateWithDocId] (file:///home/szgy/data/src/dd/mapeo-core-next/src/datatype/index.js:132:17)
    at async IconApi.create (file:///home/szgy/data/src/dd/mapeo-core-next/src/icon-api.js:41:12)
    at async file:///home/szgy/data/src/dd/mapeo-core-next/tests/icon-api.js:36:19
    at async Test._run (/home/szgy/data/src/dd/mapeo-core-next/node_modules/brittle/index.js:573:7)

**Comment by achou11** [comment]
sounds vaguely familiar. TL;DR: what I was seeing was that i'd be creating a document but it wasn't being indexed because of a bad id I provided, so reading it using the datatype returned nothing.

---

think in my case, i was passing he wrong doc id when calling datatype create, which caused the catch in the index writer here to trigger:

https://github.com/digidem/mapeo-core-next/blob/c9b135d2dafe6d6379aaa5b1305da51e5416197f/src/index-writer/index.js#L62-L68

this causes the batch call here to await okay, which eventually makes the datastore write resolve:

https://github.com/digidem/mapeo-core-next/blob/c9b135d2dafe6d6379aaa5b1305da51e5416197f/src/datastore/index.js#L107

**Comment by tomasciccola** [comment]
I was forgeting to instance an IndexWriter and doing `batch: async(entries) => indexWriter.batch(entries)`

File: tests/icon-api.js [hunk 2]
```diff
@@ -0,0 +1,129 @@
...
+import { migrate } from 'drizzle-orm/better-sqlite3/migrator'
+
+import IconApi from '../src/icon-api.js'
+import { DataType } from '../src/datatype/index.js'
+import { DataStore } from '../src/datastore/index.js'
+import { createCoreManager } from './helpers/core-manager.js'
+import { iconTable } from '../src/schema/project.js'
+import { IndexWriter } from '../src/index-writer/index.js'
+
+const smallIcon = 'myIconSmall'
```
tomasciccola, line 14: the blobs representing an actual icon are simple strings. Does it make sense to test with an actual icon - that we commit to the repo or load from a url?? -

src/icon-api.js
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  /**
+   * @param {import('@mapeo/schema').IconValue['variants']} variants
+   * @param {object} opts
+   * @param {string} [opts.size]
+   * @param {number} [opts.pixelDensity]
+   * @param {ValidMimeType} [opts.mimeType]
+   **/
+  #getBestVariant(variants, { size, pixelDensity, mimeType }) {
```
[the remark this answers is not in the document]
gmaclennan: What you say makes sense, the implementation doesn't exactly do that from what I can see:

- If the mimeType does not match, in this implementation you will still get back the first variant (we should probably throw / return not found)
- We probably want the variant that is closest in terms of size and pixel density. With the current implementation anything that has one thing matching (e.g. match score 1) will replace the previous one as the "choice", and not sure that is always the best choice?

This is probably worth unit testing - it's the kind of function that could easily have bugs that are hard to track down in other tests.

tests/icon-api.js
```diff
@@ -0,0 +1,164 @@
...
+        size: 'small',
+        pixelDensity: 1,
+        mimeType: 'image/png',
+        blob: Buffer.from(icon),
+      },
+    ],
+  })
+
+  const expectedIcon = await iconApi.getIcon({ iconId: iconDoc.docId })
+  t.is(icon, expectedIcon.toString())
```
gmaclennan: this is kind of the wrong way round (still works, but reads a bit strange). The "expected icon" is `Buffer.from(icon)`. The actual returned icon is what you get back from `iconApi.getIcon()`.

So:

```
t.is(await iconApi.getIcon({ iconId: iconDoc.docId }), Buffer.from('icon'))
```
(feel free to assign to variables too)

tests/icon-api.js
gmaclennan: It's fine, but can be a bit confusing reading these tests. Why not just test with random buffers, e.g. `smallIcon = randomBytes(100)`

gmaclennan: Best to get as close to real use as possible

gmaclennan (changes requested):
Generally looks good, but the best icon algorythm needs some work. The result depends on variant ordering right now, which is not ideal for testing / getting what we expect. Maybe think through what could be returned and what we actually want.

For the tests, it's just semantics, but your naming and ordering of the arguments for test assertions is the wrong way round - the first argument of `t.is()` should be the actual data from the test, and the second argument is the expected value (e.g. what you are expecting to get back - not what you get back from the API call). Also best test with buffers, using `t.alike(buf, expectedbuf)`

src/icon-api.js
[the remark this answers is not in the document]
tomasciccola: Ok. Just to clarify the criteria for this - to see if I'm getting it right:
* Matching mimeType is required, so if asking for something with `mimeType: image/png` that doesn't have a match, we should throw an Error
* regarding `size` and `pixelDensity`, are the two equally important? (in terms of code, do they add the same score when looking for the best match?)
* What happens with a tie? (i.e. we ask for `{pixelDensity:1, size:2}` and the variants that exist are  `[{pixelDensity: 1, size:1}, {pixelDensity: 2, size: 1}]`, which means each variant have one match, resulting en equal score?). We can flip a coin (`Math.random() > 0.5`) ....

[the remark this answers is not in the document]
tomasciccola: Mh, I realize that flipping a coin shouldn't be a possibility - more so with testing... -. We should probably choose between:
1. Weighting the score of `pixelDensity` and `size` differently (would this eliminate ties??)
2. Have a predictable behaviour in case of a tie (choose first? choose last?)

tests/icon-api.js
```diff
@@ -0,0 +1,187 @@
...
+      {
+        size: 'small',
+        pixelDensity: 1,
+        mimeType: 'image/svg+xml',
+        blob: Buffer.from(expectedSmallIcon),
+      },
+    ],
+  })
+
+  t.exception(
```
tomasciccola, line 78: I don't seem to get how to test an exception. Calling that method throws as expected,  but it doesn't produce a passing test, it just terminates the program. If I wrap it in a try catch, I think it just skips it...

tomasciccola: This - I think - the only think 'missing' to be able to merge this. 
Test on icon-api could be improved.
i.e. having fixtures for DRY, and maybe exploring cases that I missed??

tests/icon-api.js
achou11, line 1: would add a `// @ts-check` here, at least while implementing, in order to catch issues. There are a couple related to suggestions I made earlier

tests/icon-api.js
```diff
@@ -0,0 +1,252 @@
...
+test('icon api - create and get with one variant', async (t) => {
+  const iconDoc = await iconApi[kCreate]({
+    name: 'myIcon',
+    schemaName: 'icon',
+    variants: [
+      {
+        size: 'small',
+        pixelDensity: 1,
+        mimeType: 'image/png',
+        blob: Buffer.from(expectedSmallIcon),
```
achou11, line 64: `expectedSmallIcon` is already a buffer
```suggestion
        blob: expectedSmallIcon,
```

tests/icon-api.js
```diff
@@ -0,0 +1,252 @@
...
+      },
+    ],
+  })
+
+  const { icon } = await iconApi[kGetIcon]({ iconId: iconDoc.docId })
+  t.alike(icon, expectedSmallIcon)
+})
+
+test(`icon api - create and fail to find variant with matching mimeType`, async (t) => {
```
achou11, line 73: Here's how you would write the exception assertion:

```ts
  await t.exception(async () => {
    await iconApi[kGetIcon]({
      iconId: iconDoc.docId,
      mimeType: 'image/png',
    })
  }, 'no matching mimeType for icon')
```

tests/icon-api.js
```diff
@@ -0,0 +1,252 @@
...
+        size: 'medium',
+        pixelDensity: 1,
+        mimeType: 'image/svg+xml',
+        blob: Buffer.from(expectedMediumIcon),
+      },
+      {
+        size: 'large',
+        pixelDensity: 2,
+        mimeType: 'image/png',
+        blob: Buffer.from(expectedLargeIcon),
```
achou11, line 118: Can remove unnecessary conversions

```suggestion
        blob: expectedSmallIcon,
      },
      {
        size: 'medium',
        pixelDensity: 1,
        mimeType: 'image/svg+xml',
        blob: expectedMediumIcon,
      },
      {
        size: 'large',
        pixelDensity: 2,
        mimeType: 'image/png',
        blob: expectedLargeIcon,
```

tests/icon-api.js
```diff
@@ -0,0 +1,252 @@
...
+        size: 'medium',
+        pixelDensity: 1,
+        mimeType: 'image/svg+xml',
+        blob: Buffer.from(expectedMediumIcon),
+      },
+      {
+        size: 'large',
+        pixelDensity: 1,
+        mimeType: 'image/png',
+        blob: Buffer.from(expectedLargeIcon),
```
achou11, line 151: Can remove unnecessary conversions:

```suggestion
        blob: expectedSmallIcon,
      },
      {
        size: 'medium',
        pixelDensity: 1,
        mimeType: 'image/svg+xml',
        blob: expectedMediumIcon,
      },
      {
        size: 'large',
        pixelDensity: 1,
        mimeType: 'image/png',
        blob: expectedLargeIcon,
```

tests/icon-api.js
```diff
@@ -0,0 +1,252 @@
...
+        size: 'medium',
+        pixelDensity: 1,
+        mimeType: 'image/svg+xml',
+        blob: Buffer.from(expectedMediumIcon),
+      },
+      {
+        size: 'large',
+        pixelDensity: 2,
+        mimeType: 'image/png',
+        blob: Buffer.from(expectedLargeIcon),
```
achou11, line 186: Can remove unnecessary conversions:

```suggestion
        blob: expectedSmallIcon,
      },
      {
        size: 'medium',
        pixelDensity: 1,
        mimeType: 'image/svg+xml',
        blob: expectedMediumIcon,
      },
      {
        size: 'large',
        pixelDensity: 2,
        mimeType: 'image/png',
        blob: expectedLargeIcon,
```

tests/icon-api.js
```diff
@@ -0,0 +1,252 @@
...
+test(`icon api - getIconUrl, test matching url`, async (t) => {
+  const iconDoc = await iconApi[kCreate]({
+    name: 'myIcon',
+    schemaName: 'icon',
+    variants: [
+      {
+        size: 'small',
+        pixelDensity: 1,
+        mimeType: 'image/png',
+        blob: Buffer.from(expectedSmallIcon),
```
achou11, line 209:
```suggestion
        blob: expectedSmallIcon,
```

tests/icon-api.js
```diff
@@ -0,0 +1,252 @@
...
+  const expectedUrl = `/${projectId}/${iconDoc.docId}/${size}/${pixelDensity}`
+  const url = await iconApi[kGetIconUrl]({
+    iconId: iconDoc.docId,
+    size: 'small',
+    pixelDensity: 1,
+  })
+  t.is(url, expectedUrl)
+})
+
+test(`icon api - getBestVariant`, async (t) => {
```
achou11, line 225: would like to see this get more thoroughly tested. potentially a follow-up though.

example cases i can think of:

- no mimetype matches
- exact match
- all same size, but different pixel densities
- all same pixel density, but different sizes

tests/icon-api.js
```diff
@@ -0,0 +1,252 @@
...
+      size: 'medium',
+      pixelDensity: 1,
+      mimeType: 'image/svg+xml',
+      blob: Buffer.from(expectedMediumIcon),
+    },
+    {
+      size: 'large',
+      pixelDensity: 2,
+      mimeType: 'image/png',
+      blob: Buffer.from(expectedLargeIcon),
```
achou11, line 243: these should be `blobVersionId` instead of `blob` based on the `kGetBestVariant` signature

achou11 (changes requested):
generally looking good! lots of suggestions but a good portion are stylistic

tests/icon-api.js
achou11: answered this here: https://github.com/digidem/mapeo-core-next/pull/335#discussion_r1372209201

src/icon-api.js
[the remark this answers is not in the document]
gmaclennan: Not sure `get()` is needed because you would use the HTTP server?

gmaclennan:
Thanks for the thorough review @achou11. I don't think it makes sense for me to step in and add any comments, I'll leave @tomasciccola to review and respond to @achou11's comments and fix as needed. Largely agree with the comments, but haven't looked thoroughly enough to "endorse" them 😆

tomasciccola:
Heh, yeah, that was thorough! I'll try to address each comment separately (when is needed). Some suggestion can be included mostly _as is_

src/icon-api.js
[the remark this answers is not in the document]
tomasciccola: Well, now that I think about it, I have doubts that `create` should be exposed. I mean, when is the case that from the frontend an icon is `created`?  But I'm pretty sure that `getIconUrl` should be exposed to the client.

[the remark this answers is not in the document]
achou11: the `create` can be exposed using the symbol property pattern, which is basically saying "expose this publicly but only for mapeo core usage"

src/icon-api.js
```diff
@@ -0,0 +1,144 @@
...
+   * }} IconVariant */
+
+  /**
+   * @param {import('@mapeo/schema').IconValue['variants']} variants
+   * @param {object} opts
+   * @param {string} [opts.size]
+   * @param {number} [opts.pixelDensity]
+   * @param {ValidMimeType} [opts.mimeType]
+   **/
+  [kGetBestVariant](variants, { size, pixelDensity, mimeType = 'image/png' }) {
```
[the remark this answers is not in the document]
tomasciccola: Yeah, so my confusion I think spans for not knowing what parameters we have control of. 
The implementation here basically 'scores'  each variant given matching `pixelDensity` or `size`; so a variant with:
* same `pixelDensity` -> score of 1
* same `pixelDensity` and `size` -> score of 2
From that we choose the variant with best score and if there's a tie (two variants with same score) we choose the first one in the array.
So, given you comment I think we can simplify it by expecting a matching size and throwing if it doesn't exist (same as we do with `mimeType`).
From that, given we only have one criteria (`pixelDensity`) we can basically choose the variant with matching or closest `pixelDensity`. 
Also we eliminate the posibility of _ties_

Am I making sense??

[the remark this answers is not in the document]
achou11: > Yeah, so my confusion I think spans for not knowing what parameters we have control of.

To be clear, I'm _also_ not sure 😄 My initial comment really stems from me not really knowing real-world cases that would benefit from having a fallback behavior for this, as opposed to just erroring completely.

Also just realized that my reference to MapboxGL doesn't really apply in this case because these icons aren't used on the actual map for Mapeo (they're used for categories). Still a good example of understanding which parameters may not be in our control, I guess...

@gmaclennan if you have any additions to this, very much welcomed

> So, given you comment I think we can simplify it by expecting a matching size and throwing if it doesn't exist (same as we do with mimeType).
> From that, given we only have one criteria (pixelDensity) we can basically choose the variant with matching or closest pixelDensity.
> Also we eliminate the posibility of ties

if my assumption about the `size` and `pixelDensity` parameters are okay, then yeah I _think_ this makes sense.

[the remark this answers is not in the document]
tomasciccola: Do you think @achou11 that moving `getBestVariant` to `src/utils.js` makes sense??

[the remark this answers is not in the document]
achou11: No strong opinion but I don' think it's strictly necessary if it's not being used elsewhere. I think exporting as a separate variable from the IconApi file would be fine

src/icon-api.js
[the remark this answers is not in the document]
gmaclennan: create() isn't needed by the front-end for the MVP, but it is necessary for adding icons to the db so I'm fine with making it public straight away even though the front-end won't use it initially.

tests/icon-api.js
tomasciccola: I think some of this cases are implicitly tested in other tests, but yeah; we could add other tests...

src/icon-api.js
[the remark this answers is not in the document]
gmaclennan: - No matching mimetype -> throw notfound
- No matching pixeldensity -> next lowest pixel density -> no lower pd -> next higher pd
- No matching size -> next smaller size -> no smaller size -> next higher size
- Prefer matching size over matching pd

The sizes for icons are relevant for bitmap icons (e.g. PNG), for SVG they are only relevant if the icon creator has chosen to make a different design of icon for display at smaller sizes - e.g. simplified

tomasciccola:
So, the thing missing given the comments is improving the `getBestVariant` implementation and adding more test cases, taking into account the comments above. 
We can merge it _as is_ and come later to improve it. 
I'm not going to merge it since I don't wanna merge it in a rush (I need to leave soon to the airport...), but I'll leave @gmaclennan and/or @achou11 to this.

Additionally [this PR](https://github.com/digidem/mapeo-core-next/pull/315) is ready to merge; after that, all the `Icon` related issues can be closed!

src/icon-api.js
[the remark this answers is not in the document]
achou11: @gmaclennan should all of those parameters be required or not? If not, wondering how the priority changes, or would we set defaults?

Currently the implementation marks mime type, size, and pixel density as optional. Would think that maybe mime type should be required at the very least? If I don't specify size, which size takes priority in the case of many valid matches? Same question for pixel density...

e.g.

```ts
// Assume these two are basically the same except for their size and blobVersionId
const smallVariant = {...}
const mediumVariant = { ... }

// What's the result here?
const result = getBestVariant([smallVariant, mediumVariant], { mimeType: ..., pixelDensity: ... })
```

achou11:
@gmaclennan  Doing some additional work on this and looking through the current implementation, I have some questions that I haven't figured out:

- in the context of serving icons, how are deciding which mime type for an icon we want to look for? I'm noticing that `getIconUrl()` currently doesn't accept a `mimeType` opt so the url doesn't contain any information about which mimetype is desired. 

Given the following sequence:

1. Client creates icon with its variants (let's say one png and one svg) using `iconApi.create()`
2. Client derives icon url using `iconApi.getIconUrl()`. currently would look something like ```/project_id/icon_id/small/1```
3. Icon url is used by some Image element or whatever and makes a request for the icon blob
4. Media server has a route for fulfilling icon requests and uses `[kGetIcon]()`. However, the url only provides the icon id, size, and pixel density. within this method, it attempts to find the best matching variant using `getBestVariant()`, which is heavily reliant on knowing which mime type to look for, but the `mimeType` opt is optional and defaults to `image/png`, which doesn't make sense to me.

Basically, shouldn't the client provide information about which mime type they want when requesting an icon? if that's the case, `[kGetIcon]()` and `getBestVariant()` should have all required params, as opposed to some optionals that have defaults.

gmaclennan:
Yeah that all makes sense to me. I think it makes sense for the client to request what type of icon we want here: bitmap or svg. We can do that via a mime type parameter, or just say type "svg" or "png" (or we can use "bitmap" instead of "png" if we don't want to tie ourselves to that format). One thing to note is that the pixel density is not relevant to SVG, so the parameter type makes sense to me as a discriminate union over the type prop, with pixel density only being a property for type: "png".

gmaclennan:
In terms of how to encode this in the URL, maybe just a file extension? It might be helpful to look at the MapBox icon rest api and see the pattern they use.

achou11:
> We can do that via a mime type parameter, or just say type "svg" or "png" (or we can use "bitmap" instead of "png" if we don't want to tie ourselves to that format). 

hm yeah not sure which way I lean on this. feel like specifying the mime type is a little verbose but it aligns with the schema definition of icon records, so maybe that's okay?

> One thing to note is that the pixel density is not relevant to SVG, so the parameter type makes sense to me as a discriminate union over the type prop, with pixel density only being a property for type: "png".

Ah right. Does that mean that for any SVG icon record, the `pixelDensity` field should be `1` by default? I guess ideally there would be a discriminate union type at the schema level but [currently](https://github.com/digidem/mapeo-schema/blob/afa83dff7266f5d61d2a89193bb9d8de5926447f/schema/icon/v1.json), the pixel density is just defined as a non-optional field that's always present.

achou11:
@gmaclennan higlighting some notable changes in my [recent push](https://github.com/digidem/mapeo-core-next/pull/335#commits-pushed-76eeb5c):

- updated `create()` to just return the doc id as opposed to the whole icon document. seemed more aligned with how we do it elsewhere and didn't seem necessary to return the whole document
- updated `getIconUrl()` to return a url formatted similarly to how mapbox's api does it
- updated `[kGetIcon]()` to `[kGetIconBlob]()` and updated the return type to just return the blob, as opposed to the blob and the mime type, which seemed redundant given you need to specify the mime type and the result has to have a matching mime type
- default the pixel density to 1 for svg variants that are stored in the db. now if you only use the IconApi, there's no way you can create an svg variant with a pixel density other than 1. ideally this wouldn't be necessary because the schema type should be smart enough to know that the pixel density doesn't apply to svgs, as you mentioned. figured pixel density of 1 would be fine for now though

gmaclennan (approved):
Looks good to me. Good work @tomasciccola and @achou11

achou11 (approved):

PR #338, opened 2023-10-11
chore: update sodium-native to latest

achou11:
Similar reasoning as #337 

Updates sodium-native (used by sodium-universal) from 4.0.1 to 4.0.4

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5363,8 +5363,9 @@
       }
     },
     "node_modules/node-gyp-build": {
-      "version": "4.5.0",
-      "license": "MIT",
+      "version": "4.6.1",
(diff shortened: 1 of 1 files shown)


achou11:
the failing CI is a concern here 🤔 seems way flakier than usual - maybe this upgrade _does_ break something??

achou11:
seems like this is causing some of the discovery/mdns tests to fail. kind of surprised since it doesn't look like they changed the underlying libsodium version that's being used 🤔 

@gmaclennan any ideas?

https://github.com/sodium-friends/sodium-native/compare/v4.0.1...v4.0.4

gmaclennan:
Replaced by #328 which fixes CI errors from tests

PR #339, opened 2023-10-16
feat: NamespaceSyncState improvements

gmaclennan:
1. Cache sync state
2. Sum peer sync states in NamespaceSyncState

Had originally omitted the sums of peer sync states for cores in a namespace because we don't need it for the UI, but it turns out we need it for the internal sync control because we need the state of each peer before we allow moving to the next stage of sync.

### DIFF
--- a/src/sync/namespace-sync-state.js
+++ b/src/sync/namespace-sync-state.js
@@ -22,6 +22,8 @@ export class NamespaceSyncState {
   #coreStates = new Map()
   #handleUpdate
   #namespace
+  /** @type {SyncState | null} */
+  #cachedState = null
 
   /**
    * @param {object} opts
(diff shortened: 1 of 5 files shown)


src/sync/namespace-sync-state.js
```diff
@@ -54,15 +52,23 @@ export class NamespaceSyncState {
...
+        coreState.remoteStates
+      )) {
+        if (!(peerId in state.remoteStates)) {
+          state.remoteStates[peerId] = createPeerState(peerCoreState.connected)
+        }
+        mutatingAddPeerState(state.remoteStates[peerId], peerCoreState)
+      }
     }
     return state
   }
```
achou11, line 74: is `this.#cachedState` supposed to be set in this function? not seeing that being done anywhere in this class despite the PR description

achou11 (changes requested):
one comment about the caching that may require changes

src/sync/namespace-sync-state.js
gmaclennan: Good catch!

achou11:
looks like tests are failing for something legitimate

gmaclennan:
> looks like tests are failing for something legitimate

oops sorry I should have run the whole test suite. My change to a test helper function was breaking a test elsewhere in the code.

PR #342, opened 2023-10-19
chore: update @mapeo/sqlite-indexer

achou11:
Upgrades @mapeo/sqlite-indexer from 1.0.0-alpha.6 to 1.0.0-alpha.7

Not sure if anyone is aware of any tests that would benefit from being updated as a result of this change.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,7 +15,7 @@
         "@hyperswarm/secret-stream": "^6.1.2",
         "@mapeo/crypto": "^1.0.0-alpha.10",
         "@mapeo/schema": "^3.0.0-next.11",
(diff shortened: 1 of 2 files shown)


gmaclennan:
Actually I seeing this made me realize that we should revert the change in @mapeo/sqlite-indexer, because `deleted` is not related to indexing, and the fact that the deletion tests work fine without this being updated shows it's not necessary. Therefore updating this dep is not necessary, since we will revert and release alpha.8 which will be the same as alpha.6. Separately, we should pin alpha deps - I did not realize that npm would resolve alpha deps to the latest alpha by default (that is why we are getting an error when using the published version of @mapeo/core)

PR #345, opened 2023-10-23
chore: update type-fest and move to normal deps

achou11:
Updates `type-fest` from 3.10.0 to 4.5.0 and moves it to a normal dependency. The latter is necessary for type definitions to work as expected when `@mapeo/core` is used as a dependency in other projects.


gmaclennan (approved):
Looks ok to me. Failing test is due to a sodium-native crash, which might be related to https://github.com/sodium-friends/sodium-native/issues/158 and seems pretty random. Let's keep an eye if this is just node 16, and if so then hopefully the upgrade to node 18 will help.

PR #347, opened 2023-10-24
feat: peer sync controller [1/2]

gmaclennan:
Hypercore manages peers on a "per-hypercore" basis, e.g. each hypercore
has a unique set of peers. However for Mapeo we need to treat all
hypercores belonging to a peer as one, so we need to combine the peer
core sync states across all cores, and then have a means for a
particular peer to control which cores / namespaces are replicated to
the replication stream.

This abstraction is for managing the enabling of namespaces with a
particular peer, and will be managed by the sync controller. This
replaces functionality that is currently in the CoreManager and the
ReplicationStateMachine that is currently returned by
coreManager.replicate().

I will include tests in a stacked pull request that implements SyncController that manages all the PeerSyncController instances.


gmaclennan:
@achou11 I finally sorted this

src/sync/peer-sync-controller.js
```diff
@@ -0,0 +1,278 @@
...
+        const cap = await this.#capabilities.getCapabilities(peerId)
+        this.#syncCapability = cap.sync
+      } catch (e) {
+        // Any error, consider sync blocked
+        this.#syncCapability = createNamespaceMap('blocked')
+      }
+    }
+    // console.log(peerId.slice(0, 7), this.#syncCapability)
+    // console.log(peerId.slice(0, 7), didUpdate)
+    // console.dir(state, { depth: null, colors: true })
```
achou11, line 130:
```suggestion
```

achou11 (approved):
Couldn't find anything noteworthy to point out in the implementation. Seems good to me based on the tests you added!

src/sync/peer-sync-controller.js
gmaclennan: I'll keep these for now for debugging integration tests.

PR #348, opened 2023-10-24
feat: Add SyncController [2/2]

gmaclennan:
The sync controller implements the `replicate()` method, manages sync state, and peer sync controller instances.

This PR is missing tests, but given the complexity of setting up unit tests to invite peers and add data, I think the time is better spent actually implementing and integrating that on the API, so that we can go straight to an e2e test.


achou11 (approved):
Will stopping/disabling sync be a follow-up?

gmaclennan:
Yes good point it should be yes

achou11:
did you mean to merge instead of close?

gmaclennan:
> did you mean to merge instead of close?

No, deleting the base branch automatically closed this (I merged the peer sync controller PR first, which maybe I shouldn't have done - I still haven't figured out the best flow for stacked PRs)

PR #352, opened 2023-10-25
chore: rename MdnsDiscovery to LocalDiscovery

gmaclennan:


achou11 (approved):
maybe also remove the readme from the directory since it seems to be pretty outdated at this point?

PR #354, opened 2023-10-25
fix: Fix sending RPC messages after connection

gmaclennan:
Trying to send an RPC message immediately after connecting the RPC
channel would fail because the protomux channel was not yet open. This
fix ensures that sending an RPC message waits for a peer to fully
connect before trying to send. This is necessary for sending device info
(e.g. the device name) immediately after connecting to a peer.

### DIFF
--- a/src/invite-api.js
+++ b/src/invite-api.js
@@ -109,7 +109,7 @@ export class InviteApi extends TypedEmitter {
 
     if (isAlreadyMember) {
       for (const peerId of peersToRespondTo) {
-        this.#sendAlreadyResponse({ peerId, projectId })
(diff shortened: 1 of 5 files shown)


achou11 (approved):
the async coloring is strong in this PR 😄 are there any other tests that need to be updated as a result? would've expected some of the existing ones around rpc/invites needing it, but maybe not?

gmaclennan:
The only place this affects tests is throwing errors, which are thrown asynchronously now. However it happened that in the tests for the invite API we always returned the method result, so the tests check the async errors fine. The main changes were needed to make the invite API async now.

Issue #316, opened 2023-10-04
send device info on first RPC channel connection


achou11:
## Description

Extracted from #245 

This requires work in the mdns implementation, whose connection events are used to determine when to create the rpc channel connection with the peer

## Tasks

- [ ] Update mdns implementation to create rpc channel when a desired connection is made
- [ ] send device info when rpc channel is created

EvanHahn:
Is this already done [here](https://github.com/digidem/mapeo-core-next/blob/30e637ba74168d106e17d7b3de349abd97f51c74/src/mapeo-manager.js#L270-L273)?

gmaclennan:
yes, done.

PR #356, opened 2023-10-26
feat: MapeoRPC -> LocalPeers

### LINKED ISSUE #355: Expand MapeoRPC to handle connections to local peers, and rename to LocalPeers
## Description

Working on the integration of LocalDiscovery and MapeoRPC for sync, I'm ending up with pieces of code throughout both MapeoManager and MapeoProject tying the pieces together. The sharing of peer names must update a peer list held somewhere, so we can expose `listLocalPeers`.

I think it makes more sense to encapsulate the code in a slightly modified version of the existing `MapeoRPC` class, handling device info events internally to update a property on the peer class and exposing the protomux instance so we can use the peers from here for sync / replication in project instances.

Public methods:
- `sendInvite()`
- `sendInviteResponse()`
- `sendDeviceInfo()`
- `disconnectAllPeers()` -> to be added post-MVP

Events:
- `peers` - when the connected peers changes
- `invite` - when an invite is received

gmaclennan:
Rename, make device info handling internal, add connectedAt and disconnectedAt properties, and expose the peer protomux connection to be used by sync.

This mainly is about renaming things to be more consistent, but also tweaking the API to add features needed for sync integration work. Fixes #355 and also #316

### DIFF
--- a/src/local-peers.js
+++ b/src/local-peers.js
@@ -1,14 +1,14 @@
 // @ts-check
 import { TypedEmitter } from 'tiny-typed-emitter'
 import Protomux from 'protomux'
-import { openedNoiseSecretStream, keyToId } from '../utils.js'
+import { openedNoiseSecretStream, keyToId } from './utils.js'
 import cenc from 'compact-encoding'
 import {
   DeviceInfo,
   Invite,
   InviteResponse,
   InviteResponse_Decision,
-} from '../generated/rpc.js'
+} from './generated/rpc.js'
 import pDefer from 'p-defer'
 
 const PROTOCOL_NAME = 'mapeo/rpc'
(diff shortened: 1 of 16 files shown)


src/local-peers.js
```diff
@@ -44,6 +53,12 @@ class Peer {
   #connected
   /** @type {Map<string, Array<DeferredPromise<InviteResponse['decision']>>>} */
   pendingInvites = new Map()
+  /** @type {string | undefined} */
+  #name
+  #connectedAt = 0
+  #disconnectedAt = 0
```
achou11, line 59: is it too inconvenient to initialize with null? minor concern is that using this technically means January 1, 1970, UTC and having to know that `0` is a magic number for us in this case.

achou11 (approved):
Left a non-blocking comment but overall changes make sense 👍

PR #358, opened 2023-10-26
feat: integrate LocalDiscovery & LocalPeers

gmaclennan:
Stacked on #356

This is a lot of plumbing to connect everything together.

- Replication with local peers now happens at the MapeoManager level, and multiple projects can sync over the same connection (this is done via `protomux.pair()` on the hypercore protocol identifier, which gets us an event with the discovery key whenever a hypercore is replicated, and we pass this discovery key to the project instance via the LocalPeers class.
- Adds a `mapeoManager.replicate()` method, for replicating directly with a stream rather that just using discovery (may move this to a symbol method - would mainly be for testing)
- Narrows types for replication methods to be clearer about what should be passed where. This is a bit confusing, mainly due to the complex actual types used by Hypercore.
- A local discovery instance is now attached to MapeoManager, and new connections are managed by LocalPeers, which sends the device name and implements the RPC methods.

I'm going to do limited testing with this - ensuring that the replicate() method works, and will add more thorough e2e tests once the `listLocalPeers()` method and e2e invites are done.

### DIFF
--- a/src/mapeo-manager.js
+++ b/src/mapeo-manager.js
@@ -17,6 +17,8 @@ import { ProjectKeys } from './generated/keys.js'
 import {
   deNullify,
   getDeviceId,
+  keyToId,
+  openedNoiseSecretStream,
   projectIdToNonce,
   projectKeyToId,
   projectKeyToPublicId,
@@ -24,6 +26,7 @@ import {
 import { RandomAccessFilePool } from './core-manager/random-access-file-pool.js'
 import { LocalPeers } from './local-peers.js'
 import { InviteApi } from './invite-api.js'
+import { LocalDiscovery } from './discovery/local-discovery.js'
 
 /** @typedef {import("@mapeo/schema").ProjectSettingsValue} ProjectValue */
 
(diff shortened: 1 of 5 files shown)


src/mapeo-manager.js
```diff
@@ -263,19 +288,25 @@ export class MapeoManager {
...
+    const project = this.#getProjectInstance(projectKeys)
+
+    // 3. Keep track of project instance as we know it's a properly existing project
+    this.#activeProjects.set(projectPublicId, project)
+
+    return project
+  }
+
+  /** @param {ProjectKeys} projectKeys */
+  #getProjectInstance(projectKeys) {
```
achou11, line 300: nit: this seems more like a create than a get

achou11 (commented):
meant to submit this earlier but forgot. seems to going in a good direction 👍

src/mapeo-manager.js
gmaclennan: Yes, makes sense

achou11 (approved):

PR #360, opened 2023-10-27
feat: `listLocalPeers()` & `local-peers` event

gmaclennan:
Fixes #349

Stacked on #358

Issue #299, opened 2023-10-03
`member.invite()` should throw immediately if device does not have capabilities


gmaclennan:
## Description

Currently calling `member.invite()` when you do not have the capability to invite with the roleId that you try to assign will throw, but only when an invite response is received. `member.invite()` should throw before sending the invite if the device does not have the capability to send the invite

## Tasks

- [ ] Add capability check in `member.invite()` before sending invite.

PR #362, opened 2023-10-27
feat: integrate sync and project invites

### LINKED ISSUE #233: Add project method to download auth + config cores
## Description

Derived from https://github.com/digidem/mapeo-core-next/pull/215#issuecomment-1697609508

When adding a project, there should be a step where we wait to download the auth + config cores for the added project before persisting the project keys to the client database. Without this step, the project is not really usable for the person adding the project.

## Tasks

- [x] Pass an instance of MDNS discovery to `MapeoProject`
- [x] Implement an instance method on `MapeoProject` that downloads the auth + config cores (uses the MDNS discovery instance internally). Not sure what the best naming is for it...

gmaclennan:
This PR integrates work done in #361, #360, #358 and #356 and fixes #233 (Add project method to download auth + config cores).

The `manager.addProject()` method now waits for the config and auth cores to sync before the project is added to the invitee's device, this ensures they have the presets and fields necessary to collect data, and they have the capability records that enable them to sync with others in the project who have not yet synced with the invitor. This does mean that a device could end up invited to the project as far as existing members are concerned, but that project not actually be added to the invitee's device. Have opened a separate issue #363 to address this situation later.

More broadly this PR also integrates earlier sync and discovery work by connecting invites, peer discovery and sync. This integration uncovered several bugs and implementation errors that have been resolved here. This PR also changes how sync is controlled - previously we used `ReplicationStateMachine` and the coreManager.replicate() method, but now we use the new SyncApi methods which authorises peers, starts auth, config and blobIndex sync automatically, and starts data and blob sync when the user calls `project.$sync.start()`. This change broke a lot of unit tests that relied on the core manager sync, so I have implemented a new `coreManager.replicate()` function for tests that replicates all cores in the project. I moved some unit tests to end-to-end tests (e.g. all the PeerSyncController unit tests are now covered in the e2e tests), but more could probably be changed to e2e tests in the future.

The full list of changes and fixes in this PR:

- fix: include the creator capabilities in `capabilities.getAll()`
- fix: track core key requests so that they are only sent once (when a core is added to the replication stream, connected peers will see its discovery key, but they need the core key to sync. They request the key for a given discovery key via an extension message, and the peer will send it if they verify the peer is a member of the project). The key requests are tracked so that if there is no response before a peer disconnects, the key will be re-requested next time it is seen.
- start downloading cores on initialization. This is how [Hypercore does it](https://github.com/holepunchto/hypercore/blob/4096ec0b820a9bdcbb435a9848fa73b4825db33c/index.js#L335) for non-sparse cores. Downloading is not peer specific - it is a state across all replicated peers.
- fix: remove download methods from peerSyncController, because these act on all peer connections so now we just auto-download on initialization (we control downloading by selectively replicating cores).
- chore: set infinite max listeners for many eventEmitters to avoid console warnings, because often every connected peer adds an event listener - this is not a memory leak like the warning suggests.
- chore: normalized some CoreManager methods to return a CoreRecord (hypercore, key and namespace) instead of just a Hypercore.
- chore: removed the previous replication method and the replication state machine from CoreManager, because this is now handled by the sync api.
- fix: moved handleDiscoveryKey to PeerSyncController, because it needs to be aware of sync state to replicate cores that we already have, but are replicated by a peer who has just received them (edge-case when multiple peers are syncing)
- chore: added a replicate function to CoreManager for unit tests - it automatically adds all cores in a project and starts replicating them (via coreStore.replicate).
- fix: fixed a (major) bug where newly received cores were never added to the indexer, so synced data was never indexed.
- feat: created LocalPeers and LocalDiscovery instances on MapeoManager and linked them so that newly connected peers are connected to LocalPeers.
- feat: made the `addProject()` method wait for initial sync - only resolves and adds the project if the role record (authorizing project membership) and project config has synced (see new internal #waitForInitialSync() method)
- feat: new methods to start and stop peer discovery - it did not seem to make sense to include these in `start()` and `stop()` methods that start and stop the server - the server should start and stop when the app goes into the background, which can be controlled by the backend, but discovery should probably be controlled by the user / frontend, so starting and stopping discovery needs to be methods that can be called from the frontend.
- feat: always replicate the creator core with any locally discovered peers, but only attempt to replicate the rest of the project if the connected device can validate that they have the project key (this is validated when trying to replicate the creator core, so any peer that is emitted by `peer-add` from the creator core we know has the project key).
- fix: only send the haves bitfield once the core is ready (otherwise the bitifeld can be empty)
- feat: add `dataToSync: boolean` and `coreCount: number` properties to the sync state


gmaclennan:
Failing tests need to be addressed

achou11 (commented):
Looking good so far! thanks for putting in the extra effort for logging - should be incredibly helpful moving forward 👍

src/local-peers.js
```diff
@@ -59,16 +60,23 @@ class Peer {
...
-  constructor({ publicKey, channel }) {
+  constructor({ publicKey, channel, logger }) {
     this.#publicKey = publicKey
     this.#channel = channel
     this.#connected = pDefer()
+    // @ts-ignore
+    this.#log = (formatter, ...args) => {
+      const log = Logger.create('peer', logger).log
+      return log.apply(null, [`[%h] ${formatter}`, publicKey, ...args])
+    }
```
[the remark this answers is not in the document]
gmaclennan: > what exactly is this ts ignoring here? the arguments not having explicit type annotations?

Yes, it's ignoring not having type annotations, because even adding `any` is complicated for this I think, and didn't seem worth the effort.

> this is basically enabling the usage of the same logger for all peer instances, but adding the appropriate peer-specific info to the log message?

Yes, it's providing a helper function that includes the peerId in each log message, to simplify the logging in the class. When reading these logs we need to know which peer each log message refers to.

test-e2e/manager-invite.js
```diff
@@ -1,166 +1,181 @@
...
+      expectedMembers.sort(memberSort),
+      'Project members match'
+    )
+  }
+
+  await disconnectPeers(managers)
+})
 
-  await creator.setDeviceInfo({ name: 'Creator' })
+// Needs fix to inviteApi to check capabilities before sending invite
```
achou11, line 115: Mostly commenting so that the relevant issue has a reference to this after it's implemented: https://github.com/digidem/mapeo-core-next/issues/299

```suggestion
// Needs fix to inviteApi to check capabilities before sending invite
```

test-e2e/utils-new.js
achou11, line 1: can we merge this with the existing utils file?

achou11 (commented):
Haven't fully wrapped my head around all of the bigger changes but overall seems promising! The e2e tests are reading quite well, which is a good sign!

src/logger.js
```diff
@@ -1,23 +1,45 @@
...
   return discoveryKey(v).toString('hex').slice(0, TRIM)
 }
 
+/**
+ * @param {import('./sync/sync-state.js').State} v
+ * @this {any} */
+createDebug.formatters.X = function (v) {
+  const mapped = mapObject(v, (k, v) => [
+    k,
+    mapObject(v, (k, v) => {
```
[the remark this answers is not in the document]
gmaclennan: I'm going to just ts-ignore this - the type checking doesn't get us anything here, since this is never type checked anyway (v could actually be `any`). We could do full validation of `unknown` to ensure that `v` is `State` and fix all the type issues, but that would really slow down the code (debugging state already causes a massive slowdown), so I think it's better to just ignore errors.

src/mapeo-manager.js
```diff
@@ -456,17 +482,108 @@ export class MapeoManager extends TypedEmitter {
...
+    } = project.$sync.getState()
+    const isCapabilitySynced = capability !== Capabilities.NO_ROLE_CAPABILITIES
+    const isProjectSettingsSynced =
+      projectSettings !== MapeoProject.EMPTY_PROJECT_SETTINGS
+    // Assumes every project that someone is invited to has at least one record
+    // in the auth store - the capability record for the invited device
+    const isAuthSynced = authState.want === 0 && authState.have > 0
+    // Assumes every project that someone is invited to has at least one record
+    // in the config store - defining the name of the project.
+    // TODO: Enforce adding a project name in the invite method
```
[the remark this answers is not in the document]
gmaclennan: done

src/mapeo-project.js
```diff
@@ -461,18 +470,17 @@ export class MapeoProject {
...
    * @returns
    */
   [kProjectReplicate](stream) {
-    const replicationStream = Hypercore.createProtocolStream(stream, {
+    // @ts-expect-error - hypercore types need updating
+    const replicationStream = this.#coreManager.creatorCore.replicate(stream, {
+      // @ts-ignore - hypercore types do not currently include this option
       ondiscoverykey: async (discoveryKey) => {
-        this.#coreManager.handleDiscoveryKey(discoveryKey, replicationStream)
+        this.#syncApi[kHandleDiscoveryKey](discoveryKey, replicationStream)
```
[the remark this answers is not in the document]
gmaclennan: Typing Hypercore is very hard, but this was an actual error too, so fixed and coerced a type.

src/sync/namespace-sync-state.js
```diff
@@ -80,6 +86,7 @@ export class NamespaceSyncState {
    * @param {Buffer} coreKey
    */
   #addCore(core, coreKey) {
+    this.#coreCount++
```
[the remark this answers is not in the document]
gmaclennan: Yes, that would be the correct way to do it, although an error here would crash Mapeo anyway (there shouldn't be a path to an error here unless there is a major bug in the code)

src/logger.js
```diff
@@ -1,23 +1,51 @@
...
-createDebug.formatters.k = (v) => {
+createDebug.formatters.k = function (v) {
   if (!Buffer.isBuffer(v)) return '[undefined]'
   return discoveryKey(v).toString('hex').slice(0, TRIM)
 }
 
+/**
+ * @param {import('./sync/sync-state.js').State} v
+ * @this {any} */
+createDebug.formatters.X = function (v) {
```
achou11, line 26: probably better to do as a follow-up but would be helpful to add a comment about what these formatters should be used for

src/mapeo-manager.js
```diff
@@ -533,6 +651,15 @@ export class MapeoManager extends TypedEmitter {
...
   }
 
+  async startLocalPeerDiscovery() {
+    return this.#localDiscovery.start()
+  }
+
+  /** @type {LocalDiscovery['stop']} */
+  async stopLocalPeerDiscovery(opts) {
+    return this.#localDiscovery.stop(opts)
+  }
```
achou11, line 661: your reasoning for new methods makes sense, but should these still be called in `manager.start()` and `manager.stop()`?

achou11 (approved):
Couple of non-blocking questions but otherwise looks great! Appreciate the detailed summary of changes (made it a lot easier to understand what to look at in the sea of diffs)

src/mapeo-manager.js
achou11: alternatively, wondering if renaming start and stop to be more specific would make sense?

gmaclennan: Yes I'd been meaning to open a PR to call the existing methods "startMediaServer" and "stopMediaServer" or something like that

achou11: created an issue for it: https://github.com/digidem/mapeo-core-next/issues/391

PR #365, opened 2023-11-01
implement media server in MapeoManager

### LINKED ISSUE #329: Set up http server for blobs and icons in MapeoManager
## Description

The plan is to set up a single server to handle http requests for both blobs (e.g. media) and icons (part of project configs). 

The implementation will roughly look similar to this:

```ts
this.#fastifyServer = new Fastify({...})

// Register the blob routes
this.#fastifyServer.register(BlobServerPlugin, {
  prefix: '/blob/',
  // whatever other opts...
)

// Register the icon routes
this.#fastifyServer.register(IconServerPlugin, {
  prefix: '/icon/',
  // whatever other opts...
)
```

## Tasks

- [ ] Create fastify server in `MapeoManager` and register the blob and icon plugins
- [x] Remove blob server creation in `MapeoProject`
- [x] Move blobServer to `MapeoManager` as a Plugin
- [x] Move iconServer to `MapeoManager` as a Plugin (#315 )

achou11:
Closes #329 

Notable changes:

- `MapeoManager`
  - adds `mediaServerOpts` constructor opt for configuring fastify server instance
  -  ~instantiates fastify server and registers blob server plugin~ (abstracted to MediaServer)
  - ~adds `async [kClose]()` method, which is needed for tests in order to properly close the fastify server~ (made into public close method)
- `MapeoProject`
  - removes blob server instantiation
  - adds constructor opt `getMediaBaseUrl(mediaType)`
  - exposes `BlobStore` instance as symbol property (used for fastify plugin registration) 
- `BlobApi`
  - adds constructor opt `getMediaBaseUrl()` and updates implementation of `getUrl()`
  - removes exposure of`BlobStore` instance as public field
- Blob server fastify plugin
  - removes `createBlobServer()` implementation entirely since it was just a convenience around combining fastify with the plugin
  - creates new directory `src/fastify-plugins/` to hold plugins (icons server plugin should go here eventually)
  - updates route param matching for project ids since they are public ids i.e. encoded as z-base-32 52-character string

Added a couple of new tests for the `BlobApi.getUrl()` changes but aside from that, mostly of the changes in the tests are cosmetic or to update given changed signatures.

[EDIT 1]

- After initial feedback:

  - introduces `MediaServer` class
  - introduces `start()` and `stop()` methods on `MapeoManager` class


src/mapeo-manager.js
```diff
@@ -425,4 +484,10 @@ export class MapeoManager {
   get invite() {
     return this.#invite
   }
+
+  async [kClose]() {
```
achou11, line 488: introduced this because it's needed to properly clean up in tests, since we start the fastify server on init. alternatively could provide an opt within `mediaServerOpts` that can be used in tests to change that behavior for tests, and avoid introducing this.

src/mapeo-manager.js
```diff
@@ -112,6 +143,32 @@ export class MapeoManager {
...
   }
 
+  /**
+   * @param {'blobs' | 'icons'} mediaType
+   * @returns
+   */
+  async #getMediaBaseUrl(mediaType) {
+    await this.#deferredMediaServerListen.promise
+
+    let address = this.#mediaServer.server.address()
```
gmaclennan, line 153: Here I would use something like we had in [`getPort()`](https://github.com/digidem/mapeo-core-next/blob/1f4abac31b84fcc5a73a9db2ef7f8fafbffeb3ba/src/blob-server/index.js#L31-L40), probably wrapped in a [p-timeout](https://github.com/sindresorhus/p-timeout). We're going to be starting and stopping the media server whenever the app goes into the background, so the deferred promise will not work.

src/mapeo-manager.js
```diff
@@ -103,6 +115,25 @@ export class MapeoManager {
...
+
+    this.#mediaServer.register(BlobServerPlugin, {
+      prefix: MEDIA_SERVER_BLOBS_PREFIX,
+      getBlobStore: async (projectPublicId) => {
+        const project = await this.getProject(projectPublicId)
+        return project[kBlobStore]
+      },
+    })
+
+    this.#deferredMediaServerListen = pDefer()
```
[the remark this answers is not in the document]
gmaclennan: I think maybe move MediaServer into its own class (since we're also going to add the icon plugin, and possible more in the future), and look at https://github.com/digidem/start-stop-state-machine to help with managing start-stop state, then you can await mediaServer.started() (we will need to start and stop listening on a port when the app is put in the background - so the `mapeoManager.start()` and `mapeoManager.stop()` will need to be exposed, at least to the backend process (ok to expose these to the client too I think). I'll need to attach DNS-SD to these start and stop methods too.

Side-note about this: with your code as-written you would need a `this.#deferredMediaServerListen.promise.catch(noop)` to avoid an uncaught error when you reject this.

src/mapeo-manager.js
```diff
@@ -103,6 +115,25 @@ export class MapeoManager {
...
+    this.#mediaServer
+      .listen({ port: mediaServerOpts?.port, host: '127.0.0.1' })
+      .then(this.#deferredMediaServerListen.resolve)
+      .catch((err) => {
+        console.error('Could not start media server', err)
+        this.#deferredMediaServerListen.reject(err)
+      })
```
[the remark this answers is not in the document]
gmaclennan: I think just logging is ok. If `getMediaBaseUrl` rejects with whatever a failed startup rejects with, that is how the client would know there is an issue.

gmaclennan (commented):
Good work on this. I think the remaining issue is around starting and stopping the media server. Because we are avoiding listening on a port when an app is in the background, the backend implementation on mobile will need to do something like:
```js
rn_bridge.app.on('pause', (pauseLock) => {
  server.close( () => pauseLock.release())
})
```
I suggest just moving mediaserver into its own class, with `start()` and `stop()` methods. In the MapeoManager constructor we can do:
```js
this.#mediaServer = new MediaServer()
this.#mediaServer.start()
```
Although I half think maybe to not start in the constructor and instead require you to call `mapeoManager.start(port?)` first? That's more "proper" but less intuitive from a dev perspective (although working with any server is generally the same - creating an instance does not auto-start)

achou11:
> Although I half think maybe to not start in the constructor and instead require you to call `mapeoManager.start(port?)` first?

yeah i agree. think it's intuitive enough. should the method name be more specific to highlight that it's just the mediaServer that's starting? or will there be other things that this start method will handle later on?

src/media-server.js
```diff
@@ -0,0 +1,148 @@
...
+  async #stopServer() {
+    await this.#fastify.close()
+  }
+
+  /**
+   * @returns {Promise<string>}
+   */
+  async #getAddress() {
+    return pTimeout(getServerAddress(this.#fastify.server), {
+      milliseconds: 1000,
```
achou11, line 68: not sure what the timeout should be here

src/media-server.js
```diff
@@ -0,0 +1,148 @@
...
+    this.#serverState = new StateMachine({
+      start: this.#startServer.bind(this),
+      stop: this.#stopServer.bind(this),
+    })
+  }
+
+  /**
+   * @param {StartOpts} [opts]
+   */
+  async #startServer({ host = '127.0.0.1', port } = {}) {
```
achou11, line 55: does it make sense to bind to `127.0.0.1` by default?

src/media-server.js
```diff
@@ -0,0 +1,148 @@
...
+
+  /**
+   * @param {StartOpts} [opts]
+   */
+  async start(opts) {
+    // Server was stopped before
+    // Need to replace the existing Fastify instance with a new one in order to listen again
+    if (this.#serverState.state.value === 'stopped') {
+      this.#fastify = this.#createFastify()
+    }
```
achou11, line 80: apparently replacing the fastify instance is the only way to re-start the server? maybe there's something i missed in my search but thought there would be something built-in for this (or a plugin)

gmaclennan: Best place to search is our own code 🙂 https://github.com/digidem/mapeo-mobile/blob/develop/src/backend/upgrade-manager/upgrade-server.js#L201-L228

src/media-server.js
gmaclennan: Yes that makes sense. This should mean that the server cannot be accessed from another machine on the same network, which is what we want. In the future we can add a debug option to listen on `0.0.0.0` so that we can test from a laptop on the same wifi.

src/media-server.js
gmaclennan: I think that's reasonable

gmaclennan (changes requested):
Looks good, but a few bugs - I added a test to catch them and fixed some of them.
Remaining bug is due to a bug in the blob-api hashing - I'd been meaning to circle back to that because I suspected it wasn't right. You can't pipe through a hash stream because the output is the hash: we were writing the hash to the blob store rather than the file contents. Also see my comment about starting and stopping fastify.

src/blob-api.js
```diff
@@ -64,8 +64,8 @@ export class BlobApi {
         variant: 'original',
         type: blobType,
       },
-      metadata,
-      contentHash
+      metadata
+      // contentHash
```
gmaclennan, line 68: This fixes the e2e media-server test, but it breaks the blob-api test because the hash is incorrect.

src/media-server.js
achou11: running into issues attempting to port that implementation. 

I removed the conditional check here, and in `#startServer()` I have:

```ts
if (this.#fastifyStarted) {
  const { server } = this.#fastify
  await promisify(server.listen.bind(server))(this.#port, this.#host)
} else {
  await this.#fastify.listen({ host: this.#host, port: this.#port })
  this.fastifyStarted = true
}
```

running into a type issue with the `promisify` and after a brief search, my guess is that it's maybe related to `@types/node`. even so, running through a lifecycle test, I still get a fastify error about the server already listening (`FastifyError [Error]: Fastify is already listening`). Guessing there's still something I'm missing...

achou11: i was setting `this.fastifyStarted` instead of `this.#fastifyStarted`... hours wasted on this 🙃

achou11: re: the promisify type error, seems like it's complaining for a potentially legitimate reason?

seems like the callback param of `server.listen()` doesn't follow the `cb: (err, val) => void` convention. turns out that the callback is only ever called on success, so you'll never get a rejection when wrapping with promisify.

https://nodejs.org/docs/latest-v16.x/api/net.html#serverlisten:

> This function is asynchronous. When the server starts listening, the 'listening' event will be emitted. The last parameter callback will be added as a listener for the 'listening' event.

Seems like doing something along the lines of this would be work though: https://github.com/nodejs/node/issues/21482#issuecomment-626025579

src/media-server.js
```diff
@@ -0,0 +1,199 @@
...
+   * This is necessary because a keep-alive connection from another device will
+   * prevent this server from closing. This hook ensures that if this server is
+   * in the "stopping", "stopped" or "error" states, then it responds with the
+   * "Connection: close" header, which tells the keep-alive client to stop. It
+   * also responds with a 503 "Service unavailable" error.
+   *
+   * @param {import('fastify').FastifyRequest} request
+   * @param {import('fastify').FastifyReply} reply
+   */
+  async #onRequestHook(request, reply) {
```
gmaclennan, line 66: I don't fully understand how this can be called, because we call `server.close()` in the same tick as changing state to `stopping`, which [stops the server accepting new connections](https://nodejs.org/api/net.html#net_server_close_callback), so I would have thought we would not receive a request when in the `stopping` or `stopped` states, but I don't fully understand how keep-alive connections work and how `server.close()` interacts with them! Anyway, TL;DR don't understand this enough to comment, but looks smart and looks like it works

tests/blob-api.js
```diff
@@ -1,124 +1,106 @@
...
     },
-    {
-      mimeType: 'image/png',
-    }
+    { mimeType: 'image/png' }
   )
 
   t.is(attachment.driveId, blobStore.writerDriveId)
   t.is(attachment.type, 'photo')
   t.alike(attachment.hash, hash.digest('hex'))
```
gmaclennan, line 38:
```suggestion
//  t.alike(attachment.hash, hash.digest('hex'))
```

maybe just comment this out and address in a follow-up?

src/mapeo-project.js
```diff
@@ -205,18 +213,10 @@ export class MapeoProject {
...
-      logger: true,
-      blobStore: this.#blobStore,
-      prefix: '/blobs/',
-      projectId: this.#projectId,
-    })
-
-    // @ts-ignore TODO: pass in blobServer
     this.$blobs = new BlobApi({
-      projectId: this.#projectId,
+      projectId: this.#projectPublicId,
```
gmaclennan, line 217: Maybe rename this in the PR to avoid confusion later

gmaclennan (changes requested):
Mainly non-blocking comments, but it would be good to temporarily get tests passing and fix projectPublicId naming because merging.

src/media-server.js
achou11: this was actually plucked from something you implemented in mapeo mobile!

https://github.com/digidem/mapeo-mobile/blob/develop/src/backend/upgrade-manager/upgrade-server.js#L107

I actually added it thinking it could help me figure out the issues I was running into, but turns out it wasn't related. can remove this if you think it's not necessary for now, since I was thinking the same thing (at least for the time being)

achou11:
@gmaclennan do you know if we need to declare fastify as a normal dependency as opposed to a peer dep with this PR? my intuition says yes but not entirely sure

gmaclennan:
> @gmaclennan do you know if we need to declare fastify as a normal dependency as opposed to a peer dep with this PR? my intuition says yes but not entirely sure

Yes because we import it. It would only be a peer dep if we were using a fastify instance as a function parameter, but we import and create an instance so that's a direct dependency.

src/media-server.js
achou11: decided to take out for now (https://github.com/digidem/mapeo-core-next/pull/365/commits/3d4bbd431aed20714cc6837443bd1f24b862e81a). can revisit if we end up seeing notable issues later

src/media-server.js
```diff
@@ -0,0 +1,199 @@
...
+    if (!this.#fastifyStarted) {
+      await this.#fastify.listen({ host: this.#host, port: this.#port })
+      this.#fastifyStarted = true
+      return
+    }
+
+    const { server } = this.#fastify
+
+    await new Promise((res, rej) => {
+      server.listen.bind(server)(this.#port, this.#host)
```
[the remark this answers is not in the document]
achou11: addressed via https://github.com/digidem/mapeo-core-next/pull/365/commits/a90efe49e622d7a9278f0fed6de7f5c51b459155

PR #366, opened 2023-11-08
chore: extract and test construction of icon url

achou11:
the original intention was to use this in tests related to the fastify plugin, but didn't actually end up doing so. still something worth isolating and testing though


gmaclennan (approved):
Looks good

PR #367, opened 2023-11-08
fix: fix core storage initialization in MapeoManager

achou11:
Seems like the `createStorage()` is not a static method on `Hypercore` for the version we're using. Can't find any record of that ever existing, and my assumption is that we want to use `Hypercore.defaultStorage()` instead

https://github.com/holepunchto/hypercore/blob/v10.17.0/index.js#L176

ideally should update @digidem/types too


gmaclennan:
My apologies, I had noticed this and fixed in one of my sync PRs, but had not merged until now. The switch to @ts-expect-error makes sense.

gmaclennan (approved):

PR #370, opened 2023-11-09
fix: fix writing to blob store when also calculating content hash

### LINKED ISSUE #368: BlobApi writes hash contents instead of blob contents to BlobStore
## Description

Surfaced by https://github.com/digidem/mapeo-core-next/pull/365#pullrequestreview-1716846341. We need to fix the implementation so that it's not piping the hash into the blob store stream.

## Tasks

- [ ] Fix implementation of `writeFile()` in `BlobApi`
- [ ] Reinstate assertion that fails due to this issue: https://github.com/digidem/mapeo-core-next/blob/9486d8a95d901774a14e0248ea928d849ec5dc82/tests/blob-api.js#L42

achou11:
Closes #368 

Honestly a super naïve attempt but hoping it does the main thing.

### DIFF
--- a/src/blob-api.js
+++ b/src/blob-api.js
@@ -1,9 +1,8 @@
 import fs from 'node:fs'
 import { Transform } from 'node:stream'
-import { pipeline } from 'node:stream/promises'
-import { createHash } from 'node:crypto'
-import sodium from 'sodium-universal'
-import b4a from 'b4a'
+// @ts-ignore - missing from streamx types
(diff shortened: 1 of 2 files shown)


gmaclennan (approved):
Great, that does just what we need. I've been meaning to do some cleanup of this code, so I added that here - I felt like the existing code was a little hard to understand because of the writeFile abstraction, and was inefficient because of serial writes

gmaclennan:
Also switched to streamx since that's what we're using elsewhere for streams

PR #375, opened 2023-11-13
feat: add `project.close()`

### LINKED ISSUE #227: Add project.close()
## Description

`project.close()` should close all storage and remove event listeners so that everything can be garbage collected, and another project instance can be created without trying to access / write to the same resources.

## Tasks

- [x] close multi-core-indexers
- [x] remove listeners on multi-core-indexers (in DataType)
- [x] close indexer storage
- [x] close hypercores
- [x] close hypercore storage
- [x] close project database
- [ ] figure out if anything else needs to be closed

tomasciccola:
should close #227 

I still need to figure out if we should create additional PRs to address closing different parts of a project (i.e. are there parts where we should expose a `close()` method that hasn't been exposed yet?)

### DIFF
--- a/src/mapeo-project.js
+++ b/src/mapeo-project.js
@@ -346,6 +346,24 @@ export class MapeoProject {
   }
 
   /**
+   */
+  async close() {
+    this.#l.log('closing project %h', this.#projectId)
+    // close hypercores for every namespace
+    // remove listeners
+    for await (let namespace of NAMESPACES) {
+      const namespaceCores = this.#coreManager.getCores(namespace)
+      for (let { core } of namespaceCores) {
+        core.removeAllListeners()
+        await core.close()
+      }
+    }
+    // for(let [_, dataType] of Object.entries(this.#dataTypes)){
+    //   // dataType.close()
+    // }
+  }
+
+  /**
    * @param {import('multi-core-indexer').Entry[]} entries
    * @param {{projectIndexWriter: IndexWriter, sharedIndexWriter: IndexWriter}} indexWriters
    */


src/mapeo-project.js
```diff
@@ -345,6 +345,24 @@ export class MapeoProject {
...
+    this.#l.log('closing project %h', this.#projectId)
+    // close hypercores for every namespace
+    // remove listeners
+    for await (let namespace of NAMESPACES) {
+      const namespaceCores = this.#coreManager.getCores(namespace)
+      for (let { core } of namespaceCores) {
+        core.removeAllListeners()
+        await core.close()
+      }
+    }
```
achou11, line 360: this probably can be abstracted to a `close()` method on the `CoreManager`, which will do this and some other things that are set up in that class.

achou11 (changes requested):
Would take a look at the tasks in the related issue as a starting point for things to address in this PR. Non-exhaustive list of tangible changes that probably need to happen:

- closing the project sqlite instance created in `MapeoProject`
- closing the indexer and core manager storage created in `MapeoProject`
- adding a `close()` method (or something simliar) to the following
  - `CoreManager`
  - `DataStore`
  - `IndexWriter`

Also needs tests (probably in e2e?) to test the following, as a minimum:

- subsequent method calls on the project related to the db/events fail after closing
- can re-create a project instance after closing it and call relevant methods without issue

src/mapeo-project.js
tomasciccola: yeah, I've realized that `CoreManager` already has a `close()` method. It basically iterates over every `core` and closes it. 
I've added a `close()` method to the `Datastore` which basically `.close()` and `.removeAllListener()` over the `#coreIndexer`

tomasciccola:
To close the `MulticoreIndexer` storage I _think_ we need to change code on the `MulticoreIndexer` since its the one in charge of instancing a `random-access-storage`. From what I've seen it would imply exposing the `storage` field from the `CoreIndexStream` class and adding a method - maybe `.closeStorage` or inside the already existing `close()` method... - 
```js
// multicoreIndexer.close()
async close(){
...
  this.#coreIndexStreams.map(indexStream => {
    indexStream.storage.close()
  })
}

```
In the `MulticoreIndexer` docs it says that the consumer should be in charge of closing the storage, buut the class itself doesn't expose it and the `random-access-storage` is actually instanced inside the class, so there's no way I've found two close the storage without changing the class...

@gmaclennan

gmaclennan:
Yes this makes sense to me @tomasciccola. I think cleaner to just keep a set of storage instances on the parent class - we create the instance in the "add core" method. Storages need to be closed _after_ the stream is closed (writes happen during the stream destroy).

tomasciccola:
> we create the instance in the "add core" method

Its weird that we also create that storage on the constructor (see [here](https://github.com/digidem/multi-core-indexer/blob/cff11aa081019264e5c37fc39b69f0b04fe9cbec/index.js#L59C5-L59C5)). 
It seems that the `indexStream` can have multiple `indexStream`s (it sound weird, right? but you can see right there a `this.#indexStream.addStream(coreIndexStream)`)

tomasciccola:
I've open an issue [here](https://github.com/digidem/multi-core-indexer/issues/19) to see approaches of exposing storage on the `multi-core-indexer` @gmaclennan

gmaclennan:
> Its weird that we also create that storage on the constructor (see [here](https://github.com/digidem/multi-core-indexer/blob/cff11aa081019264e5c37fc39b69f0b04fe9cbec/index.js#L59C5-L59C5)). It seems that the `indexStream` can have multiple `indexStream`s (it sound weird, right? but you can see right there a `this.#indexStream.addStream(coreIndexStream)`)

Since we're indexing multiple cores, there needs to be storage for index state for each core (index state is a bitfield). We create a stream for each core of all data that exists in the core and all data that is added to the core. The multi core index stream is then the merge of all those core index streams. Storage is created in two places because we add cores to the indexer in two places: the cores that exist when we first create the indexer, then any cores that are added after.

src/mapeo-project.js
```diff
@@ -345,6 +346,20 @@ export class MapeoProject {
     await Promise.all([this.#coreManager.ready(), this.#ownershipWriteDone])
   }
 
+  /**
+   */
+  async close() {
+    this.#l.log('closing project %h', this.#projectId)
+    this.#coreManager.close()
```
gmaclennan, line 353: This is an async function

src/mapeo-project.js
```diff
@@ -345,6 +346,20 @@ export class MapeoProject {
...
+
+    // eslint-disable-next-line no-unused-vars
+    for await (let [_, dataStore] of Object.entries(this.#dataStores)) {
+      await dataStore.close()
+    }
```
gmaclennan, line 358: This is not an async iterator, so no need for `for await`. Also just use `Object.values()`. Also it's not performance critical, but better to close in parallel rather than serial.

```suggestion
    for (const dataStore of Object.values(this.#dataStores)) {
      await dataStore.close()
    }
```

src/mapeo-project.js
gmaclennan, line 353: Better to close everything in parallel. The pattern I normally follow is to create an array like `const closePromises = []` then push each promise to that, then finally `await Promise.all(closePromises)`

gmaclennan (changes requested):
left some comments on what is here. I think there is storage that needs closed / cleaned up?

tomasciccola:
I'm struggling to see how to close the `coreManager` storage from `mapeoProject`. It seems that we pass and destructure `{coreStorage,dbPath}` from `mapeoManager` [here](https://github.com/digidem/mapeo-core-next/blob/19094097c49ec8761c428c9040a672eaa845dd77/src/mapeo-manager.js#L205), where `coreStorage` is basically a path that gets to a `corestore` instance inside `coreManager` [here](https://github.com/digidem/mapeo-core-next/blob/19094097c49ec8761c428c9040a672eaa845dd77/src/core-manager/index.js#L110).
Since we are passing a path, this means that everything gets saved on disc. Is that an `random-access-storage` instance that we should `close()` ? Looking at the api of corestore storage is not exposed and looking at the source, it seems that it uses `Hypercore.defaultStorage` which seems to use `random-access-file`, but I don't seem to get how to access that to be able to close it (maybe corestore manages closing that?)...
Basically it all comes down to the following:
does calling `core.close()` means also closing the storage?
 @gmaclennan

gmaclennan:
> does calling `core.close()` means also closing the storage?

Yes. See https://github.com/holepunchto/hypercore/blob/4096ec0b820a9bdcbb435a9848fa73b4825db33c/index.js#L436-L483 and https://github.com/holepunchto/hypercore/blob/4096ec0b820a9bdcbb435a9848fa73b4825db33c/lib/core.js#L827-L838

`coreManager.close()` already does this, so you just need to close that.

src/mapeo-project.js
```diff
@@ -351,11 +351,11 @@ export class MapeoProject {
...
     this.#coreManager.close()
-
-    // eslint-disable-next-line no-unused-vars
-    for await (let [_, dataStore] of Object.entries(this.#dataStores)) {
-      await dataStore.close()
+    let dataStorePromises = []
+    for (const dataStore of Object.values(this.#dataStores)) {
+      dataStorePromises.push(dataStore.close())
     }
+    Promise.all(dataStorePromises)
```
gmaclennan, line 358: you need to await this.

test-e2e/project-crud.js
```diff
@@ -124,6 +124,37 @@ test('CRUD operations', async (t) => {
...
+      }, 'should fail updating since the project is already closed')
+
+      await st.exception(async () => {
+        for (const value of values) {
+          // @ts-ignore
+          await project[schemaName].create(value)
+        }
+      }, 'should fail creating since the project is already closed')
+
+      await st.exception(async () => {
```
tomasciccola, line 154: trying to see why this test fails since we expect an error here (the db should be close by now and tests show that). I don't know why it actually fails, maybe it doesn't throw an `Error` ? @achou11

achou11: yeah this is specific to brittle, which will fail the assertion since it's a native error that is raised in this case: https://github.com/holepunchto/brittle/tree/5eda53924e146ce99c3ed9f8770703c2c369df64#texceptionpromisefunctionasync-function-error-message

here's an updated version of the assertion that passes:

```js
await st.exception(async () => {
  try {
    await project[schemaName].getMany()
  } catch (err) {
    // A TypeError is originally raised but Brittle fails this assertion for native errors, including the TypeError that's originally raised here
    // https://github.com/holepunchto/brittle/tree/5eda53924e146ce99c3ed9f8770703c2c369df64#texceptionpromisefunctionasync-function-error-message
    throw new Error(err)
  }
}, 'should fail updating since the project is already closed')
 ```

File: test-e2e/project-crud.js [hunk 6]
```diff
@@ -124,6 +124,62 @@ test('CRUD operations', async (t) => {
...
+        const project = await manager.getProject(projectId)
+        // close it
+        await project.close()
+        // re create project
+        const newProjectId = await manager.createProject()
+        const newProject = await manager.getProject(newProjectId)
+        const newValues = new Array(5).fill(null).map(() => {
+          return getUpdateFixture(value)
+        })
+
```
gmaclennan, line 173: This isn't really testing anything - it's closing one project and creating a different one. Is this meant to test closing then re-opening a project? Then it should use the same projectId and check the data that was written before close is still there.

gmaclennan (changes requested):
One small change about removing event listeners - generally `removeAllListeners()` is a bad idea unless we're absolutely sure nothing else could have attached a listener, because if it has then this can lead to hard-to-track-down bugs. It's better to only remove listeners that a class has added within a class `close()` method.

I left a note about the tests too. However I don't think it's a priority to get strong test coverage for `project.close()` at this stage, because it is only used in tests. I would just delete the test that doesn't really test anything.

test-e2e/project-crud.js
tomasciccola: Yeah, I realize this was the purpose of this tests. But I'm wondering how to achieve that. 
If I close a project I'm also closing the storage, and there's not an 'project.open()' or smth similiar. For now I've commented that test...

src/datastore/index.js
```diff
@@ -192,4 +192,9 @@ export class DataStore extends TypedEmitter {
     if (!block) throw new Error('Not Found')
     return block
   }
+
+  async close() {
+    this.#coreIndexer.removeAllListeners()
```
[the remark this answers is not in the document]
tomasciccola: Ok. The MultiCoreIndexer is already doing that, so I'll just remove that line

test-e2e/project-crud.js
achou11: @tomasciccola all of the storage instances will be re-created when you call `manager.getProject(...)`, but one thing that needs to be done is removing the cached project instances from the manager (`#activeProjects`) after closing the project. I have working changes that update the MapeoProject to emit a `'close'` event that we use to do this when creating the project instances in the MapeoManager. can push those changes with an updated working test if that's helpful

achou11: see https://github.com/digidem/mapeo-core-next/pull/375/commits/5fee8466fa877445ec5727e957aeb215956d83b4 for relevant change

src/datastore/index.js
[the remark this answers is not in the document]
achou11: added relevant change in https://github.com/digidem/mapeo-core-next/pull/375/commits/1ad70570a6a3e705d92d4b98c886421b8a71ef63 to remove the listeners added to the `LocalPeers` instance in the constructor. @gmaclennan lmk if this seems like the correct way to go about it

achou11:
@gmaclennan attempted to update the flaky sync test and now running into another flaky thing (for the same test, it seems). any ideas what this is, and is it something that needs to be addressed in this PR?

https://github.com/digidem/mapeo-core-next/actions/runs/7106378606/job/19345663428?pr=375#step:6:2569

gmaclennan:
> @gmaclennan attempted to update the flaky sync test (see [fd83222](https://github.com/digidem/mapeo-core-next/commit/fd8322248b67d6113171374339fc6ea892fee36c)) and now running into another flaky thing (for the same test, it seems). any ideas what this is? haven't been able to reproduce it locally yet, so hard to tell if this is an issue that this PR needs to address or something else

I tried and failed to write a test to catch this - it's really hard to reproduce. I believe what is happening is the hypercores are closing but indexing is still happening, so the indexer attempts to write to a closed hypercore. I've added a possible fix that ensures that the indexing is closed / stopped before we close the cores - basically closing the data stores before we close the core manager.

gmaclennan (approved):

PR #376, opened 2023-11-13
chore: remove unused/unneeded deps

achou11:
just some housekeeping. one thing i wasn't too sure about was removing `fastify` from dev deps, since we already have it as a direct dep.


gmaclennan (changes requested):
Yes fastify (and any dep) should only be in one of dependencies or devDependencies. In fastify's case it's a production `dependency`, so can remote from devDeps.

Otherwise looks good.

PR #378, opened 2023-11-14
fix: handle duplicate connections to LocalPeers

### LINKED ISSUE #374: LocalPeers needs to handle multiple connections per device
## Description

Work on end-to-end tests revealed a bug in how `LocalPeers` works. The class indexes peers by `peerId` (the device ID of the peer), but it's possible to have more than one connection to a peer. This happens when both peers try to connect to each other at the same time. `LocalDiscovery` will de-duplicate these connections, deterministically closing one of them, but `LocalPeers` only keeps a reference to one of the connections (because it only support one connection per device), which can be the closed "deduplicated" connection.

`LocalPeers` needs an internal refactor to handle multiple connections per peer, intelligently select the best connection to use for sending RPC messages, and to abstract some of the complexity of duplicate connections when emitting peers, e.g. only emitting the open connection.

gmaclennan:
Fixes #374

### DIFF
--- a/src/local-peers.js
+++ b/src/local-peers.js
@@ -1,7 +1,7 @@
 // @ts-check
 import { TypedEmitter } from 'tiny-typed-emitter'
 import Protomux from 'protomux'
-import { openedNoiseSecretStream, keyToId } from './utils.js'
+import { keyToId } from './utils.js'
 import cenc from 'compact-encoding'
 import {
   DeviceInfo,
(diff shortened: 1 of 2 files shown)


achou11 (approved):
Fixes generally make sense! Added some non-blocking suggestions

gmaclennan:
> Fixes generally make sense! Added some non-blocking suggestions

Thanks for the suggestions, once tests pass I'll merge.

PR #379, opened 2023-11-15
chore: fix release workflow configuration

### LINKED ISSUE #359: Release workflow being triggered each time a pull request is merged to main
## Description

Seems like the workflow is erroneously being scheduled when we merge non-release PRs into main

examples:

- https://github.com/digidem/mapeo-core-next/actions/runs/6656591076
- https://github.com/digidem/mapeo-core-next/actions/runs/6648503868

Not exactly sure how to fix but wanted to document since I keep getting emails about failed jobs 😄

achou11:
Fixes #359 

Noticed an error about missing a `build` npm script: https://github.com/digidem/mapeo-core-next/actions/runs/6882793274/job/18722037582#step:3:164

> npm ERR! Missing script: "build"

Not entirely sure what the necessary change is, but figured at least removing the step that fails would be a good start.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -48,4 +48,3 @@ jobs:
           npm-token: ${{ secrets.NPM_TOKEN }}
           build-command: |
             npm install
-            npm run build


gmaclennan (approved):
Yes, that should fix it. The necessary build step for npm is in the `prepack` script, which will automatically run when the package is packed for publishing, so removing the line here should just fix it.

PR #381, opened 2023-11-16
chore: pin more hyper deps

achou11:
Tomas was dealing with issues because of a mangled lockfile, which updated a transitive dep (hyperblobs) and caused issues in our tests.

This change won't fix that scenario, but just to avoid other potential headaches in the future, pinning any hyper deps that haven't been pinned yet.


gmaclennan (approved):
Looks good, maybe add hyperblobs as a dep too in order to pin it? And try to include this in the release to avoid issues

PR #383, opened 2023-11-21
chore: fix Brittle type definitions

achou11:
Noticed issues with the definitions for `t.exception` and all of the ones that are coercible

### DIFF
--- a/types/brittle.d.ts
+++ b/types/brittle.d.ts
@@ -1,32 +1,37 @@
 // Types generated from brittle readme with chatgpt
 
 declare module 'brittle' {
+  interface CoercibleAssertion {
+    (actual: any, expected: any, message?: string): void
+    coercively(actual: any, expected: any, message?: string): void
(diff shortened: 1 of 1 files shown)


gmaclennan:
Sorry I've read through this twice now and can't spot the types that have changed, could you explain it to me?

achou11:
> Sorry I've read through this twice now and can't spot the types that have changed, could you explain it to me?

yeah it's fixing `exception.all()` and the ones where you can use `.coercively()` e.g. `is.coercively`. the syntax used in the current implementation (i.e. the string keys) does not properly represent this interface. for example, right now ts won't complain if you do `t['exception.all'](...)` but it doesn't work in brittle when trying to run the test, since you can't access a nested key with this syntax

gmaclennan:
Ah thanks completely missed that! Makes sense now

gmaclennan (approved):

PR #384, opened 2023-11-21
feat: dbMigrationPathOpt - add db migration path to `MapeoProject` and `MapeoManager`

### LINKED ISSUE #341: Add path pointing to db migrations folder as opts for MapeoManager and MapeoProject
## Description

Currently both MapeoManager and MapeoProject make assumptions about where the drizzle migration folders live:

https://github.com/digidem/mapeo-core-next/blob/110d6d8086c3d2e3cb052482420fa8f8256946db/src/mapeo-project.js#L87-L89

https://github.com/digidem/mapeo-core-next/blob/110d6d8086c3d2e3cb052482420fa8f8256946db/src/mapeo-manager.js#L68-L70

This generally is fine with our setup where where we publish the folders as part of the module, but becomes more problematic in cases where bundling is used (as we do on mobile app). To provide more flexibility, we should make these paths constructor opts that are specified for each class, that way the consumer can fully decide where these migration folders live in their application. 

MapeoManager will require an opt for each db e.g. `clientMigrationPath` and `projectMigrationPath` while MapeoProject just needs one for the project db e.g. `projectMigrationPath`.

## Tasks

- [ ] Add constructor opts for client migration and project migration folders for MapeoManager
- [ ] Add constructor opt for project migration folder for MapeoProject

tomasciccola:
Should close #341 

For now:

* add optional opt for projectMigrationFolder to mapeoProject with default value
* add optional opt for projectMigrationFolder and clientMigrationFolder to mapeoManager with default values

Since I put a default value to both params I made both of them optionals, but maybe it not the right approach?
Also the types for both are strings and then turned into a path with `new URL`

### DIFF
--- a/src/mapeo-manager.js
+++ b/src/mapeo-manager.js
@@ -88,8 +88,8 @@ export class MapeoManager extends TypedEmitter {
   constructor({
     rootKey,
     dbFolder,
-    projectMigrationFolder = '../drizzle/project',
-    clientMigrationFolder = '../drizzle/client',
+    projectMigrationsFolder,
+    clientMigrationsFolder,
     coreStorage,
     mediaServerOpts,
   }) {
--- a/src/mapeo-manager.js
+++ b/src/mapeo-manager.js
@@ -67,6 +67,8 @@ export class MapeoManager extends TypedEmitter {
   /** @type {import('./types.js').CoreStorage} */
   #coreStorage
   #dbFolder
+  /** @type {string} */
+  #projectMigrationFolder
   #deviceId
   #localPeers
   #invite
@@ -78,15 +80,25 @@ export class MapeoManager extends TypedEmitter {
    * @param {Object} opts
    * @param {Buffer} opts.rootKey 16-bytes of random data that uniquely identify the device, used to derive a 32-byte master key, which is used to derive all the keypairs used for Mapeo
    * @param {string} opts.dbFolder Folder for sqlite Dbs. Folder must exist. Use ':memory:' to store everything in-memory
+   * @param {string} [opts.projectMigrationFolder] path for drizzle migration folder for project
+   * @param {string} [opts.clientMigrationFolder] path for drizzle migration folder for client
    * @param {string | import('./types.js').CoreStorage} opts.coreStorage Folder for hypercore storage or a function that returns a RandomAccessStorage instance
    * @param {{ port?: number, logger: import('fastify').FastifyServerOptions['logger'] }} [opts.mediaServerOpts]
    */
-  constructor({ rootKey, dbFolder, coreStorage, mediaServerOpts }) {
+  constructor({
+    rootKey,
+    dbFolder,
+    projectMigrationFolder = '../drizzle/project',
+    clientMigrationFolder = '../drizzle/client',
+    coreStorage,
+    mediaServerOpts,
+  }) {
     super()
     this.#keyManager = new KeyManager(rootKey)
     this.#deviceId = getDeviceId(this.#keyManager)
     this.#l = new Logger({ deviceId: this.#deviceId, ns: 'manager' })
     this.#dbFolder = dbFolder
+    this.#projectMigrationFolder = projectMigrationFolder
     const sqlite = new Database(
       dbFolder === ':memory:'
         ? ':memory:'
@@ -94,7 +106,8 @@ export class MapeoManager extends TypedEmitter {
     )
     this.#db = drizzle(sqlite)
     migrate(this.#db, {
-      migrationsFolder: new URL('../drizzle/client', import.meta.url).pathname,
+      migrationsFolder: new URL(clientMigrationFolder, import.meta.url)
+        .pathname,
     })
 
     this.#localPeers = new LocalPeers({ logger: this.#l })
@@ -352,6 +365,7 @@ export class MapeoManager extends TypedEmitter {
     return new MapeoProject({
       ...this.#projectStorage(projectId),
       ...projectKeys,
+      projectMigrationFolder: this.#projectMigrationFolder,
       keyManager: this.#keyManager,
       sharedDb: this.#db,
       sharedIndexWriter: this.#projectSettingsIndexWriter,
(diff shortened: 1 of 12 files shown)


src/mapeo-manager.js
```diff
@@ -67,6 +67,8 @@ export class MapeoManager extends TypedEmitter {
   /** @type {import('./types.js').CoreStorage} */
   #coreStorage
   #dbFolder
+  /** @type {string} */
+  #projectMigrationFolder
```
achou11, line 71:
```suggestion
  #projectMigrationsFolder
```

src/mapeo-manager.js
```diff
@@ -78,23 +80,34 @@ export class MapeoManager extends TypedEmitter {
    * @param {Object} opts
    * @param {Buffer} opts.rootKey 16-bytes of random data that uniquely identify the device, used to derive a 32-byte master key, which is used to derive all the keypairs used for Mapeo
    * @param {string} opts.dbFolder Folder for sqlite Dbs. Folder must exist. Use ':memory:' to store everything in-memory
+   * @param {string} [opts.projectMigrationFolder] path for drizzle migration folder for project
+   * @param {string} [opts.clientMigrationFolder] path for drizzle migration folder for client
```
achou11, line 84: Think we should be explicit about these options and not have them be optional with a default value

```suggestion
   * @param {string} opts.projectMigrationsFolder path for drizzle migrations folder for project database
   * @param {string} opts.clientMigrationsFolder path for drizzle migrations folder for client database
```

src/mapeo-manager.js
```diff
@@ -78,23 +80,34 @@ export class MapeoManager extends TypedEmitter {
...
    * @param {string | import('./types.js').CoreStorage} opts.coreStorage Folder for hypercore storage or a function that returns a RandomAccessStorage instance
    * @param {{ port?: number, logger: import('fastify').FastifyServerOptions['logger'] }} [opts.mediaServerOpts]
    */
-  constructor({ rootKey, dbFolder, coreStorage, mediaServerOpts }) {
+  constructor({
+    rootKey,
+    dbFolder,
+    projectMigrationFolder = '../drizzle/project',
+    clientMigrationFolder = '../drizzle/client',
```
achou11, line 92: Shouldn't use a default value here

```suggestion
    projectMigrationsFolder,
    clientMigrationsFolder,
```

src/mapeo-manager.js
```diff
@@ -78,23 +80,34 @@ export class MapeoManager extends TypedEmitter {
...
+    coreStorage,
+    mediaServerOpts,
+  }) {
     super()
     this.#keyManager = new KeyManager(rootKey)
     this.#deviceId = getDeviceId(this.#keyManager)
     this.#l = new Logger({ deviceId: this.#deviceId, ns: 'manager' })
     this.#dbFolder = dbFolder
+    this.#projectMigrationFolder = projectMigrationFolder
```
achou11, line 101:
```suggestion
    this.#projectMigrationsFolder = projectMigrationsFolder
```

src/mapeo-manager.js
```diff
@@ -78,23 +80,34 @@ export class MapeoManager extends TypedEmitter {
...
     const sqlite = new Database(
       dbFolder === ':memory:'
         ? ':memory:'
         : path.join(dbFolder, CLIENT_SQLITE_FILE_NAME)
     )
     this.#db = drizzle(sqlite)
     migrate(this.#db, {
-      migrationsFolder: new URL('../drizzle/client', import.meta.url).pathname,
+      migrationsFolder: new URL(clientMigrationFolder, import.meta.url)
```
achou11, line 109:
```suggestion
      migrationsFolder: clientMigrationsFolder,
```

src/mapeo-manager.js
```diff
@@ -352,6 +365,7 @@ export class MapeoManager extends TypedEmitter {
     return new MapeoProject({
       ...this.#projectStorage(projectId),
       ...projectKeys,
+      projectMigrationFolder: this.#projectMigrationFolder,
```
achou11, line 368:
```suggestion
      projectMigrationsFolder: this.#projectMigrationsFolder,
```

src/mapeo-project.js
```diff
@@ -69,6 +69,7 @@ export class MapeoProject {
   /**
    * @param {Object} opts
    * @param {string} opts.dbPath Path to store project sqlite db. Use `:memory:` for memory storage
+   * @param {string} [opts.projectMigrationFolder] path for drizzle migration folder for project
```
achou11, line 72:
```suggestion
   * @param {string} opts.projectMigrationsFolder path for drizzle migration folder for project
```

src/mapeo-project.js
```diff
@@ -83,6 +84,7 @@ export class MapeoProject {
    */
   constructor({
     dbPath,
+    projectMigrationFolder = '../drizzle/project',
```
achou11, line 87:
```suggestion
    projectMigrationsFolder,
```

src/mapeo-project.js
```diff
@@ -102,7 +104,8 @@ export class MapeoProject {
     const sqlite = new Database(dbPath)
     const db = drizzle(sqlite)
     migrate(db, {
-      migrationsFolder: new URL('../drizzle/project', import.meta.url).pathname,
+      migrationsFolder: new URL(projectMigrationFolder, import.meta.url)
```
achou11, line 107:
```suggestion
      migrationsFolder: projectMigrationsFolder,
```

achou11 (changes requested):
I don't think the options should be optional with a default value, as it assumes the location of the migrations directory, which is what we're trying to avoid with this change.

Will need to update/fix the tests to account for the suggested changes

src/mapeo-project.js
tomasciccola: do you think the URL should be build before passing to to MapeoManager and MapeoProject (so, in charge of the consumer?) ?

achou11: yep! would say it provides the most flexibility and correctness based on the context of the consumer (e.g. helps to have directory flexibility in the mobile app due to app structure)

PR #385, opened 2023-11-23
fix: fix brittle types

gmaclennan:
stacked on #383 to fix some additional type issues, mainly that the default export and `test.solo()` were not typed correctly, plus some fixes for incorrect docs (types reflect actual source code)

### DIFF
--- a/types/brittle.d.ts
+++ b/types/brittle.d.ts
@@ -47,43 +47,40 @@ declare module 'brittle' {
     timeout(ms: number): void
     comment(message: string): void
     end(): void
-    test(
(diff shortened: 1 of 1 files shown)


achou11 (approved):
we should eventually consider node's built-in runner at this point 😄

gmaclennan:
> we should eventually consider node's built-in runner at this point 😄

I'd be ok with that. I don't think I really use any of the features that brittle has over `tape`, and the different assertion names mean that I always have to look at the docs when writing tests, I can never remember that `is` means `equals`.

PR #389, opened 2023-11-25
fix: wait for index idle before returning data

gmaclennan:
Stacked on #388 -> #386 -> #362

Depends on new multi-core-indexer being published, but otherwise works. This switches to using file storage in tests, to use memory storage set the environment variable `FAST_TESTS=1`.

To cleanup before publishing:

- [x] replace github install of multi-core-indexer with published version
- [x] delete `types/temp.d.ts` (temporary types for multi-core-indexer)


test-e2e/sync.js
```diff
@@ -252,5 +252,9 @@ test('no sync capabilities === no namespaces sync apart from auth', async (t) =>
     t.alike(invitorState[ns].localState, inviteeState[ns].localState)
   }
 
+  // Temp fix until we have .close() method - waits for indexing idle to ensure
+  // we don't close storage in teardown while index is still being written.
+  await Promise.all(projects.map((p) => p.$getProjectSettings()))
```
achou11, line 257: referencing https://github.com/digidem/mapeo-core-next/issues/227 to return to this when implemented

PR #390, opened 2023-11-27
feat: don't send core keys or haves to peers without capabilities

gmaclennan:
This changes how core keys are shared and when "pre-have" messages are shared. "pre-haves" are information about which blocks (data records) a peer has, and they are used to show the user before sync starts how much data there is to sync, but we don't want to leak this data to unauthorized clients.

Currently any peer with the project key can receive keys for all cores in the project, and receive the haves for all cores in the project.

This PR makes the following changes:

- Only auth core keys are sent to connected peers
- All auth core keys are "pushed" (rather than previously they were requested ("pulled") when the peer tried to replicate that core.
- Other core keys are added from core ownership records, but only once we verify that the peer has previously been invited to the project (keys are sent even if they have since been removed from the project, because it can still be useful to have their data)
- Pre-have messages are only sent to peers that have sync capability for a particular namespace.

### DIFF
--- a/src/core-manager/index.js
+++ b/src/core-manager/index.js
@@ -6,9 +6,6 @@ import { HaveExtension, ProjectExtension } from '../generated/extensions.js'
 import { CoreIndex } from './core-index.js'
 import * as rle from './bitfield-rle.js'
 import { Logger } from '../logger.js'
-import { keyToId } from '../utils.js'
-import { discoveryKey } from 'hypercore-crypto'
-import Hypercore from 'hypercore'
 
 export const kCoreManagerReplicate = Symbol('replicate core manager')
 // WARNING: Changing these will break things for existing apps, since namespaces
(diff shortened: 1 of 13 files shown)


src/core-manager/index.js
```diff
@@ -342,69 +322,39 @@ export class CoreManager extends TypedEmitter {
...
-      'Requesting core key for discovery key %h from peer %h',
-      discoveryKey,
-      peerKey
-    )
-    const message = ProjectExtension.fromPartial({
-      wantCoreKeys: [discoveryKey],
-    })
-    this.#projectExtension.send(message, peer)
+  /**
+   * Sends auth core keys to the given peer, skipping any keys that we know the
```
gmaclennan, line 335: I don't think this skips any keys yet, which is fine, probably a perf optimization for later.

RangerMauve:
Is this issue still relevant?

Issue #228, opened 2023-08-27
Wait for indexing to be done before returning data


gmaclennan:
## Description

Any read operations on data types (e.g. `project.observations.getMany()` should wait until all available data has been indexed before returning data. This is particularly important on first start of the app with unindexed data.

One way to implement this is to add a `dataType.onIndexIdle()` function. It should wait for the cores in the dataStore namespace to be `ready()` first, then wait a tick so that indexing can start, then if the index state is not `idle`, wait until it is `idle`. Either that or we should set the initial index state to `indexing`, and set to idle once the core is ready and we know there is no data.

## Tasks

- [ ]

PR #392, opened 2023-11-28
feat: remove project.ready()

gmaclennan:
Re-opening #386 which merged into the wrong branch.

This PR lays some groundwork for #228, removing the `project.ready()` method and the need to await it. Writing the coreownership owner record is moved inside the core ownership function, and reading core ownership data will wait for the core ownership record to be written before reading data.


gmaclennan:
Same as #386 which was approved, so merging without approval here.

PR #393, opened 2023-11-29
Feat: add fields to invites

### LINKED ISSUE #275: Add fields to invites
## Description

As part of the invite that's sent to a peer, we need to include a `role` field that identifies the role associated with the invite. This way the frontend can inform the user about the role they're being invited to join as

Also should ensure that the project has a name before allowing to invite peers

## Tasks

- [x] Add role name field
- [x] Add role description field
- [x] Invitor name
- [x] Assert the project has a name before allowing invite

tomasciccola:
should close #275

### DIFF
--- a/src/capabilities.js
+++ b/src/capabilities.js
@@ -303,3 +303,15 @@ export class Capabilities {
 function isKnownRoleId(roleId) {
   return roleId in DEFAULT_CAPABILITIES
 }
+
+/**
+ * @param {string} name
+ * @returns {RoleId | undefined}
+ */
+export function roleIdFromName(name) {
+  for (const [roleId, caps] of Object.entries(DEFAULT_CAPABILITIES)) {
+    if (name.toLowerCase() === caps.name.toLowerCase()) {
+      return /** @type {RoleId} */ (roleId)
+    }
+  }
+}
--- a/src/capabilities.js
+++ b/src/capabilities.js
@@ -251,7 +251,7 @@ export class Capabilities {
    * capability for the given roleId
    *
    * @param {string} deviceId
-   * @param {keyof DEFAULT_CAPABILITIES} roleId
+   * @param {keyof typeof DEFAULT_CAPABILITIES} roleId
    */
   async assignRole(deviceId, roleId) {
     let fromIndex = 0
--- a/src/capabilities.js
+++ b/src/capabilities.js
@@ -303,15 +303,3 @@ export class Capabilities {
 function isKnownRoleId(roleId) {
   return roleId in DEFAULT_CAPABILITIES
 }
-
-/**
- * @param {string} name
- * @returns {RoleId | undefined}
- */
-export function roleIdFromName(name) {
-  for (const [roleId, caps] of Object.entries(DEFAULT_CAPABILITIES)) {
-    if (name.toLowerCase() === caps.name.toLowerCase()) {
-      return /** @type {RoleId} */ (roleId)
-    }
-  }
-}
--- a/src/invite-api.js
+++ b/src/invite-api.js
@@ -6,7 +6,7 @@ import { projectKeyToId, projectKeyToPublicId } from './utils.js'
 /**
  * @typedef {Object} InviteApiEvents
  *
- * @property {(info: { projectId: string, projectName?: string, peerId: string, roleName: string, roleDescription: string }) => void} invite-received
+ * @property {(info: { projectId: string, projectName?: string, peerId: string, roleName: string, roleDescription: string, invitorName: string }) => void} invite-received
  */
 
 /**
@@ -68,7 +68,7 @@ export class InviteApi extends TypedEmitter {
 
   /**
    * @param {string} peerId
-   * @param {import('./generated/rpc.js').Invite} invite
+   * @param {import('./generated/rpc.js').Invite & { invitorName: string }} invite
    */
   async #handleInvite(peerId, invite) {
     const projectId = projectKeyToId(invite.projectKey)
@@ -96,6 +96,7 @@ export class InviteApi extends TypedEmitter {
       projectName: invite.projectInfo?.name,
       roleName: invite.roleName,
       roleDescription: invite.roleDescription || '',
+      invitorName: invite.invitorName,
     })
   }
 
--- a/src/invite-api.js
+++ b/src/invite-api.js
@@ -6,7 +6,7 @@ import { projectKeyToId, projectKeyToPublicId } from './utils.js'
 /**
  * @typedef {Object} InviteApiEvents
  *
- * @property {(info: { projectId: string, projectName?: string, peerId: string, roleName: string, roleDescription: string, invitorName: string }) => void} invite-received
+ * @property {(info: { projectId: string, projectName?: string, peerId: string, roleName: string, roleDescription: string | undefined, invitorName: string }) => void} invite-received
  */
 
 /**
@@ -95,7 +95,7 @@ export class InviteApi extends TypedEmitter {
       projectId: projectPublicId,
       projectName: invite.projectInfo?.name,
       roleName: invite.roleName,
-      roleDescription: invite.roleDescription || '',
+      roleDescription: invite.roleDescription,
       invitorName: invite.invitorName,
     })
   }
--- a/src/invite-api.js
+++ b/src/invite-api.js
@@ -6,7 +6,7 @@ import { projectKeyToId, projectKeyToPublicId } from './utils.js'
 /**
  * @typedef {Object} InviteApiEvents
  *
- * @property {(info: { projectId: string, projectName?: string, peerId: string }) => void} invite-received
+ * @property {(info: { projectId: string, projectName?: string, peerId: string, roleName: string, roleDescription: string }) => void} invite-received
  */
 
 /**
@@ -94,6 +94,8 @@ export class InviteApi extends TypedEmitter {
       peerId,
       projectId: projectPublicId,
       projectName: invite.projectInfo?.name,
+      roleName: invite.roleName,
+      roleDescription: invite.roleDescription || '',
     })
   }
 
--- a/src/invite-api.js
+++ b/src/invite-api.js
@@ -68,7 +68,7 @@ export class InviteApi extends TypedEmitter {
 
   /**
    * @param {string} peerId
-   * @param {import('./generated/rpc.js').Invite & { invitorName: string }} invite
+   * @param {import('./generated/rpc.js').Invite} invite
    */
   async #handleInvite(peerId, invite) {
     const projectId = projectKeyToId(invite.projectKey)
--- a/src/member-api.js
+++ b/src/member-api.js
@@ -1,11 +1,7 @@
 import { TypedEmitter } from 'tiny-typed-emitter'
 import { InviteResponse_Decision } from './generated/rpc.js'
 import { projectKeyToId } from './utils.js'
-import {
-  DEFAULT_CAPABILITIES,
-  MEMBER_ROLE_ID,
-  COORDINATOR_ROLE_ID,
-} from './capabilities.js'
+import { DEFAULT_CAPABILITIES } from './capabilities.js'
 
 /** @typedef {import('./datatype/index.js').DataType<import('./datastore/index.js').DataStore<'config'>, typeof import('./schema/project.js').deviceInfoTable, "deviceInfo", import('@mapeo/schema').DeviceInfo, import('@mapeo/schema').DeviceInfoValue>} DeviceInfoDataType */
 /** @typedef {import('./datatype/index.js').DataType<import('./datastore/index.js').DataStore<'config'>, typeof import('./schema/client.js').projectSettingsTable, "projectSettings", import('@mapeo/schema').ProjectSettings, import('@mapeo/schema').ProjectSettingsValue>} ProjectDataType */
@@ -63,7 +59,7 @@ export class MemberApi extends TypedEmitter {
    * @returns {Promise<import('./generated/rpc.js').InviteResponse_Decision>}
    */
   async invite(deviceId, { roleId, roleName, roleDescription, timeout }) {
-    if (!(roleId === MEMBER_ROLE_ID || roleId === COORDINATOR_ROLE_ID)) {
+    if (!DEFAULT_CAPABILITIES[roleId]) {
       throw new Error('Invalid role id')
     }
 
@@ -71,9 +67,10 @@ export class MemberApi extends TypedEmitter {
 
     // since we are always getting #ownDeviceId,
     // this should never throw (see comment on getById), but it pleases ts
-    if (!deviceName) throw new Error('Invalid deviceName')
-
-    if (!roleName) throw new Error('Invalid roleId')
+    if (!deviceName)
+      throw new Error(
+        'Internal error trying to read own device name for this invite'
+      )
 
     const projectId = projectKeyToId(this.#projectKey)
     const project = await this.#dataTypes.project.getByDocId(projectId)
--- a/src/member-api.js
+++ b/src/member-api.js
@@ -57,7 +57,7 @@ export class MemberApi extends TypedEmitter {
    * @param {string} deviceId
    *
    * @param {Object} opts
-   * @param {'Coordinator' | 'Member'} opts.roleName
+   * @param {import('./generated/rpc.js').Invite_RoleName} opts.roleName
    * @param {string} [opts.roleDescription]
    * @param {number} [opts.timeout]
    *
--- a/src/member-api.js
+++ b/src/member-api.js
@@ -1,6 +1,7 @@
 import { TypedEmitter } from 'tiny-typed-emitter'
 import { InviteResponse_Decision } from './generated/rpc.js'
 import { projectKeyToId } from './utils.js'
+import { DEFAULT_CAPABILITIES } from './capabilities.js'
 
 /** @typedef {import('./datatype/index.js').DataType<import('./datastore/index.js').DataStore<'config'>, typeof import('./schema/project.js').deviceInfoTable, "deviceInfo", import('@mapeo/schema').DeviceInfo, import('@mapeo/schema').DeviceInfoValue>} DeviceInfoDataType */
 /** @typedef {import('./datatype/index.js').DataType<import('./datastore/index.js').DataStore<'config'>, typeof import('./schema/client.js').projectSettingsTable, "projectSettings", import('@mapeo/schema').ProjectSettings, import('@mapeo/schema').ProjectSettingsValue>} ProjectDataType */
@@ -46,24 +47,47 @@ export class MemberApi extends TypedEmitter {
     this.#dataTypes = dataTypes
   }
 
+  /** @typedef {import('./capabilities.js').RoleId} COSO */
+
+  /** @type {Array<any>}
+   * @property {import('./capabilities.js').RoleId} 0
+   * @property {import('./capabilities.js').Capability} 1 */
+
   /**
    * @param {string} deviceId
    *
    * @param {Object} opts
-   * @param {import('./capabilities.js').RoleId} opts.roleId
+   * @param {'Coordinator' | 'Member'} opts.roleName
+   * @param {string} [opts.roleDescription]
    * @param {number} [opts.timeout]
    *
    * @returns {Promise<import('./generated/rpc.js').InviteResponse_Decision>}
    */
-  async invite(deviceId, { roleId, timeout }) {
+  async invite(deviceId, { roleName, roleDescription, timeout }) {
+    /** @type {[
+     * roleId: import('./capabilities.js').RoleId,
+     * cap: import('./capabilities.js').Capability] | undefined} */
+    // @ts-ignore TODO:  force cast to literal type
+    const entry = Object.entries(DEFAULT_CAPABILITIES).find(
+      /* eslint-disable no-unused-vars */
+      ([roleId, cap]) => cap.name === roleName
+    )
+    if (entry === undefined) throw new Error('invalid roleName')
+    const [roleId] = entry
+
     const projectId = projectKeyToId(this.#projectKey)
 
     const project = await this.#dataTypes.project.getByDocId(projectId)
+
+    if (!project.name)
+      throw new Error('Project must have a name to invite people')
 
     const response = await this.#rpc.invite(deviceId, {
       projectKey: this.#projectKey,
       encryptionKeys: this.#encryptionKeys,
       projectInfo: { name: project.name },
+      roleName,
+      roleDescription,
       timeout,
     })
 
--- a/src/member-api.js
+++ b/src/member-api.js
@@ -59,6 +59,11 @@ export class MemberApi extends TypedEmitter {
    */
   async invite(deviceId, { roleId, roleDescription, timeout }) {
     const roleName = DEFAULT_CAPABILITIES[roleId].name
+    const { name: deviceName } = await this.getById(this.#ownDeviceId)
+
+    // since we are always getting #ownDeviceId,
+    // this should never throw (see comment on getById), but it pleases ts
+    if (!deviceName) throw new Error('Invalid deviceName')
 
     if (!roleName) throw new Error('Invalid roleId')
 
@@ -74,6 +79,7 @@ export class MemberApi extends TypedEmitter {
       projectInfo: { name: project.name },
       roleName,
       roleDescription,
+      invitorName: deviceName,
       timeout,
     })
 
--- a/src/member-api.js
+++ b/src/member-api.js
@@ -1,7 +1,7 @@
 import { TypedEmitter } from 'tiny-typed-emitter'
 import { InviteResponse_Decision } from './generated/rpc.js'
 import { projectKeyToId } from './utils.js'
-import { roleIdFromName } from './capabilities.js'
+import { DEFAULT_CAPABILITIES } from './capabilities.js'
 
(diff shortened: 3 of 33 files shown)


tomasciccola:
the e2e tests that are failing are all related to calling `manager.addProject` which expects an `Invite`. I don't know from the context what the `roleName` (the missing field) should be in any case, or if maybe the signature for that function should `Omit` the `roleName` from `Invite`. 
For now I'll go through every failing test and see if I can sort of guess the context?

tomasciccola:
`test-e2e/members.js` are failing with a timeout. It seems there's a bunch of promises not being resolved (see `test-e2e/utils.js:93`)

achou11:
> `test-e2e/members.js` are failing with a timeout. It seems there's a bunch of promises not being resolved (see `test-e2e/utils.js:93`) The reason seems to be the error check added [here](https://github.com/digidem/mapeo-core-next/blob/d1d398d09ed861c6ed2e2ad08eb00ff5d5cc6c36/src/member-api.js#L80). One solution would be to add a project name on the e2e test itself (haven't found a method to do that besides passing the name to `localPeers.invite()` which is different from `memberApi.invite()` ...) 

The fix for the test would be to update this line in the test so that the project that's created has a name e.g. `await invitor.createProject({ name: 'mapeo' })`:

https://github.com/digidem/mapeo-core-next/blob/d1d398d09ed861c6ed2e2ad08eb00ff5d5cc6c36/test-e2e/members.js#L129

> Another question is, in `ProjectSettings` schema, name is optional. Should we keep it like that?

i think it should be kept that way. i believe we explicitly decided to let the consumer decide the fallback name for projects that don't have a name

tomasciccola:
> The fix for the test would be to update this line in the test so that the project that's created has a name e.g. `await invitor.createProject({ name: 'mapeo' })`:

heh, I missed that method signature :smile:

src/member-api.js
```diff
@@ -46,24 +47,45 @@ export class MemberApi extends TypedEmitter {
...
+   * @property {import('./capabilities.js').RoleId} 0
+   * @property {import('./capabilities.js').Capability} 1 */
+
   /**
    * @param {string} deviceId
    *
    * @param {Object} opts
-   * @param {import('./capabilities.js').RoleId} opts.roleId
+   * @param {import('./generated/rpc.js').Invite_RoleName} opts.roleName
+   * @param {string} [opts.roleDescription]
```
achou11, line 59: Not sure the params should change here. To me, the roleId should still be used here, and the name and description sent with the invite are derived based on the capabilities associated with the roleId.

The other (tangential) uncertainty i have is how this all works with translations. if the name and description are built into the capability definition, how does internationalization work when you want to surface those things are part of the UI? (easier for name than description in that case)

thoughts @gmaclennan ?

achou11 (changes requested):
raised a couple of concerns but seems like the foundational pieces are there! there's a related issue around translations that isn't entirely clear to me just yet, which would be good to get a resolution on

achou11:
is adding the invitor name and id also going to be part of this PR? doesn't seem like it's doing so at the moment, but I think those a later addition to that issue's checklist.

if not, let's make sure to update the PR title and description to be precise about which fields are being added by these changes

tomasciccola:
> is adding the invitor name and id also going to be part of this PR? doesn't seem like it's doing so at the moment, but I think those were a later addition to that issue's checklist.
> 
> if not, let's make sure to update the PR title and description to be precise about which fields are being added by these changes

I must double check with @gmaclennan but I think what we talked about was that sending the invitorId (`peerId`) is redundant since its already sended with the invite? 
Regarding invitorName, I still need to add it. I think we resolved to being the `deviceName` by calling `mapeoProject.getOwnDeviceName`.
If I'm correct I'll update the issue and remove `Invitor Name` from the checklist, does that sound right??

src/member-api.js
tomasciccola: I think what you say makes sense, but syncing this morning with gregor we arrived at that change maybe there's some reason I'm forgetting??

achou11:
> > is adding the invitor name and id also going to be part of this PR? doesn't seem like it's doing so at the moment, but I think those were a later addition to that issue's checklist.
> > if not, let's make sure to update the PR title and description to be precise about which fields are being added by these changes
> 
> I must double check with @gmaclennan but I think what we talked about was that sending the invitorId (`peerId`) is redundant since its already sended with the invite? Regarding invitorName, I still need to add it. I think we resolved to being the `deviceName` by calling `mapeoProject.getOwnDeviceName`. If I'm correct I'll update the issue and remove `Invitor Name` from the checklist, does that sound right??

yeah i think that makes sense 👍

src/member-api.js
achou11: hm i see. if this was discussed already then feel free to disregard my initial comment! i admit the advantage of providing the description is that there's more freedom from the consumer to describe the role/capability, as opposed to us hardcoding it in the capability definition

gmaclennan: Yes good points @achou11, it was late last night when discussing and I hadn't fully wrapped my head around this - I was wondering where the frontend would get the roleId from, but having a fixed role name doesn't allow the frontend to internationalize what is sent in the invite.

1. We need to export the DEFAULT_CAPABILITIES from @mapeo/core so the front end can read those - ideally they are accessible from the mapeo client without needing to do an async call to the backend - maybe an opportunity to use package.exports @achou11 :) 
2. We keep `roleId` as the required param here, and both role name and role description optional. Role name can default to the `name` on the capability record. The frontend can pass a translated role name (the invitor would decide on the translation, rather than the invitee - because in the future capabilities and capability names might not be available on a device before it joins a project)
3. Need to add a check for a valid role ID and throw if invalid.

gmaclennan (changes requested):
a couple of cleanup suggestions, but some requested changes around role names and passing the role ID

gmaclennan:
> If I'm correct I'll update the issue and remove `Invitor Name` from the checklist, does that sound right??

Invitor ID can be removed from the checklist, but invitor name is still needed, as you say, from `mapeoProject.getOwnDeviceName`

src/member-api.js
tomasciccola: If we keep `roleId`, shouldn't we remove `roleName` from params since it can be grabbed from the `roleId`. I dunno about `roleDescription` since `DEFAULT_CAPABILITIES` doesn't have a `roleDescription` (maybe the idea is to add it to that record?)

src/invite-api.js
```diff
@@ -68,7 +68,7 @@ export class InviteApi extends TypedEmitter {
 
   /**
    * @param {string} peerId
-   * @param {import('./generated/rpc.js').Invite} invite
+   * @param {import('./generated/rpc.js').Invite & { invitorName: string }} invite
```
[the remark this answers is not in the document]
tomasciccola: I'm not to sure about this, since it basically means adding `& {invitorName: string}` to a bunch of places and it can become hard to track? If an invite actually contains that field, then maybe we should add it to the proto...
@gmaclennan wdyt??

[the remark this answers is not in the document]
tomasciccola: dismiss this. I realize that since we're sending the invite through a protobuf, this needs to be on the `Invite` proto

src/member-api.js
achou11: > If we keep roleId, shouldn't we remove roleName from params since it can be grabbed from the roleId.

it's still useful for roleName to be an optional param (not a required param) because otherwise we wouldn't be able to easily provide a translated string. Gregor's suggestion is to have the optional name param and if that isn't provided use the name associated with the roleId. For example, an application should be able to do `members.invite({ roleName: "Miembro", ... })` if they want the invited device to see the translated name of the role. Otherwise the default is the english name we're using in the role id definition

tomasciccola:
`test-e2e/sync.js` is failing with `  code: 'SQLITE_READONLY_DBMOVED'`. It happens only on linux and I can repro locally

achou11:
@tomasciccola mind trying this #396 to fix the tests?

https://github.com/digidem/mapeo-core-next/blob/98a6aa70cd9dca452ccb923916c4b8591f144496/test-e2e/sync.js

src/member-api.js
```diff
@@ -51,19 +52,34 @@ export class MemberApi extends TypedEmitter {
...
     const project = await this.#dataTypes.project.getByDocId(projectId)
 
+    if (!project.name)
+      throw new Error('Project must have a name to invite people')
+
     const response = await this.#rpc.invite(deviceId, {
       projectKey: this.#projectKey,
       encryptionKeys: this.#encryptionKeys,
       projectInfo: { name: project.name },
+      roleName,
```
achou11, line 80: would move the check on roleId as the first thing to do (small perf improvement). suggestion also incorporates specifying the roleName as a param as mentioned in my other comment

```suggestion
  async invite(deviceId, { roleId, roleName, roleDescription, timeout }) {
    if (!(roleId in DEFAULT_CAPABILITIES)) throw new Error('Unknown roleId')

    const { name: deviceName } = await this.getById(this.#ownDeviceId)

    // since we are always getting #ownDeviceId,
    // this should never throw (see comment on getById), but it pleases ts
    if (!deviceName) throw new Error('Invalid deviceName')

    const projectId = projectKeyToId(this.#projectKey)
    const project = await this.#dataTypes.project.getByDocId(projectId)

    if (!project.name)
      throw new Error('Project must have a name to invite people')

    const response = await this.#rpc.invite(deviceId, {
      projectKey: this.#projectKey,
      encryptionKeys: this.#encryptionKeys,
      projectInfo: { name: project.name },
      roleName: roleName || DEFAULT_CAPABILITIES[roleId].name,
```

test-e2e/utils.js
```diff
@@ -76,7 +76,7 @@ export function connectPeers(managers, { discovery = true } = {}) {
  *   invitor: MapeoManager,
  *   projectId: string,
  *   invitees: MapeoManager[],
- *   roleId?: import('../src/capabilities.js').RoleId,
+ *   roleId?: import('../src/capabilities.js').RoleId ,
```
achou11, line 79:
```suggestion
 *   roleId?: import('../src/capabilities.js').RoleId,
```

src/member-api.js
```diff
@@ -51,19 +52,34 @@ export class MemberApi extends TypedEmitter {
...
    * @param {Object} opts
    * @param {import('./capabilities.js').RoleId} opts.roleId
+   * @param {string} [opts.roleDescription]
    * @param {number} [opts.timeout]
    *
    * @returns {Promise<import('./generated/rpc.js').InviteResponse_Decision>}
    */
-  async invite(deviceId, { roleId, timeout }) {
-    const projectId = projectKeyToId(this.#projectKey)
+  async invite(deviceId, { roleId, roleDescription, timeout }) {
```
achou11, line 60: think we should also limit the `roleId` to just member and coordinator, as it doesn't make sense to try to invite someone to be blocked, for example. not sure if best to update the param type or do a runtime check. ideally add a test for this too

src/capabilities.js
```diff
@@ -303,3 +303,15 @@ export class Capabilities {
...
+ * @param {string} name
+ * @returns {RoleId | undefined}
+ */
+export function roleIdFromName(name) {
+  for (const [roleId, caps] of Object.entries(DEFAULT_CAPABILITIES)) {
+    if (name.toLowerCase() === caps.name.toLowerCase()) {
+      return /** @type {RoleId} */ (roleId)
+    }
+  }
+}
```
achou11, line 317: no longer needed so can remove
```suggestion
```

achou11 (changes requested):
Overall changes are looking good! still noted a couple of items that should be addressed in this PR and ideally have tests added for them.

src/member-api.js
```diff
@@ -51,19 +56,38 @@ export class MemberApi extends TypedEmitter {
...
+      throw new Error('Invalid role id')
+    }
+
+    const { name: deviceName } = await this.getById(this.#ownDeviceId)
+
+    // since we are always getting #ownDeviceId,
+    // this should never throw (see comment on getById), but it pleases ts
+    if (!deviceName) throw new Error('Invalid deviceName')
 
+    if (!roleName) throw new Error('Invalid roleId')
```
achou11, line 76: This shouldn't be needed (and the error message is misleading). It's also causing some tests to fail when they shouldn't (since the roleName is truly optional and has a fallback)

```suggestion
```

achou11 (changes requested):
There's a check in the invite method that shoudn't be there, which is causing the tests to fail.

Non-blocking, but if they don't already exist, I think there should be test(s) to ensure that:

1. You can provide an arbitrary `roleName` to the invite and the invited person receives that `roleName` e.g. `invite({ roleName: 'friend', ... })`
2. You can provide no `roleName` to the invite and the invited person receives the fallback name that's defined for the capability of the associated `roleId` that's provided (e.g. `'Member'` or `'Coordinator'`)

src/member-api.js
tomasciccola: heh, yeah, I pushed in a rush :face_in_clouds:

src/member-api.js
```diff
@@ -51,19 +56,38 @@ export class MemberApi extends TypedEmitter {
...
    * @param {number} [opts.timeout]
    *
    * @returns {Promise<import('./generated/rpc.js').InviteResponse_Decision>}
    */
-  async invite(deviceId, { roleId, timeout }) {
-    const projectId = projectKeyToId(this.#projectKey)
+  async invite(deviceId, { roleId, roleName, roleDescription, timeout }) {
+    if (!(roleId === MEMBER_ROLE_ID || roleId === COORDINATOR_ROLE_ID)) {
+      throw new Error('Invalid role id')
+    }
```
gmaclennan, line 68: This will break some tests that invite a user as a blocked user. Could we maybe do:

```suggestion
    if (!(DEFAULT_CAPABILITIES[roleId])) {
      throw new Error('Invalid role id')
    }
```

src/member-api.js
```diff
@@ -51,19 +56,38 @@ export class MemberApi extends TypedEmitter {
...
+
+    const { name: deviceName } = await this.getById(this.#ownDeviceId)
+
+    // since we are always getting #ownDeviceId,
+    // this should never throw (see comment on getById), but it pleases ts
+    if (!deviceName) throw new Error('Invalid deviceName')
```
gmaclennan, line 74:
```suggestion
    if (!deviceName) throw new Error('Internal error trying to read own device name for this invite')
```

Trying to make the error message a bit more descriptive, since the current one is misleading because `deviceName` is not a parameter for this function.

test-e2e/manager-invite.js
```diff
@@ -123,7 +123,7 @@ skip("member can't invite", async (t) => {
   const creatorProject = await creator.getProject(createdProjectId)
 
   const responsePromise = creatorProject.$member.invite(member.deviceId, {
-    roleId: MEMBER_ROLE_ID,
+    roleId: COORDINATOR_ROLE_ID,
```
gmaclennan, line 126: Why has this test been changed? It's changed what it's testing. I know it's a skipped test for now, but the change could get lost down the line.

tests/invite-api.js
```diff
@@ -38,20 +39,24 @@ test('invite-received event has expected payload', async (t) => {
...
   r1.on('peers', (peers) => {
     t.is(peers.length, 1)
-
-    r1.invite(peers[0].deviceId, {
+    const invite = {
       projectKey,
       encryptionKeys,
       projectInfo: { name: 'Mapeo' },
-    })
+      roleName: DEFAULT_CAPABILITIES[MEMBER_ROLE_ID].name,
```
gmaclennan, line 46: This is missing a `roleId` - shouldn't this fail?

tests/invite-api.js
```diff
@@ -127,10 +136,14 @@ test('Reject invite', async (t) => {
   r1.on('peers', async (peers) => {
     t.is(peers.length, 1)
 
-    const response = await r1.invite(peers[0].deviceId, {
+    const invite = {
       projectKey,
       encryptionKeys,
-    })
+      projectInfo: { name: 'Mapeo' },
+      roleName: DEFAULT_CAPABILITIES[MEMBER_ROLE_ID].name,
```
gmaclennan, line 143: missing `roleId`

tests/invite-api.js
```diff
@@ -169,10 +182,14 @@ test('Receiving invite for project that peer already belongs to', async (t) => {
     r1.on('peers', async (peers) => {
       t.is(peers.length, 1)
 
-      const response = await r1.invite(peers[0].deviceId, {
+      const invite = {
         projectKey,
         encryptionKeys,
-      })
+        projectInfo: { name: 'Mapeo' },
+        roleName: DEFAULT_CAPABILITIES[MEMBER_ROLE_ID].name,
```
gmaclennan, line 189: missing `roleId`

tests/invite-api.js
```diff
@@ -212,11 +229,15 @@ test('Receiving invite for project that peer already belongs to', async (t) => {
...
       r1.on('peers', async (peers) => {
         t.is(peers.length, 1)
-
-        const response = await r1.invite(peers[0].deviceId, {
+        const invite = {
           projectKey,
           encryptionKeys,
-        })
+          projectInfo: { name: 'Mapeo' },
+          roleName: DEFAULT_CAPABILITIES[MEMBER_ROLE_ID].name,
```
gmaclennan, line 236: missing `roleId`

tests/invite-api.js
```diff
@@ -257,17 +278,18 @@ test('Receiving invite for project that peer already belongs to', async (t) => {
     })
 
     r1.on('peers', async (peers) => {
-      const response1 = await r1.invite(peers[0].deviceId, {
+      const invite = {
         projectKey,
         encryptionKeys,
-      })
+        projectInfo: { name: 'Mapeo' },
+        roleName: DEFAULT_CAPABILITIES[MEMBER_ROLE_ID].name,
```
gmaclennan, line 285: missing `roleId`

tests/invite-api.js
```diff
@@ -322,10 +344,14 @@ test('invitor disconnecting results in accept throwing', async (t) => {
   r1.on('peers', async (peers) => {
     if (peers.length !== 1 || peers[0].status === 'disconnected') return
     await t.exception(() => {
-      return r1.invite(peers[0].deviceId, {
+      const invite = {
         projectKey,
         encryptionKeys,
-      })
+        projectInfo: { name: 'Mapeo' },
+        roleName: DEFAULT_CAPABILITIES[MEMBER_ROLE_ID].name,
```
gmaclennan, line 351: missing `roleId`

gmaclennan (changes requested):
A couple of comments, and also the unit tests seem to be missing the `roleId` from the invite - they should be failing because of that no?
I think it's important to add the tests that @achou11 recommends before merging.

test-e2e/manager-invite.js
tomasciccola: Yeah, thanks for catching that, its a typo

src/member-api.js
gmaclennan: I think that's probably correct, however we didn't restrict this previously and in the e2e tests for sync authentication I invite as a blocked user to check that the blocked user cannot sync. Once we have the api for changing someone's role then we can update the tests, but maybe make this restriction looser for now to avoid lots of work on the sync tests and role api?

tests/invite-api.js
tomasciccola: same as comment avobe. The `invite` here refers to the protobuf, so `roleId` is not needed

tests/invite-api.js
gmaclennan: Ah yes thanks you're right. Was reading only the diff without the context.

achou11:
> I've added a couple of tests on `test-e2e/member.js` to test a custom `roleName` and an empty `roleName` (which should be derived from `roleId`). I don't know if those are enough? Also, I had to change the signature of `test-e2e/utils.invite` function to return the invite response. It looks kinda ugly (it returns a `Promise<InviteResponse | undefined>` and there may be a better way.

think it would be easier to avoid updating the `invite()` utility and instead attaching a listener to the `.invite` namespace for the peer receiving the invite e.g.

```ts
  invitee.invite.on('invite-received', ({ roleName }) => {
    t.is(roleName, 'friend', 'roleName should be equal')
  })

  await invite({
    invitor,
    projectId,
    invitees: [invitee],
    roleName: 'friend',
    reject: true,
  })
```

test-e2e/members.js
```diff
@@ -155,3 +155,50 @@ test('getting invited member after invite accepted', async (t) => {
...
+
+  const inviteResponse = await invite({
+    invitor,
+    projectId,
+    invitees: [invitee],
+    roleName: 'friend',
+    reject: true,
+  })
+
+  t.is(inviteResponse?.roleName, 'friend', 'roleName should be equal')
```
achou11, line 176: would avoid updating the `invite()` utility and do this instead:

```suggestion
  invitee.invite.on('invite-received', ({ roleName }) => {
    t.is(roleName, 'friend', 'roleName should be equal')
  })
  
  await invite({
    invitor,
    projectId,
    invitees: [invitee],
    roleName: 'friend',
    reject: true,
  })
```

test-e2e/members.js
```diff
@@ -155,3 +155,50 @@ test('getting invited member after invite accepted', async (t) => {
...
+    projectId,
+    invitees: [invitee],
+    reject: true,
+  })
+
+  t.is(
+    inviteResponse?.roleName,
+    DEFAULT_CAPABILITIES[MEMBER_ROLE_ID].name,
+    '`roleName` should use the fallback by deriving `roleId`'
+  )
```
achou11, line 201: similar suggestion about intercepting the received invite payload

```suggestion
  invitee.invite.on('invite-received', ({ roleName }) => {
    t.is(
      roleName,
      DEFAULT_CAPABILITIES[MEMBER_ROLE_ID].name,
      '`roleName` should use the fallback by deriving `roleId`'
    )
  })

  await invite({
    invitor,
    projectId,
    invitees: [invitee],
    reject: true,
  })
```

test-e2e/members.js
```diff
@@ -155,3 +155,50 @@ test('getting invited member after invite accepted', async (t) => {
   // TODO: Test that device info of invited member can be read from invitor after syncing
   await disconnectPeers(managers)
 })
+
+test('invite with abritrary name', async (t) => {
```
achou11, line 159:
```suggestion
test('invite uses custom role name when provided', async (t) => {
```

test-e2e/members.js
```diff
@@ -155,3 +155,50 @@ test('getting invited member after invite accepted', async (t) => {
...
+    roleName: 'friend',
+    reject: true,
+  })
+
+  t.is(inviteResponse?.roleName, 'friend', 'roleName should be equal')
+
+  await disconnectPeers(managers)
+})
+
+test('invite without roleName', async (t) => {
```
achou11, line 181:
```suggestion
test('invite uses default role name when not provided', async (t) => {
```

test-e2e/utils.js
```diff
@@ -69,6 +69,16 @@ export function connectPeers(managers, { discovery = true } = {}) {
...
+/** @typedef {{
+ * peerId: string,
+ * projectId: string,
+ * projectName: string,
+ * roleName: string,
+ * roleDescription?: string
+ * invitorName: string
+ * }}InviteResponse
+ */
+
```
achou11, line 81: Can remove since not necessary to update the util in general

```suggestion
```

test-e2e/utils.js
```diff
@@ -77,27 +87,35 @@ export function connectPeers(managers, { discovery = true } = {}) {
  *   projectId: string,
  *   invitees: MapeoManager[],
  *   roleId?: import('../src/capabilities.js').RoleId,
+ *   roleName?: string
  *   reject?: boolean
  * }} opts
+ * @returns {Promise<InviteResponse | undefined>}
```
achou11, line 93: remove

```suggestion
```

test-e2e/utils.js
```diff
@@ -77,27 +87,35 @@ export function connectPeers(managers, { discovery = true } = {}) {
...
   roleId = MEMBER_ROLE_ID,
+  roleName,
   reject = false,
 }) {
   const invitorProject = await invitor.getProject(projectId)
   const promises = []
 
+  /** @type {InviteResponse | undefined} */
+  let inviteResponse
+
```
achou11, line 108: remove

```suggestion
```

test-e2e/utils.js
```diff
@@ -77,27 +87,35 @@ export function connectPeers(managers, { discovery = true } = {}) {
...
   for (const invitee of invitees) {
     promises.push(
       invitorProject.$member.invite(invitee.deviceId, {
         roleId,
+        roleName,
       })
     )
     promises.push(
       once(invitee.invite, 'invite-received').then(([invite]) => {
+        inviteResponse = invite
```
achou11, line 118: remove

```suggestion
```

test-e2e/utils.js
```diff
@@ -106,6 +124,7 @@ export async function invite({
   }
 
   await Promise.allSettled(promises)
+  return inviteResponse
```
achou11, line 127: remove

```suggestion
```

achou11 (changes requested):
just pointing to specific code to revert or update based on my other comment about the recently added tests. think updating the utility would make a little more sense if you returned a mapped result e.g `Record<MapeoManager, InvitePayload>`, but I don't think it's necessary to introduce that here in general.

other than that, seems ready for merging!

achou11 (approved):

PR #394, opened 2023-11-29
chore: fix param type for Capabilities.assignRole()

achou11:
Not sure how TS was still able to infer the roleId param type correctly prior to this with the incorrect declaration, but not worth musing over 😄


gmaclennan (approved):
Sorry reviewed a while back but somehow forgot to approve. TS does seem to somehow read this correctly without `typeof`, which it shouldn't, but it's inconsistent. Thanks for the fix.

Issue #268, opened 2023-09-14
Add non-auth cores via coreOwnership and capability records


gmaclennan:
## Description

#264 implements sharing of all hypercore keys over extension messages. This introduces a small security risk: a bad actor who has a project key could make other clients add lots of cores that do not actually belong to the project and inject data.

Ideally we change this before the MVP because it avoids having to maintain backwards compatibility.

One solution to this is to verify whether cores belong to a device that is part of the project before indexing them, but this still means that disk space could be taken with cores that are not part of the project.

A more robust solution (but we should maybe also do the solution above) is to only add `auth` namespace cores via extension messages (as we were doing before), and to add other cores once we know that they are part of the project.

"being part of the project" means in this case that _any_ capability record refers for a given device, because we still want to include cores that are from a device that was previously part of the project and is now "blocked" (if we want to exclude that data we should do so at index time, or avoid downloading their data at sync time, but we should still keep the cores in the project).

We will also need to check the `coreOwnership` records to get the actual core keys for a device.

Unfortunately our indexing is unordered, so we might get core ownership records before capability records, or vice-versa.

One easy way to solve this problem is to listen for the `index-done` events, and for each event read all the capability records, look up the corresponding coreOwnership records, and add any missing cores. However this could add a lot of overhead - `index-done` happens a lot, e.g. whenever a user adds or edits data, and on every sync, and in many cases there will be no cores to add.

A more efficient way would be to add an event whenever a new `coreOwnership` or `capability` record is indexed. We could then listen on both events:

1. `coreOwnership` record added -> look up capability record for device -> if it exists then add cores listed in `coreOwnership` record, otherwise ignore
2. `capability` record added -> look up core ownership for device -> if it exists add cores, otherwise ignore.

## Tasks

- [ ] Add a `onValue()` method to `DataType` - I think this is better than just emitting a `value` event for every record, because that would add a lot of unnecessary overhead emitting thousands of events.
- [ ] Add events to `CoreOwnership` and `Capabilities` classes (implemented internally on the `dataType.onValue()`)
- [ ] In core manager, listen for new `coreOwnership` and `capability` records, and add core keys accordingly
- [ ] Remove the sharing of non-`auth` cores from the project extension message

EvanHahn:
@gmaclennan I "sketched" out a fix for this in #737. Could you give it a look? If my draft looks good, I'll fix it up properly.

EvanHahn:
The MVP parts of this work are done.

PR #396, opened 2023-11-30
feat: dataType.on('updated-docs') when docs update

gmaclennan:
Emits an array of updated docs for the data type whenever the docs
update (rather than emitting one doc at a time, it waits for indexing to
complete, then emits all the docs that have been updated. If there are
several updates to a particular docId, then only the resolved `head`
version is returned).

Attaching this listener to a data type that has many values / is
frequently updated will have a performance impact, because large amounts
of data could be read from the database and there is no limit, e.g. if
this was attached to the `observation` dataType and a new device synced
with a project with 10,000 observations, an attached listener would be
called with 10,000 observations.

The main reason for this feature is for internal use for capabilities,
listening for changes and modifying permissions accordingly, initially in order to implement #268, which requires adding cores as device capabilities are updated.

### DIFF
--- a/src/datastore/index.js
+++ b/src/datastore/index.js
@@ -3,10 +3,8 @@ import { encode, decode, getVersionId, parseVersionId } from '@mapeo/schema'
 import MultiCoreIndexer from 'multi-core-indexer'
 import pDefer from 'p-defer'
 import { discoveryKey } from 'hypercore-crypto'
+import { createMap } from '../utils.js'
 
-/**
- * @typedef {import('multi-core-indexer').IndexEvents} IndexEvents
- */
 /**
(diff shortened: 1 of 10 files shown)


gmaclennan:
> lgtm. seems like you ran into some potentially gnarly TS issues - think you made the right choice to ignore them 😄

Yeah, I wasted quite a lot of time trying to make TS happy, then just had a realization "why am I bothering". In the internals of functions like these dealing with complex data types, I don't think jumping through hoops to keep TS happy really gains us anything, especially we're hitting limitations of TS. The issue I keep getting frustrated by is trying to narrow a type with `arrayOfValidTypes.includes(maybeValidType)`, but dependent types is often an issue (TS does not support them).

PR #405, opened 2023-12-05
Feat: mapeoProject.importConfig

### LINKED ISSUE #395: Tracking issue: Load default presets
## Description

When a project is created the user should get the default presets with its associated fields and icons. What we end up after talking with @gmaclennan is the following:
1. Bundle the default config file inside the app (probably on this repo). This can change in terms of user need. i.e. the config could be distributed as a regular file that gets loaded through a dialog in the frontent.
2. When creating a new project, we first delete all the presets in the db (to avoid db bloat) and then load and write to the db the default fields, icons and lastly the presets.
3. After inviting a new peer to the project, they would get all the default presets from the project creator

Additionally we discussed replacing the format from a tarball to a zip (since it can be open in every OS without a command line) and the extension from `.mapeosettings` to `.mapeoconfig`.
 I've already created the issue on the [mapeo-settings-builder repo](https://github.com/digidem/mapeo-settings-builder/issues/50) to take that into account for future configs

- [x] #401 
- [x] #402

### LINKED ISSUE #401: mapeoProject.importConfig
## Description
Add an 'importConfig({path})` method to `MapeoProject`

tomasciccola:
This should close #401

- [x] create implementation of `importConfig`
- [x] create zip file with default configuration (grabbed from [here](https://github.com/digidem/presets-MapeoyMonitoreo)
- [x] add tests


src/mapeo-project.js
```diff
@@ -501,6 +504,63 @@ export class MapeoProject {
...
+      for await (const entry of zip) {
+        if (entry.filename === 'presets.json') {
+          /* eslint-disable no-unused-vars */
+          const { fields, presets } = await readPresets(
+            await zip.openReadStream(entry)
+          )
+          for (let [fieldName, field] of Object.entries(fields)) {
+            const fieldDoc = {
+              // shouldn't schemaName be derived when calling .create?
+              schemaName: 'field',
```
tomasciccola, line 531: Here I'm adding `schemaName` manually, but it should be redundant. In `dataType[kCreateWithDocId]` we can get `schemaName` from `this.#schemaName` and add it to the `value` directly

src/mapeo-project.js
```diff
@@ -501,6 +504,104 @@ export class MapeoProject {
...
+        for (let field of preset.fields || []) {
+          preset.fieldIds.push(fieldIds[field])
+        }
+        delete preset.fields
+        preset.iconId = iconIds[preset.icon]
+        delete preset.icon
+        preset.name = presetName
+        // TODO: this is ugly
+        preset.addTags = {}
+        preset.removeTags = {}
```
tomasciccola, line 594: we could add those two fields (`addTags` and `removeTags`) directly to the `presets.json` file

src/mapeo-project.js
```diff
@@ -501,6 +504,104 @@ export class MapeoProject {
...
+        if (!preset.terms) preset.terms = []
```
tomasciccola, line 595: sometimes `preset.terms` doesn't exist, hence the check. We could add it to the default config to every preset

src/mapeo-project.js
```diff
@@ -501,6 +504,104 @@ export class MapeoProject {
...
+   */
+  async importConfig({ configPath }) {
+    try {
+      await fs.stat(path.resolve(configPath))
+    } catch (e) {
+      console.log(`error loading config file ${configPath}`, e)
+    }
+    const zip = await yauzl.open(configPath)
+    // check for already present presets and delete them if exists
+    // TODO: do the same for fields and icons?
```
tomasciccola, line 519: should I also delete every field and icon? or does it makes more sense to get the `{iconName, iconId}` and `{fieldName, fieldId}` so that when loading presets we directly reference those?

tomasciccola:
There are some files on the config that I don't know what to do with:
- icons.json -> has size, placement (I think) and pixelRatio
- icons.{png,svg} -> I don't think we need this
- metadata.json -> I don't think we need this
- translations.json -> spanish names for fields
- style.css
- VERSION -> maybe we need it to have a runtime check (fail if version is older than new config format)

src/mapeo-project.js
tomasciccola: For now I implemented the first case (basically delete every preset, field and doc). Matching icons of fields against names doesn't seem to be super reliable...

tomasciccola:
All the deletion of fields, icons and presets prior to importing the config is commented until #417 and #418 gets merged

src/mapeo-project.js
```diff
@@ -501,6 +504,104 @@ export class MapeoProject {
...
+        delete preset.fields
+        preset.iconId = iconIds[preset.icon]
+        delete preset.icon
+        preset.name = presetName
+        // TODO: this is ugly
+        preset.addTags = {}
+        preset.removeTags = {}
+        if (!preset.terms) preset.terms = []
+        if (preset.sort) delete preset.sort
+        if (preset.matchScore) delete preset.matchScore
```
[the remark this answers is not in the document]
tomasciccola: @gmaclennan

src/mapeo-project.js
gmaclennan: The advantage of keeping it is that ensures stricter typing - another schema could have the same types for other properties, which could leave an implementer calling with wrong object. It's a way of making explicit which type we are using. I'm open to arguments for implying it and removing it from the `create` type though.

src/mapeo-project.js
gmaclennan: Let's keep it simple to use legacy preset data with CoMapeo, and just add them on import.

src/mapeo-project.js
gmaclennan: I think ok to add it.

src/mapeo-project.js
gmaclennan: Yes I would delete every field. For icons it would be nice to re-use them somehow (e.g. checking hashes) but for now let's just delete them all.

gmaclennan (changes requested):
Good work on this. There were a few bugs and I got a bit carried away reviewing and fixing, and ended up with with a bit of a restructure which is here: https://github.com/digidem/mapeo-core-next/commit/17e34e00046a66b4d4c286f0bf13268dc391ea41

I haven't tested this, but hopefully can serve as guidance. It creates a `readConfig` abstraction, so that we don't end up with a huge method in `project.importConfig` - we can keep the file parsing separate to actually adding the presets to the database.

There were a few bugs that I addressed:

- (not exactly a bug): the code was reading all icons into memory (because they were all added to the `icons` object), which would be nice to avoid because this could take a lot of memory.
- `field.key` is not the same as the `id` that is referenced in `preset.fields`. `preset.fields` refers to `Object.keys(presetsFile.fields)`.
- It's important to validate all the data we import, because it could actually be anything. I've added validation functions, and I avoid importing extra properties that we don't know about (by reading from the schema, rather than just spreading the object from the config file)

I've tried to be really strict about validating everything, but I don't throw on everything - I thought it was better to try to import what we can.

**Things that throw**

- No `presets` prop
- No `fields` prop
- A preset names an icon but it doesn't exist
- A preset names a field but it doesn't exist
- Not being able to read the config file at all

**Things that do not throw**

- Invalid preset, field or icon (although it may cause one of the errors above if the field or icon is referenced)


I'm not sure what to do with these errors / warnings and they aren't returned to the user right now. I think it's a question for the design team about how to present a warning like "We could import the presets but there were errors". Maybe they will decide to throw on any error.

I'll answer your other questions in a comment with inline replies

gmaclennan:
> There are some files on the config that I don't know what to do with:

Good point, I had forgotten about these.

> * icons.json -> has size, placement (I think) and pixelRatio
> * icons.{png,svg} -> I don't think we need this

We are not yet using these. These are the sprite sheets that Mapbox uses for rendering icons on the map (the json references the icon location within the sprite sheet `icons.png`). The goal was to render the map with icons for observations, but we can ignore this data for now until we add that feature (and maybe generate the sprite sheet at runtime if we need it)

> * metadata.json -> I don't think we need this

Yes I think it's ok to ignore for now. A question for the design team if we want this to update the project name and/or description, or any other project settings. Could be both useful and dangerous (e.g. could be confusing to see project settings changed when importing a config file)

> * translations.json -> spanish names for fields

Hmmm yes I had not thought about this. Can you maybe have a think about how we might modify the schema to include this in the field / preset records we store in the database? I don't think it makes sense to store it as a separate data type?

> * style.css

This was used in ID Editor (Mapeo territory view) to style certain features on the map in a particular way (we only used it in Waorani mapping as far as I know). I think it's ok to skip this for now.

> * VERSION -> maybe we need it to have a runtime check (fail if version is older than new config format)

Yes, I think we should throw an error if either an older or a newer version.

tomasciccola:
> Good work on this. There were a few bugs and I got a bit carried away reviewing and fixing, and ended up with with a bit of a restructure which is here: [17e34e0](https://github.com/digidem/mapeo-core-next/commit/17e34e00046a66b4d4c286f0bf13268dc391ea41)
Thanks for this! I'll take a look at it. Yeah, I knew I was doing an inefficient way of reading icons (mostly because I needed to group variants/aggregate them by icon name) but I didn't want  to mess with generators; I was basically going with the most straight away implementation for starters...

src/config-import.js
```diff
@@ -159,6 +159,8 @@ export async function readConfig(configPath) {
               ? preset.icon
               : undefined,
           value: presetValue,
+          //@ts-ignore TODO: remove this
```
tomasciccola, line 162: Ideally I would remove this but I haven't found a way to please ts yet

src/mapeo-project.js
```diff
@@ -554,34 +554,24 @@ export class MapeoProject extends TypedEmitter {
...
-
-    //     // delete preset
-    //     const { deleted: deletedPreset } = await this.preset.delete(
-    //       presetVersionId,
-    //       preset
-    //     )
-    //     if (!deletedPreset) throw new Error('error deleting preset from db')
-    //   }
-    // }
+    // check for already present fields and presets and delete them if exist
```
tomasciccola, line 557: does it make sense to do this here instead of moving it outside to a function in `src/config-imprort.js`?  I left it here since we are using a couple of methods of the class and it felt weird passing them to the function as params...

src/mapeo-project.js
```diff
@@ -627,10 +620,20 @@ export class MapeoProject extends TypedEmitter {
       this.preset.create(preset)
     )
     const createdPresets = await Promise.all(presetPromises)
-    const presetIds = createdPresets.map(({ docId }) => docId)
+    const sortedPresetIds = createdPresets
```
tomasciccola, line 623: I think it would be cleaner to return the presets already sorted but since *presets is a generator that returns a preset by one I haven't found a way to sort it there...
@gmaclennan wdyt?

src/config-import.js
```diff
@@ -0,0 +1,220 @@
...
+          value: fieldValue,
+        }
+      }
+    },
+
+    /**
+     * @returns {Iterable<{ fieldNames: string[], iconName: string | undefined, value: import('@mapeo/schema').PresetValue, sortValue: number }>}
+     */
+    *presets() {
+      const { presets } = presetsFile
```
gmaclennan, line 127:
```suggestion
      const { presets } = presetsFile
      const sortedEntries = Object.entries(presets).sort(/** sort function */)
      
      for (let [presetName, preset] of sortedEntries)
```

src/config-import.js
```diff
@@ -0,0 +1,220 @@
...
+              ? preset.fields
+              : [],
+          iconName:
+            'icon' in preset && typeof preset.icon === 'string'
+              ? preset.icon
+              : undefined,
+          value: presetValue,
+          //@ts-ignore TODO: remove this
+          sortValue: preset['sort'],
+        }
```
gmaclennan, line 164:
```suggestion
          sortValue: 'sort' in preset && typeof preset.sort === 'number' ? preset.sort : 0,
```

gmaclennan: However I would lean towards yielding in sort order, and not passing this

src/mapeo-project.js
```diff
@@ -548,6 +549,100 @@ export class MapeoProject extends TypedEmitter {
...
+        for (const fieldId of fieldIds) {
+          const field = await this.field.getByDocId(fieldId)
+          const { deleted } = await this.field.delete(field.versionId)
+          if (!deleted) throw new Error('error deleting field from db')
+        }
+        // delete preset
+        const { deleted: deletedPreset } = await this.preset.delete(
+          presetVersionId
+        )
+        if (!deletedPreset) throw new Error('error deleting preset from db')
```
gmaclennan, line 572: If there is an error deleting the preset, then `preset.delete()` will throw. `deletedPreset` being false here would indicate that there is a bug in the code, so this is code to write in a test, not runtime code.

src/mapeo-project.js
```diff
@@ -548,6 +549,100 @@ export class MapeoProject extends TypedEmitter {
...
     return this.#iconApi
   }
+
+  /** @param {Object} opts
+   *  @param {string} opts.configPath
+   */
+  async importConfig({ configPath }) {
+    // check for already present fields and presets and delete them if exist
+    const presets = await this.preset.getMany()
+    if (presets.length !== 0) {
```
gmaclennan, line 559: You don't need this check, just do `for (const preset of presets)`

src/mapeo-project.js
```diff
@@ -548,6 +549,100 @@ export class MapeoProject extends TypedEmitter {
...
+      for (const preset of presets) {
+        const { fieldIds, versionId: presetVersionId } = preset
+        // delete fields
+        for (const fieldId of fieldIds) {
+          const field = await this.field.getByDocId(fieldId)
+          const { deleted } = await this.field.delete(field.versionId)
+          if (!deleted) throw new Error('error deleting field from db')
```
gmaclennan, line 566: no need for this - see below

src/mapeo-project.js
```diff
@@ -548,6 +549,100 @@ export class MapeoProject extends TypedEmitter {
...
+   *  @param {string} opts.configPath
+   */
+  async importConfig({ configPath }) {
+    // check for already present fields and presets and delete them if exist
+    const presets = await this.preset.getMany()
+    if (presets.length !== 0) {
+      for (const preset of presets) {
+        const { fieldIds, versionId: presetVersionId } = preset
+        // delete fields
+        for (const fieldId of fieldIds) {
```
gmaclennan, line 563: Just do the same `this.field.getMany()` and delete them all, rather than this. You could end up trying to delete the same field twice, which will throw an error.

src/mapeo-project.js
```diff
@@ -548,6 +549,100 @@ export class MapeoProject extends TypedEmitter {
...
+          const field = await this.field.getByDocId(fieldId)
+          const { deleted } = await this.field.delete(field.versionId)
+          if (!deleted) throw new Error('error deleting field from db')
+        }
+        // delete preset
+        const { deleted: deletedPreset } = await this.preset.delete(
```
gmaclennan, line 569: Because of how we implement sqlite writes, this is not really an issue, but generally better to do something like this in parallel rather than serial.

src/mapeo-project.js
gmaclennan: I'm fine with it here. Could also add a `deleteAll()` method to `DataType` or we could do a quick helper:

```
async function deleteAll(dataType) {
  const deletions = []
  for (const doc of await dataType.getMany()) {
    deletions.push(dataType.delete(doc.versionId))
  }
  return Promise.all(deletions)
}
```

src/mapeo-project.js
gmaclennan: Yes I think easier / simpler to sort before we return from the preset reader. I added a comment with one way of doing it.

gmaclennan (commented):
Added some comments and suggestions

src/mapeo-project.js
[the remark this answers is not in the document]
tomasciccola: As per the talk with gregor, sort is actually expressed in the project settings, with the `defaultPresets` property, were smth like the `point` property is an array of ids (of field docs) which determines the ordering. 
The ordering happens on the import config `*presets()` generator

src/mapeo-project.js
tomasciccola: I'm wondering about this. If I do smth like:
```js
// this would actually be in parallel
const fields = await this.field.getMany()
for(const {versionId} of fields){
    await this.field.delete(versionId)
}
```
I'm actually not bothering to check if the fields that I'm deleting are referenced by a preset; I'm just deleting every field. Could there be an issue with that? (since we're also deleting every preset I would say there isn't any issue, I'm just double checking...)

tomasciccola:
> * VERSION -> maybe we need it to have a runtime check (fail if version is older than new config format)

> Yes, I think we should throw an error if either an older or a newer version.

Currently `VERSION` is 3.4.1. If we are bumping the `config-builder` version to a new major version (since changing from tar to a zip means a breaking change, right?), should we check against `VERSION` === 4.x.x ?

tomasciccola:
I'm sorry @gmaclennan I edited your big comment regarding all the other files I didn't know what to do with (my intention was to `quote reply`...)

src/mapeo-project.js
gmaclennan: A loop with an await is always serial. The only way to get parallel async is with promise.all / allSettled. It doesn't actually make much difference for this code because better SQLite is synchronous. But could change in the future, so best to do parallel as best practice.

gmaclennan: No, I can't see any issue with deleting every field.

tomasciccola:
I was wondering how to write tests for this and moreso how to write fixtures that cover different failing cases. One possibility is to have different zip files with different malformed parts and commit them; another possibility is to have a folder with all the necessary parts and have a build step that programatically creates a series of zip files with different failing parts. The second is maybe more involved but at the same time reduces space?
@gmaclennan @achou11 wdyt??

achou11:
> I was wondering how to write tests for this and moreso how to write fixtures that cover different failing cases. One possibility is to have different zip files with different malformed parts and commit them; another possibility is to have a folder with all the necessary parts and have a build step that programatically creates a series of zip files with different failing parts. The second is maybe more involved but at the same time reduces space? @gmaclennan @achou11 wdyt??

Assuming the zipfiles aren't taking up gigabytes of space, I would lean towards the first option of having various already-built zip files. having a build step for assets that won't take up non-trivial amounts of space doesn't seem necessary to me

src/mapeo-project.js
```diff
@@ -643,6 +711,16 @@ function extractEditableProjectSettings(projectDoc) {
...
 
+// TODO: maybe a better signature than a bunch of any?
+/** @param {DataType<any,any,any,any,any>} dataType */
+async function deleteAll(dataType) {
+  const deletions = []
+  for (const { versionId } of await dataType.getMany()) {
+    deletions.push(dataType.delete(versionId))
+  }
+  return Promise.all(deletions)
+}
```
EvanHahn, line 722: Is there a way to do all of this in a single database transaction? It's possible that we'll get a list from `getMany` which will be different by the time we delete them.

src/config-import.js
```diff
@@ -0,0 +1,231 @@
...
+
+/**
+ * @param {string} configPath
+ */
+export async function readConfig(configPath) {
+  /** @type {Error[]} */
+  const warnings = []
+
+  const zip = await yauzl.open(configPath)
+  if (zip.entryCount > MAX_ENTRIES) {
```
EvanHahn, line 29: 1. Do we want a maximum ZIP file size as well?
2. [yauzl notes](https://github.com/overlookmotel/yauzl-promise/blob/e52bdeda0fd1f0bf2292f13d7f89208e17607170/README.md#zipentrycountiscertain) that you can't always know the number of entries in a ZIP file upfront. It seems like the maximum is 65,535&mdash;much larger than `MAX_ENTRIES`&mdash;but I could see adding a comment warning future developers about increasing this limit. Alternatively, we could handle that case.

src/config-import.js
```diff
@@ -0,0 +1,231 @@
+import yauzl from 'yauzl-promise'
+import { validate, valueSchemas } from '@mapeo/schema'
+import { text, buffer } from 'node:stream/consumers'
+
+// Throw error if a zipfile contains more than 10,000 entries
+const MAX_ENTRIES = 10_000
+
+/**
+ * @typedef {Omit<yauzl.Entry, 'fileName'> & { filename: string }} Entry
```
EvanHahn, line 9: Nit: I was confused by this before I realized that `@types/yauzl-promise` is incorrect. Would you mind adding an explanation here?

One possible option:

```suggestion
 * @typedef {Omit<yauzl.Entry, 'fileName'> & { filename: string }} Entry
 * `@types/yauzl-promise` has the wrong name for this key. We should
 * eventually submit this patch upstream.
```

Feel free to ignore if you think it's obvious.

src/config-import.js
```diff
@@ -0,0 +1,231 @@
...
+  /** @type {Entry[]} */
+  const entries = /** @type {any} */ (await zip.readEntries(MAX_ENTRIES))
+
+  const presetsEntry = entries.find((e) => e.filename === 'presets.json')
+  if (!presetsEntry) {
+    throw new Error('Zip file does not contain presets.json')
+  }
+  const presetsFile = JSON.parse(
+    await text(await presetsEntry.openReadStream())
+  )
```
EvanHahn, line 41: Nit: consider using the [`json` consumer](https://nodejs.org/api/webstreams.html#streamconsumersjsonstream) instead.

```suggestion
const presetsFile = await json(await presetsEntry.openReadStream())
```

You'll need to import it up top.

src/config-import.js
```diff
@@ -0,0 +1,231 @@
...
+  const warnings = []
+
+  const zip = await yauzl.open(configPath)
+  if (zip.entryCount > MAX_ENTRIES) {
+    throw new Error(`Zip file contains too many entries. Max is ${MAX_ENTRIES}`)
+  }
+  /** @type {Entry[]} */
+  const entries = /** @type {any} */ (await zip.readEntries(MAX_ENTRIES))
+
+  const presetsEntry = entries.find((e) => e.filename === 'presets.json')
```
EvanHahn, line 35: I wonder if we could avoid reading all the entries into memory. Would something like this work?

```suggestion
  /** @type {undefined | Entry} */
  let presetsEntry
  for await (const entry of zip) {
    if (entry.filename === 'presents.json') {
      presetsEntry = entry
      break
    }
  }
```

See my suggestion in `icons` below.

src/config-import.js
```diff
@@ -0,0 +1,231 @@
...
+    /**
+     * @returns {AsyncIterable<IconData>}
+     */
+    async *icons() {
+      const iconEntries = entries
+        .filter((e) => e.filename.startsWith('/icons/'))
+        .sort()
+      /** @type {IconData | undefined} */
+      let icon
+      for (const entry of iconEntries) {
```
EvanHahn, line 63: If we want to avoid reading all entries into memory:

```suggestion
      /** @type {IconData | undefined} */
      let icon
      for await (const entry of zip) {
        if (!entry.filename.startsWith('/icons/')) continue
```

src/config-import.js
```diff
@@ -0,0 +1,231 @@
...
+        const buf = await buffer(await entry.openReadStream())
```
EvanHahn, line 64: Do we want a maximum icon file size?

EvanHahn (commented):
Partial review. I'll look at the rest soon!

src/config-import.js
```diff
@@ -0,0 +1,231 @@
...
+  const [_, name, size, pixelDensity, extension] = matches
+  const density = Number(pixelDensity.replace('x', ''))
+  if (!(density === 1 || density === 2 || density === 3)) {
+    throw new Error(`Error loading icon. invalid pixel density ${density}`)
+  }
+  if (!(size === 'small' || size === 'medium' || size === 'large')) {
+    throw new Error(`Error loading icon. invalid size ${size}`)
+  }
+  if (!name) {
+    throw new Error('Error loading icon. missing name')
```
tomasciccola, line 203: I think we can remove this assertion because the regex matches one or more occurrences of any alfanum char, so name being empty would be caught by the first assertion (`if(!matches)`). But maybe there's a case I'm missing?

src/config-import.js
tomasciccola: Yeah, I don't its obvious at all, so I'll add the comment

src/config-import.js
EvanHahn: I agree that it's redundant for now, but I'd prefer to keep this in case we want to enable [unchecked indexed access][0] in the future.

[0]: https://www.typescriptlang.org/tsconfig#noUncheckedIndexedAccess

src/config-import.js
```diff
@@ -0,0 +1,246 @@
...
+  }
+  return {
+    name,
+    variant: {
+      size,
+      mimeType: extension === 'png' ? 'image/png' : 'image/svg+xml',
+      pixelDensity: density,
```
EvanHahn, line 225: A few comments about this implementation:

1. I think we can simplify the `pixelDensity` capture group to avoid having to remove the `x` later, by using `(\d+)x` instead of `(\d+x)`.
1. Do we want to make the extension case-insensitive?
1. Do we want to accept SVGs too?
1. Do we want to treat JPEGs as SVGs?

```suggestion
  const parsedFilename = path.parse(filename)

  const matches = parsedFilename.name.match(/([a-zA-Z0-9-]+)-([a-z]+)@(\d)x/)
  if (!matches) {
    throw new Error(`Unexpected icon filename ${filename}`)
  }
  const [name, size, pixelDensityStr] = matches.slice(1)
  const pixelDensity = Number(pixelDensityStr)
  if (!(pixelDensity === 1 || pixelDensity === 2 || pixelDensity === 3)) {
    throw new Error(`Error loading icon. invalid pixel density ${pixelDensityStr}`)
  }
  if (!(size === 'small' || size === 'medium' || size === 'large')) {
    throw new Error(`Error loading icon. invalid size ${size}`)
  }
  if (!name) {
    throw new Error('Error loading icon. missing name')
  }

  /** @type {'image/png' | 'image/svg+xml'} */
  let mimeType
  switch (parsedFilename.ext.toLowerCase()) {
    case '.png':
      mimeType = 'image/png'
      break
    case '.svg':
      mimeType = 'image/svg+xml'
      break
    default:
      throw new Error(`Unexpected icon extension ${parsedFilename.ext}`)
  }

  return {
    name,
    variant: {
      size,
      mimeType,
      pixelDensity,
```

src/config-import.js
```diff
@@ -0,0 +1,246 @@
...
+            warnings.push(new Error(`invalid preset ${presetName}`))
+          }
+          return !isInvalidPreset
+        })
+        .map((presetName) => {
+          /** @type {any} */
+          const preset = presets[presetName]
+          if (!preset.sort) {
+            // if there's no sort field, put a big value - puts the field at the end -
+            preset.sort = 100
```
EvanHahn, line 154: Nit: could we use `Infinity` instead of `100`?

```suggestion
            preset.sort = Infinity
```

src/config-import.js
```diff
@@ -0,0 +1,246 @@
...
+     * @returns {Iterable<{ name: string, value: import('@mapeo/schema').FieldValue }>}
+     */
+    *fields() {
+      const { fields } = presetsFile
+      for (const [name, field] of Object.entries(fields)) {
+        if (typeof field !== 'object' || field === null || !('key' in field)) {
+          warnings.push(new Error(`Invalid field ${name}`))
+          continue
+        }
+        /** @type {any} */
```
EvanHahn, line 111: It'd be nice to avoid the `any` here. I think we can do this:

```suggestion
        if (!isRecord(field)) {
          warnings.push(new Error(`Invalid field ${name}`))
          continue
        }
        /** @type {Record<string, unknown>} */
```

And then define `isRecord` below:

```js
/**
 * @param {unknown} value
 * @returns {value is Record<string, unknown>}
 */
function isRecord(value) {
  return value !== null && typeof value === 'object' && !Array.isArray(value)
}
```

src/config-import.js
```diff
@@ -0,0 +1,246 @@
...
+        const fieldValue = {
+          schemaName: 'field',
+          tagKey: field.key,
+        }
+        for (const key of Object.keys(valueSchemas.field.properties)) {
+          if (key in field) {
```
EvanHahn, line 117: If `key` is an inherited property (such as `hasOwnProperty`) we'll have subtle bugs. Maybe we could do something like this:

```suggestion
          if (hasOwn(field, key)) {
```

And then define `hasOwn` below:

```js
/**
 * @param {Record<string | symbol, unknown>} obj
 * @param {string | symbol} prop
 */
function hasOwn(obj, prop) {
  return Object.prototype.hasOwnProperty.call(obj, prop)
}
```

(It'd be nice to use `Object.hasOwn` but it [isn't supported by our Node versions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn#browser_compatibility).)

src/config-import.js
tomasciccola: 1. Added that change
2. I would say yes 
3. Eventually we will support svg too, trying to match with jpeg or jpg doesn't make sense, so added your changes to allow only svg or png

src/config-import.js
tomasciccola: Wouldn't this cause an issue with JSON since `Infinity` is not supported? (I don't remember is JSONSchema `JSON.stringify`'s stuff)

src/config-import.js
```diff
@@ -0,0 +1,271 @@
...
+      }
+    },
+
+    /**
+     * @returns {Iterable<{ name: string, value: import('@mapeo/schema').FieldValue }>}
+     */
+    *fields() {
+      const { fields } = presetsFile
+      for (const [name, field] of Object.entries(fields)) {
+        if (!isRecord(field) && !hasOwn(field, 'key')) {
```
EvanHahn, line 103: Should this be an `||`?

```suggestion
        if (!isRecord(field) || !hasOwn(field, 'key')) {
```

tests/fixtures/config/toBigOfAZip.zip
EvanHahn, line 1: I've found that Git sometimes gets slow on larger files, so I found a way to shrink this to ~⅓ the size: [tooBigOfAZip.zip](https://github.com/digidem/mapeo-core-next/files/14131857/tooBigOfAZip.zip)

It's a zip with 10,001 files. (If you're curious, [here's the script I used][0].)

[0]: https://gist.github.com/EvanHahn/ebcd06e5eda5957c35e03e5ff4aadabd

src/config-import.js
```diff
@@ -0,0 +1,271 @@
...
+import path from 'node:path'
+
+// Throw error if a zipfile contains more than 10,000 entries
+const MAX_ENTRIES = 10_000
+
+/**
+ * @typedef {Omit<yauzl.Entry, 'fileName'> & { filename: string }} Entry
+ * `@types/yauzl-promise` has the wrong name for this key. We should
+ * eventually submit this patch upstream.
+ */
```
EvanHahn, line 13: It looks like you're now getting type errors because this is insufficient. I think we can fix it by updating `types/yauzl-promise.d.ts`:

```diff
diff --git a/types/yauzl-promise.d.ts b/types/yauzl-promise.d.ts
index 1faad07..acf6a23 100644
--- a/types/yauzl-promise.d.ts
+++ b/types/yauzl-promise.d.ts
@@ -3,4 +3,10 @@ declare module 'yauzl-promise' {
   export interface ZipFile extends yauzl.ZipFile {
     [Symbol.asyncIterator]()
   }
+
+  export interface Entry extends yauzl.Entry {
+    // `@types/yauzl-promise` has the wrong name for this key.
+    // We should eventually submit this patch upstream.
+    filename: string;
+  }
 }
```

We can then import the Entry type which should fix things:

```suggestion
/** @typedef {yauzl.Entry} Entry */
```

src/config-import.js
EvanHahn: It looks like you defined `hasOwn` but didn't update this line. Would you mind updating it?

```suggestion
          if (hasOwn(field, key)) {
```

src/config-import.js
```diff
@@ -0,0 +1,271 @@
...
+     * @returns {Iterable<{ name: string, value: import('@mapeo/schema').FieldValue }>}
+     */
+    *fields() {
+      const { fields } = presetsFile
+      for (const [name, field] of Object.entries(fields)) {
+        if (!isRecord(field) && !hasOwn(field, 'key')) {
+          warnings.push(new Error(`Invalid field ${name}`))
+          continue
+        }
+        /** @type {Record<string,unknown>} */
```
EvanHahn, line 107: The smallest possible nitpick:

```suggestion
        /** @type {Record<string, unknown>} */
```

src/config-import.js
```diff
@@ -0,0 +1,271 @@
...
+        const fieldValue = {
+          schemaName: 'field',
+          tagKey: field.key,
+        }
+        for (const key of Object.keys(valueSchemas.field.properties)) {
+          if (key in field) {
+            // @ts-ignore - we validate below
```
EvanHahn, line 114: I think we can remove this if you apply some of my changes above.

```suggestion
```

src/config-import.js
```diff
@@ -0,0 +1,271 @@
...
+              ? preset.fields
+              : [],
+          iconName:
+            'icon' in preset && typeof preset.icon === 'string'
+              ? preset.icon
+              : undefined,
+          value: presetValue,
+        }
+      }
+    },
```
EvanHahn, line 188: I'd love if we could remove some of the `any`s and `ts-ignore`s in this function. Here's my take on how to do that:

```suggestion
    *presets() {
      const { presets } = presetsFile

      /** @type {Record<string, unknown>[]} */
      const validPresets = []
      for (const [presetName, preset] of Object.entries(presets)) {
        if (isRecord(preset)) {
          validPresets.push(preset)
        } else {
          warnings.push(new Error(`invalid preset ${presetName}`))
        }
      }

      validPresets.sort((a, b) => {
        const aSort = typeof a === 'number' ? a : Infinity
        const bSort = typeof b === 'number' ? b : Infinity
        return bSort - aSort
      })

      for (const preset of validPresets) {
        /** @type {Record<string, unknown>} */
        const presetValue = {
          schemaName: 'preset',
          fieldIds: [],
          addTags: {},
          removeTags: {},
          terms: [],
        }
        for (const key of Object.keys(valueSchemas.preset.properties)) {
          if (hasOwn(preset, key)) {
            presetValue[key] = preset[key]
          }
        }
        if (!validate('preset', presetValue)) {
          warnings.push(new Error(`Invalid preset ${preset.name}`))
          continue
        }
        yield {
          fieldNames:
            'fields' in preset &&
            Array.isArray(preset.fields) &&
            preset.fields.every((f) => typeof f === 'string')
              ? preset.fields
              : [],
          iconName:
            'icon' in preset && typeof preset.icon === 'string'
              ? preset.icon
              : undefined,
          value: presetValue,
        }
      }
    },
```

(Pretty similar to what you had, just removing some `any`s.)

src/config-import.js
EvanHahn: I actually think we don't want `sort` on the final output at all, so maybe we could onnly define it in `.sort`? See my suggested code below for my take.

(I might be wrong—maybe we _do_ want sort! If so, `Infinity` gets stringified as `null`, which is probably not what we want.)

src/config-import.js
```diff
@@ -0,0 +1,271 @@
...
+function parseIcon(filename, buf) {
+  const parsedFilename = path.parse(filename)
+  const matches = parsedFilename.base.match(
+    /([a-zA-Z0-9-]+)-([a-zA-Z]+)@(\d+)x\.(png|svg)$/
+  )
+  if (!matches) {
+    throw new Error(`Unexpected icon filename ${filename}`)
+  }
+  /* eslint-disable no-unused-vars */
+  const [_, name, size, pixelDensityStr, extension] = matches
```
EvanHahn, line 206: `extension` is unused here. I think we should either use it (replacing `parsedFilename.ext` below) or drop it, and drop it from the regex too. I'd rather drop it here and from the regex so we can get a case-insensitive extension check below, but I don't feel strongly.

src/config-import.js
```diff
@@ -0,0 +1,271 @@
...
+ * @param {unknown} presetsFile
+ * @returns {asserts presetsFile is PresetsFile}
+ */
+function validatePresetsFile(presetsFile) {
+  if (
+    !isRecord(presetsFile) ||
+    !('presets' in presetsFile) ||
+    typeof presetsFile.presets !== 'object' ||
+    !('fields' in presetsFile) ||
+    typeof presetsFile.fields !== 'object'
```
EvanHahn, line 250: Checking for `!== 'object'` could return false positives for `null` or arrays. Maybe we could use `isRecord` instead?

```suggestion
    !isRecord(presetsFile) ||
    !isRecord(presetsFile.presets) ||
    !isRecord(presetsFile.fields)
```

src/config-import.js
```diff
@@ -0,0 +1,271 @@
...
+function isRecord(value) {
+  return value !== null && typeof value === 'object' && !Array.isArray(value)
+}
+
+/**
+ * @param {Record<string | symbol, unknown>} obj
+ * @param {string | symbol} prop
+ */
+function hasOwn(obj, prop) {
+  if (!isRecord(obj)) return
```
EvanHahn, line 269: Nit: why is this check necessary?

```suggestion
```

tests/config-import.js
```diff
@@ -0,0 +1,173 @@
+import { test } from 'brittle'
+import { readConfig } from '../src/config-import.js'
+
+test('config import - loading', async (t) => {
+  t.exception(async () => await readConfig(''), /ENOENT/, 'file not found')
```
EvanHahn, line 5: [Brittle's docs][0] say that "if a `t.exception` is async, then you're supposed to await it." I think we need to add `await` before each of these calls.

[0]: https://www.npmjs.com/package/brittle#texceptionpromisefunctionasync-function-error-message

src/mapeo-project.js
```diff
@@ -631,6 +632,73 @@ export class MapeoProject extends TypedEmitter {
...
+  /** @param {Object} opts
+   *  @param {string} opts.configPath
+   */
+  async importConfig({ configPath }) {
+    // check for already present fields and presets and delete them if exist
+    await deleteAll(this.preset)
+    await deleteAll(this.field)
+
+    const config = await readConfig(configPath)
+    /** @type {Map<string, string>} */
```
EvanHahn, line 645: Love that these are using maps—I think `Map` is very underrated!

src/mapeo-project.js
```diff
@@ -631,6 +632,73 @@ export class MapeoProject extends TypedEmitter {
...
+      defaultPresets: {
+        point: presetIds,
+        line: [],
+        area: [],
+        vertex: [],
+        relation: [],
+      },
+    })
+
+    await config.close()
```
EvanHahn, line 700: What if we did this a little earlier in the function so we could close the file sooner (and avoid the case where some operation, like `$setProjectSettings`, fails and we fail to close the file)?

EvanHahn (changes requested):
All done with my review. Sorry for all the back-and-forth!

Let me know if you have questions on any of this. Happy to pair next week if you want.

types/yauzl-promise.d.ts
```diff
@@ -0,0 +1,6 @@
+import * as yauzl from 'yauzl-promise'
+declare module 'yauzl-promise' {
+  export interface ZipFile extends yauzl.ZipFile {
+    [Symbol.asyncIterator]()
+  }
```
EvanHahn, line 5: [I just patched this upstream](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/68387), so we can update `@types/yauzl-promise` and delete this.

src/mapeo-project.js
tomasciccola: I don't think there's a way now of doing a bunch of deletions in a single transaction. I think this would imply changes to `src/datatype.js` and also probably the sqlite-indexer? Probably @gmaclennan can cheap in

src/config-import.js
tomasciccola: I think it makes sense to guard against icon file size, what do you think a good limit should be? (I'm thinking less than 10MBs)

src/config-import.js
```diff
@@ -0,0 +1,271 @@
...
+    }
+  }
+  if (!presetsEntry) {
+    throw new Error('Zip file does not contain presets.json')
+  }
+  const presetsFile = await json(await presetsEntry.openReadStream())
+  validatePresetsFile(presetsFile)
+
+  return {
+    get errors() {
```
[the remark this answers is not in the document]
tomasciccola: Yeah, renaming to warnings makes sense, since this should throw. 
The idea is to be able to grab warnings from the front-end to present them to the user in a friendly manner, but haven't sync with the frontend team in order to see how we would do this - maybe it can be addressed in a future PR? ? -
I think for now returning the warnings from importConfig makes sense

src/config-import.js
tomasciccola: I need to double check, but if obj is null, then `hasOwnPropery.call` will fail (and we call hasOwn with stuff that can be null)

src/config-import.js
tomasciccola: Yeah, you're right, we don't need sort on the output, since we return the presets already sorted. The use for this is when calling `this.$setProjectSettings` and we pass a list of presetIds, the order determines the sorting (see [here](https://github.com/digidem/mapeo-core-next/blob/8e31ab08dccbaeb0a78ff32a0e0867b321fd658c/src/mapeo-project.js#L691))

src/config-import.js
tomasciccola: oops, yeah I missed this and I incidentally this fixed a couple of type errors I had with `field`

src/config-import.js
tomasciccola: (I don't see any suggested code below :|), but I've committed some changes:
1. basically define sort as 
```js
const sort = preset.sort || Infinity
const nextSort = nextPreset.sort || Infinity
```
on the sort fn
2. `delete preset.sort` on a `.map` afterwards (I don't think this is strictly necessary, just a nice-to-do)

src/mapeo-project.js
tomasciccola: Regarding where to put close, I'm on the fence between putting it after reading the presets file [here](https://github.com/digidem/mapeo-core-next/blob/8e31ab08dccbaeb0a78ff32a0e0867b321fd658c/src/mapeo-project.js#L684) or after creating the presets records [here](https://github.com/digidem/mapeo-core-next/blob/8e31ab08dccbaeb0a78ff32a0e0867b321fd658c/src/mapeo-project.js#L687). 
For now I'm putting it after reading the presets file since that's when we're sure we won't be needing that handle anymore...

EvanHahn:
I'll plan to re-review this today or tomorrow.

src/config-import.js
EvanHahn: Fine to do later, but it doesn't look like the extension is case-insensitive right now.

src/config-import.js
tomasciccola: Yeah, probably before continuing we should push to the warnings a message detailing the icon culprit

src/config-import.js
```diff
@@ -0,0 +1,277 @@
...
+        const presetValue = {
+          schemaName: 'preset',
+          fieldIds: [],
+          addTags: {},
+          removeTags: {},
+          terms: [],
+        }
+        for (const key of Object.keys(valueSchemas.preset.properties)) {
+          if (key in preset) {
+            // @ts-ignore - we validate below
```
EvanHahn, line 176: We can get some type safety (by removing `any`s and `ts-ignore`s) with some minor changes:

```suggestion
      /** @type {Array<Record<string, unknown>>} */
      const sortedPresets = []
      for (const [presetName, preset] of Object.entries(presets)) {
        if (isRecord(preset)) {
          sortedPresets.push(preset)
        } else {
          warnings.push(new Error(`invalid preset ${presetName}`))
        }
      }
      sortedPresets.sort((preset, nextPreset) => {
        const sort = typeof preset.sort === 'number' ? preset.sort : Infinity
        const nextSort =
          typeof nextPreset.sort === 'number' ? nextPreset.sort : Infinity
        return sort - nextSort
      })

      // 5. for each preset get the corresponding fieldId and iconId, add them to the db
      for (let preset of sortedPresets) {
        /** @type {Record<string, unknown>} */
        const presetValue = {
          schemaName: 'preset',
          fieldIds: [],
          addTags: {},
          removeTags: {},
          terms: [],
        }
        for (const key of Object.keys(valueSchemas.preset.properties)) {
          if (hasOwn(preset, key)) {
```

(This suggestion looks larger than it really because it doesn't show the proper diff...)

EvanHahn (changes requested):
LGTM other than a few small nitpicks!

tomasciccola:
before merging, I've created an issue [here](https://github.com/digidem/mapeo-core-next/issues/470) to see if it makes sense implementing `dataType.deleteAll()`

PR #410, opened 2023-12-07
feat: add project.leave()

### LINKED ISSUE #241: How do we leave a project?
## Description

As part of the project api, we want to add a `project.$leave()` method that removes yourself from the project. However, there are some question about what happens internally when doing this:

- can/do we delete project data? maybe some stuff would have to remain in order to inform others that we've left
- do we just write the "blocked" capability for ourself?

achou11:
Closes #241 

TODO:

- Delete indexed data from the project-specific database (will be follow-up)
- Update sync mode? (may need a separate PR to introduce said sync mode) (will be follow-up)


gmaclennan (changes requested):
This approach is looking correct to me. However we don't currently have a way to assign a role, so it's not exactly working as expected - see #413. Initial fix in #414 but that just causes project.$leave() to throw right now.

This also raises questions for me about the implementation of Capabilities. The idea is that DEFAULT_CAPABILITIES will eventually move into the database, and can be edited by users. I think there are two separate types of roles:

1. "invitable roles" - e.g. roles that someone with the right capability can assign to a new device in a project. These should also be the only roles that can be assigned with `$member.assignRole()`.
2. "special roles" used to implement internal details of the capability system - "creator", "blocked device" and "left of own choice". I think those assignments should only be used internally, e.g. we should implement a `$member.block()` method, and we have this `project.$leave()` method.

I think we should implement a `capabilities.getRoles()` function or similar which currently returns the corodinator and member roles, but in the future will read from the database. We can then use this as an assertion in assignRole().

I think maybe we give the "leave" role special treatment - any role can assign the leave role to themselves, apart from a device with the blocked role - hence blocked should be treated as a "special role" with slightly different rules.

I think this PR is blocked for now by #413 which we should resolve first.

src/mapeo-project.js
```diff
@@ -548,6 +564,67 @@ export class MapeoProject extends TypedEmitter {
...
+      'data',
+      'blobs',
+      'blobIndex',
+    ])) {
+      const deletionPromises = []
+      const coreRecords = this.#coreManager.getCores(namespace)
+
+      for (const { core } of coreRecords) {
+        deletionPromises.push(core.purge())
+      }
```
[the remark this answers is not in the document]
achou11: Addressing in https://github.com/digidem/mapeo-core-next/pull/419

src/mapeo-project.js
```diff
@@ -548,6 +564,67 @@ export class MapeoProject extends TypedEmitter {
...
+
+    // clear data from cores
+    // TODO: only clear synced data
+    const namespacePromises = []
+    for (const namespace of /** @type {import('./core-manager/core-index.js').Namespace[]} */ ([
+      'config',
+      'data',
+      'blobs',
+      'blobIndex',
+    ])) {
```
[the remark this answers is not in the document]
achou11: addressed via https://github.com/digidem/mapeo-core-next/pull/410/commits/6b6ae3faed005fdd4e2530b173b37c98d97ad1a8

src/mapeo-project.js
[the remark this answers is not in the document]
achou11: addressed via https://github.com/digidem/mapeo-core-next/pull/410/commits/ecc5e7183739de94bb098f4d005733745b94838e

achou11:
@gmaclennan do you mind taking another look at the changes here and providing input on what else is needed? Not sure what from https://github.com/digidem/mapeo-core-next/pull/410#pullrequestreview-1776816519 should be done separately, and not sure if the TODOs I have in the changes diff reflect what's needed

src/capabilities.js
```diff
@@ -70,7 +76,8 @@ export const NO_ROLE_CAPABILITIES = {
       { readOwn: true, writeOwn: true, readOthers: false, writeOthers: false },
     ]
   }),
-  roleAssignment: [],
+  // TODO: Does this make sense?
+  roleAssignment: [LEFT_ROLE_ID],
```
achou11, line 80: The more I'm looking at this, the more it seems that this change isn't necessary (i.e. NO_ROLE being able to assign LEFT), but would like confirmation before undoing

src/capabilities.js
```diff
@@ -138,6 +150,28 @@ export const DEFAULT_CAPABILITIES = {
...
+    }),
+    roleAssignment: [],
+    sync: {
+      auth: 'allowed',
+      config: 'blocked',
+      data: 'blocked',
+      blobIndex: 'blocked',
+      blob: 'blocked',
+    },
+  },
```
achou11, line 174: is this an accurate representation for this role?

src/mapeo-project.js
```diff
@@ -548,6 +565,59 @@ export class MapeoProject extends TypedEmitter {
...
+          nonce
+        ),
+      })
+      .where(eq(projectKeysTable.projectId, this.#projectId))
+      .run()
+
+    this.#sharedDb
+      .delete(projectSettingsTable)
+      .where(eq(projectSettingsTable.docId, this.#projectId))
+      .run()
```
achou11, line 596: I _think_ this makes sense but not entirely sure. My assumption is that once you leave a project, you shouldn't be able to access this kind of information, but maybe that's too strict in this case?

src/mapeo-project.js
```diff
@@ -548,6 +565,59 @@ export class MapeoProject extends TypedEmitter {
...
+        'blobIndex',
+      ])
+
+    await Promise.all(
+      namespacesWithoutAuth.map((namespace) =>
+        this.#coreManager.deleteData(namespace, { deleteOwn: true })
+      )
+    )
+
+    // 3. Clear data from project database
```
achou11, line 614: Haven't implemented yet because I have questions about how this should be done. Is it just a naive db.delete(_table_)? (where `db = new Database(this.#sqlite)`)

Wondering if the deletion of data in sqlite should be handled by something like `CoreManager.deleteData()`? A little bit tangled with the different boundaries and what should be responsible for this 😅

src/mapeo-project.js
```diff
@@ -603,3 +673,43 @@ function mapAndValidateDeviceInfo(doc, { coreDiscoveryKey }) {
...
+  const deviceIdToCapsList = Object.entries(deviceIdToCaps)
+
+  if (deviceId in deviceIdToCaps) {
+    if (deviceIdToCapsList.length === 1) {
+      throw new Error('Cannot leave a project as the only device')
+    }
+
+    if (
+      deviceIdToCaps[deviceId].name ===
+      DEFAULT_CAPABILITIES[BLOCKED_ROLE_ID].name
```
achou11, line 693: Is this a better way to do an equality check than using the `name`?

src/capabilities.js
```diff
@@ -254,6 +288,10 @@ export class Capabilities {
    * @param {keyof typeof DEFAULT_CAPABILITIES} roleId
    */
   async assignRole(deviceId, roleId) {
+    if (deviceId !== this.#ownDeviceId && roleId === LEFT_ROLE_ID) {
```
gmaclennan, line 291: I think it's ok to do it this way, although it does mean that we need to make it a hard requirement that all capability records can assign this (although there is the open question of whether a blocked user can also leave - they wouldn't be able to leave for other reasons, because they would not be able to write to the project any more).

An alternative approach would be to not include `LEFT_ROLE_ID` in the `roleAssignments` property and instead add this check to line 318 (`if (!ownCapabilities.roleAssignment.includes(roleId)) {`).

gmaclennan (commented):
This looks good to me, I think the behaviour is all correct, although there are a couple of changes that could be made to make it a bit less fragile and also a possible structure change:

- Comparing roles based on the `capability.name` property is fragile if we start storing these records in the database (instead of hard-coding them) then we need to remember to never allow the name to change. Listing role records I think is a more robust way of doing what you want (I think that would work, but haven't tried it myself).
- Possible change to whether we keep `LEFT_ROLE_ID` in the `roleAssignment` array, or treat it as a special case. I think it's fine as you have it, but we should document somewhere to make sure it is always included in `roleAssignment` if we use this approach.
- Looking at all the parameters that now need to be passed to a project instance, I wonder whether it makes sense to split `project.$leave` into two methods? A private `project[kleave]()` method and `manager.leaveProject(projectId)`. I think it's fine as it is, I think just that adding all these other parameters that a project instance must know about feels not quite right to me, but not sure if that's just aesthetics or there is a concrete reason for that.

I think the only thing missing is deleting the indexed data in the sqliteDb, but maybe that can be in a follow-up PR so that we can get this completed and merged?

src/mapeo-project.js
```diff
@@ -603,3 +673,43 @@ function mapAndValidateDeviceInfo(doc, { coreDiscoveryKey }) {
...
+
+  const deviceIdToCapsList = Object.entries(deviceIdToCaps)
+
+  if (deviceId in deviceIdToCaps) {
+    if (deviceIdToCapsList.length === 1) {
+      throw new Error('Cannot leave a project as the only device')
+    }
+
+    if (
+      deviceIdToCaps[deviceId].name ===
```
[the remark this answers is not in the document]
achou11: ah yes, makes sense

achou11:
> Looking at all the parameters that now need to be passed to a project instance, I wonder whether it makes sense to split `project.$leave` into two methods? A private `project[kleave]()` method and `manager.leaveProject(projectId)`. I think it's fine as it is, I think just that adding all these other parameters that a project instance must know about feels not quite right to me, but not sure if that's just aesthetics or there is a concrete reason for that.

yes this sounds like a good idea. don't think there's a reason for the big surface area of project instance that this PR adds aside from me not thinking about how to improve it initially 😅 you probably noticed a couple of fixes I made to a ts-ignored parameter in order to have the current implementation.

> Possible change to whether we keep LEFT_ROLE_ID in the roleAssignment array, or treat it as a special case. I think it's fine as you have it, but we should document somewhere to make sure it is always included in roleAssignment if we use this approach.

i don't have strong thoughts on it at the moment. given what https://github.com/digidem/mapeo-core-next/pull/410#pullrequestreview-1776816519 may entail, is one approach easier than the other?
 
> I think the only thing missing is deleting the indexed data in the sqliteDb, but maybe that can be in a follow-up PR so that we can get this completed and merged?

if it's as straightforward as what I postured then happy to include it in this PR. just didn't know i was missing any notable edge cases.

gmaclennan:
> > I think the only thing missing is deleting the indexed data in the sqliteDb, but maybe that can be in a follow-up PR so that we can get this completed and merged?
> 
> if it's as straightforward as what I postured then happy to include it in this PR. just didn't know if I was missing any notable edge cases.

I'm not sure either. Let's check-in about this and talk it through.

src/mapeo-project.js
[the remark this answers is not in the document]
achou11: this is getting trickier than expected. feels like doing some logic that's done by `Capabilities.getAll()` in order to account for edge cases 🤔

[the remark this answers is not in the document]
achou11: never mind, I figured it out 😄 addressed via https://github.com/digidem/mapeo-core-next/pull/410/commits/5fab0af0dd96be8a058718ec56778172e5115733

src/capabilities.js
achou11: addressed via https://github.com/digidem/mapeo-core-next/pull/410/commits/ca03ef35b9392a6cfd54196836ac892b30a60b89

gmaclennan (approved):

PR #417, opened 2023-12-13
Fix: pass doc to data type delete

tomasciccola:
we werent passing the original doc to delete, so when encoding the protobuf it would fail (since there were a bunch of missing fields)

### DIFF
--- a/src/datatype/index.js
+++ b/src/datatype/index.js
@@ -203,13 +203,15 @@ export class DataType extends TypedEmitter {
   /**
    * Not yet implemented
    * @param {string | string[]} versionId
+   * @param {T} value
    */
-  async delete(versionId) {
+  async delete(versionId, value) {
     await this.#dataStore.indexer.idle()
     const links = Array.isArray(versionId) ? versionId : [versionId]
     const { docId, createdAt, createdBy } = await this.#validateLinks(links)
     /** @type {any} */
     const doc = {
+      ...value,
       docId,
       createdAt,
       updatedAt: new Date().toISOString(),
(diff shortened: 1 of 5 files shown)


src/datatype/index.js
```diff
@@ -203,13 +203,15 @@ export class DataType extends TypedEmitter {
   /**
    * Not yet implemented
    * @param {string | string[]} versionId
+   * @param {T} value
    */
-  async delete(versionId) {
+  async delete(versionId, value) {
```
achou11, line 208: Shouldn't need to update the signature here (also this is prone to several issues around value integrity).  Internally, this method should be getting the existing doc via `getByVersionId()` and using that as the base value. Something along the lines of:

```ts
  async delete(versionId) {
    await this.#dataStore.indexer.idle()
    const links = Array.isArray(versionId) ? versionId : [versionId]
    const { docId, createdAt, createdBy } = await this.#validateLinks(links)

    const existingDoc = await this.getByVersionId(links[links.length - 1])

    /** @type {any} */
    const doc = {
      ...existingDoc,
      docId,
      createdAt,
      updatedAt: new Date().toISOString(),
      createdBy,
      links,
      schemaName: this.#schemaName,
      deleted: true,
    }
    await this.#dataStore.write(doc)
    return this.getByDocId(docId)
  }
```

technically could update `#validateLinks` to return the existing doc in its entirety to avoid doing repeat work, but not sure if i'd suggest that given that it mixes concerns a bit

achou11 (changes requested):

src/datatype/index.js
tomasciccola: yeah getting the doc by `versionId` makes sense (I was just passing the value to match `dataType.update`).

achou11:
can you add a test for this? ideally one that failed prior to this but now passes with the change?

tomasciccola:
I borked the branch when changing base, so here's the new PR #420  @achou11

PR #418, opened 2023-12-13
Chore: expose icon data type

tomasciccola:
this PR exposes `dataType`  on the `iconApi` , mostly so we can `getById` and `delete` icons (i.e. when importing a new config we delete every field, preset and icon to avoid duplicate stuff).
We could only expose `.getByVersionId` and `delete` instead of exposing all of `dataType`

### DIFF
--- a/src/icon-api.js
+++ b/src/icon-api.js
@@ -127,6 +127,10 @@ export class IconApi {
       })
     )
   }
+
+  get dataType() {
+    return this.#dataType
+  }
 }
 
 /**


achou11 (changes requested):
Can you elaborate on the reason for this? How and where would this be used?

Without any additional context, think it would make more sense to expose specific methods on the IconApi that use the dataType internally, as opposed to exposing the whole dataType.

tomasciccola:
> Can you elaborate on the reason for this? How and where would this be used?

Updated the PR description with more info

achou11:
hm yeah i think given the context, it would be preferable to expose methods (maybe private symbol ones, depending on where this is being used)

tomasciccola:
I'm closing this PR since I messed stuff with a rebase and this is the least time consuming solution :|, new PR is #421

PR #419, opened 2023-12-13
feat: add method for deleting core data from storage

achou11:
Towards #241 . Extracted from #410 

Implements a `CoreManager.deleteData()` method for deleting data from cores for a given namespace. By default, it will not delete data that resides in its writer core but that can be changed using the `deleteOwn` opt.

### DIFF
--- a/src/core-manager/index.js
+++ b/src/core-manager/index.js
@@ -478,6 +478,25 @@ export class CoreManager extends TypedEmitter {
     })
     return this.#corestore.replicate(stream)
   }
+
+  /**
+   * @param {Exclude<typeof NAMESPACES[number], 'auth'>} namespace
+   * @param {Object} [opts]
+   * @param {boolean} [opts.deleteOwn=false]
+   */
+  async deleteData(namespace, { deleteOwn = false } = {}) {
+    const coreRecords = this.getCores(namespace)
+    const ownWriterCore = this.getWriterCore(namespace)
+
+    const deletionPromises = []
+
+    for (const { core, key } of coreRecords) {
+      if (!deleteOwn && key === ownWriterCore.key) continue
+      deletionPromises.push(core.purge())
+    }
+
+    return Promise.all(deletionPromises)
+  }
 }
 
 /**
--- a/src/core-manager/index.js
+++ b/src/core-manager/index.js
@@ -491,7 +491,7 @@ export class CoreManager extends TypedEmitter {
     const deletionPromises = []
 
     for (const { core, key } of coreRecords) {
-      if (!deleteOwn && key === ownWriterCore.key) continue
+      if (!deleteOwn && key.equals(ownWriterCore.key)) continue
       deletionPromises.push(core.purge())
     }
 
--- a/src/core-manager/index.js
+++ b/src/core-manager/index.js
@@ -483,6 +483,7 @@ export class CoreManager extends TypedEmitter {
    * @param {Exclude<typeof NAMESPACES[number], 'auth'>} namespace
    * @param {Object} [opts]
    * @param {boolean} [opts.deleteOwn=false]
+   * @returns {Promise<void>}
    */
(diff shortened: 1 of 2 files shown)


src/core-manager/index.js
```diff
@@ -478,6 +478,25 @@ export class CoreManager extends TypedEmitter {
...
+   * @param {boolean} [opts.deleteOwn=false]
+   */
+  async deleteData(namespace, { deleteOwn = false } = {}) {
+    const coreRecords = this.getCores(namespace)
+    const ownWriterCore = this.getWriterCore(namespace)
+
+    const deletionPromises = []
+
+    for (const { core, key } of coreRecords) {
+      if (!deleteOwn && key === ownWriterCore.key) continue
```
gmaclennan, line 494: This is risky - I _think_ the `key` from coreRecords is the same instance of the key Buffer as the one used for the core instance, but Hypercore could change their implementation and it could be a copy.

```suggestion
      if (!deleteOwn && key.equal(ownWriterCore.key)) continue
```

gmaclennan (changes requested):
This looks good to me, but a bug that needs fixing.

In terms of tests, I would unit-test this, since even if we try to e2e test we still need to reach into the internals to actually check if it has worked.

One approach you might take:

1. Create a random-access-file storage function that saves storage names (e.g. `const createStorage = name => { storageNames.push(name); return raf(name) }`)
2. Create two core manager instances, write random data to each writer core in each namespace (can just be `(await cm.getWriterCore(ns)).append(Array(100).fill(null).map((_, i) => 'block' + i))`)
3. Replicate the two cores, download all the data.
4. Call the delete data function
5. Check the files are deleted from disk / not deleted, using filenames from the storageNames array. You'll [need this](https://github.com/holepunchto/corestore/blob/v6.8.4/index.js#L240) (where `id` is the discovery key) to filter which storage names you expect to be deleted if you test `deleteOwn: true`

It's a bit clunky but important to do I think - would need to catch something like the key equality bug here.

src/core-manager/index.js
achou11: ah yes, forgot that this is a buffer instance and not a string - thanks for spotting!

achou11: addressed via https://github.com/digidem/mapeo-core-next/pull/419/commits/afa53f35344465881c4e2a4e0677a28e3b8ec063

achou11:
@gmaclennan added test for testing the method and its effect on the filesystem. Wasn't sure about how to write a test that would catch the key equality bug though.

tests/core-manager.js
```diff
@@ -457,6 +458,163 @@ test('unreplicate', async (t) => {
...
+      )
+
+      st.ok(
+        peer2DataStorageDeletedForPeer1,
+        'peer 1 no longer has `data` storage for peer 2'
+      )
+    })
+
+    /// Delete their own data
+    await t.test('with deleteOwn=true', async (st) => {
```
achou11, line 593: thought it'd be easier to add this as part of the test but can split it out into its own test if that's preferred

tests/core-manager.js
```diff
@@ -457,6 +458,163 @@ test('unreplicate', async (t) => {
   })
 })
 
+test('deleteData() deletes relevant core storage files', async (t) => {
```
achou11, line 461: apologies for the length of this test. hopefully easy enough to follow but open to recommendations for improving it

gmaclennan: I'm fine with this, I can read it and it doesn't need to be perfect code. It works and tests what we need it to.

tests/core-manager.js
gmaclennan: Fine for it to be here I think.

gmaclennan (approved):
Looks good to me.

> Wasn't sure about how to write a test that would catch the key equality bug though.

The current tests catch it. The previous code would evaluate `key === core.key` as false, and I checked just putting `false` there, and the tests failed.

PR #420, opened 2023-12-14
grab existing doc when calling `dataType.delete`, test it

tomasciccola:


tests/data-type.js
```diff
@@ -141,6 +141,14 @@ test('test validity of `createdBy` field from another peer', async (t) => {
   await destroy()
 })
 
+test('test deletion of doc', async (t) => {
+  t.plan(1)
+  const projectKey = randomBytes(32)
+  const { dataType } = await testenv({ projectKey })
+  const { versionId } = await dataType.create(obsFixture)
+  t.execution(async () => await dataType.delete(versionId), `deletion failed`)
+})
```
achou11, line 150: some suggestions for this:

1. personal preference but probably would rename the test to something like `delete()`

2. the assertion at the end has an incorrect message. the message should describe what's happening when the assertion succeeds i.e. it should say `deletion succeeded`. although in this case, the assertion is not strictly necessary (see 3)

3. you don't need to use `t.execution()` here. Instead would call it without being wrapped and use the returned value for additional checks (see 4)

4. We should have additional assertions about what the returned deleted doc looks like e.g. deleted and updateAt properties change

example implementation would look something like:

```ts
test('delete()', async (t) => {
  const projectKey = randomBytes(32)
  const { dataType } = await testenv({ projectKey })
  const createdDoc = await dataType.create(obsFixture)

  t.is(createdDoc.deleted, false, '`deleted` field is `false` before deletion')

  const deletedDoc = await dataType.delete(createdDoc.versionId)

  t.is(deletedDoc.deleted, true, '`deleted` field is `true` for after deletion')

  t.not(createdDoc.updatedAt, deletedDoc.updatedAt, 'updatedAt has changed')

  t.alike(
    deletedDoc.links,
    [createdDoc.versionId],
    'deleted doc links back to created doc'
  )

  const retrievedDocByDocId = await dataType.getByDocId(deletedDoc.docId)

  t.alike(
    retrievedDocByDocId,
    deletedDoc,
    'retrieving doc by doc id returns deleted doc'
  )
})
```

could potentially add an assertion for making sure that the `getByVersionId()` also returns the deleted doc, but ran into needing to do additional work because it returns a value with a `forks` field

achou11 (changes requested):
Added a suggestion for improving the test

tests/data-type.js
tomasciccola: right, yeah, I'll add the comparison of the links and the retrieved `deleted` doc by docId

tests/data-type.js
```diff
@@ -141,6 +141,27 @@ test('test validity of `createdBy` field from another peer', async (t) => {
   await destroy()
 })
 
+test('delete()', async (t) => {
+  t.plan(4)
```
achou11, line 145: not really needed for a test like this. plan is mostly useful when dealing with things like event emitters or where the execution of code is not sequential

```suggestion
```

tests/data-type.js
```diff
@@ -141,6 +141,27 @@ test('test validity of `createdBy` field from another peer', async (t) => {
...
+  const projectKey = randomBytes(32)
+  const { dataType } = await testenv({ projectKey })
+  const doc = await dataType.create(obsFixture)
+  t.is(doc.deleted, false, `'deleted' field is false before deletion`)
+  const deletedDoc = await dataType.delete(doc.versionId)
+  t.is(deletedDoc.deleted, true, `'deleted field is true after deletion`)
```
achou11, line 151:
```suggestion
  t.is(deletedDoc.deleted, true, `'deleted' field is true after deletion`)
```

achou11 (approved):

PR #421, opened 2023-12-14
add iconApi.delete(docId)

tomasciccola:


achou11 (commented):
Change seems fine but wondering if it would make sense to expose a `delete()` on the IconApi method in general, which is what #405 really needs.

Something like:

```ts
/**
 * @param {string} docId
 */
async delete(docId) {
  // Could maybe extract into a public get() method for the IconApi if the need arises
  const doc = await this.#dataType.getByDocId(docId)
  // Maybe return the result here? idk
  await this.#dataType.delete(doc.versionId)
} 
```

Just trying to imagine the general usage of this API and the contexts in which it comes up. Not sure if anything using the IconApi should ever be concerned with versionIds - it seems like in practice we'd want to target the latest version of a doc and so we'd only need a docId to be provided, especially given that the `create()` method returns the docId. For example, would be nice for the api to look something like:

```ts
const iconApi = new IconApi(...)

const iconId = iconApi.create(...)

// Sometime in some distant future...
iconApi.delete(iconId)
```

Thoughts?

tomasciccola:
yeah, what you say makes a lot of sense, I'm going to implemented like that

tomasciccola:
I've added a couple of tests of the `iconApi.delete` method. 
I basically test that we get the correct `docId` of the deleted doc and that the `deleted` field is true

src/icon-api.js
```diff
@@ -94,6 +96,14 @@ export class IconApi {
     return blob
   }
 
+  /**
+   * @param {string} docId
+   */
+  async delete(docId) {
+    const { versionId } = await this.#dataType.getByDocId(docId)
+    return await this.#dataType.delete(versionId)
```
achou11, line 104: await not needed in this case

```suggestion
    return this.#dataType.delete(versionId)
```

achou11 (changes requested):
Another thing I just realized is that shouldn't this method also delete the blob in the hyperdrive? As of now, it's only marking it as deleted. Maybe could be done as a follow-up?

achou11:
Trying to reason about my question around deleting the actual blob as part of this method. Think part of my uncertainty is what it means to "delete". Does deletion mean you're marking the record as deleted and that's it, or does it also mean actually deleting any relevant data/assets? It's obvious to me that we eventually want to remove actual resources, but does it have to be tied to the the action of "deleting"? Would it make sense to have another action related to resources that uses that deletion marker as an indicator - basically terms similar to "garbage collection", "purge", "vacuum", etc

cc @gmaclennan

achou11 (approved):
Switching my evaluation to approve because i think the big question i posed can be addressed as a follow-up, regardless of answer

tomasciccola:
After a discussion with @gmaclennan today, we realize that - for now - Icons shouldn't be deleted, more so taking into account whats discussed in [here](https://github.com/digidem/mapeo-core-next/issues/422). 
After PR #405 is merged, there's probably gonna be a follow-up PR that introduces a better approach than deleting everything and re-importing the config.

Issue #422, opened 2023-12-18
fix: `dataType.delete()` should use `docId` not `versionId`


gmaclennan:
## Description

Reading through #420 made me realize that I think our current `delete()` method does not make sense. Trying to "delete" an arbitrary version of a doc doesn't actually do anything unless it is a "head". It will actually create a fork, which is kind of confusing.

e.g. we have a doc with docId `X`, created with versionId `A`. It is then edited so we have version `B`. Now a user tries to delete version `A`. They will create a new version `C`, and we now have two "heads" or "forks": `B` and `C`.

I think the delete method should actually accept the `docId`, since what you actually expect from deletion is that the document is no longer returned from `getMany()`. The delete function should:

1. Get the existing doc for that `docId`, and get the versionId
2. Check the `forks` property, and also get the existing doc(s) for those forks, to get the versionIds
3. Create a doc with `deleted: true` that has `links` with the versionIds above.
4. For the document data that is written to the delete doc, use the value from (1) (e.g. the doc that would be returned from `dataType.getByDocId(docId)`.

@achou11 @tomasciccola thoughts or comments?

gmaclennan:
This could raise the question, "how do I delete a fork", which you do as:

```js
const obs = await project.observation.getByDocId(docId)
const merged = await project.observation.update([obs.versionId, ...obs.forks], valueOf(obs))
```

Could look at creating a helper for this in the API if needed once we introduce a UI for managing forks.

achou11:
yeah i think this makes sense. So basically:

- `DataType.delete()` is reserved for deleting the latest locally known version of a record
- if needed, you could technically use `DataType.update()` to delete an older version or delete a fork?

gmaclennan:
yeah, although "deleting an older version" and "deleting a fork" doesn't really have a clear meaning.

PR #424, opened 2023-12-19
fix!: delete() by docId not versionId(s)

gmaclennan:
Fixes "`dataType.delete()` should use `docId` not `versionId`" (#422)

The deletion document links to all forks, if forks exist.


src/datatype/index.js
```diff
@@ -209,24 +209,22 @@ export class DataType extends TypedEmitter {
 
   /**
    * Not yet implemented
```
achou11, line 211: can we remove this as part of this PR? seems pretty implemented at this point 😁

achou11 (approved):

PR #425, opened 2023-12-20
chore: add tsconfig for tests/ directory

### LINKED ISSUE #416: type checking in `tests/` directory doesn't use tsconfig (fully?)
Working on a new file in `tests/` and noticed that our vendored digidem types definitions aren't being detected in it. Most of the files in that directory don't have typechecking enabled and the ones that do are done via `// @ts-check` in order to gradually integrate typechecking in that directory. However, with the `// @ts-check`, it seems like TS is just inferring types based on reading the source module as opposed to using the tsconfig to resolve some of the external module types, which works in some cases but overall doesn't work given the complexity of some of the hypercore-related types.

if you do `npx tsc --showConfig`, you'll notice that - with the exception of one file, `tests/schema.js` (which is explicitly listed in the `files` tsconfig field) - none of the `tests/` files with the `// @ts-check` annotation are listed in the `files` field i.e. having the tsconfig being applied to it. This leads me to believe that using `// @ts-check` doesn't (fully) apply the closest found `tsconfig.json` file as I expected. I find it hard to believe that TS will just apply its default configuration in this case, so maybe it does something clever that isn't transparent to me. 

The solution would be to update the tsconfig to either

- add each file in the `files` field (granular, but tedious)
- adding the `tests/**/*` in the `include` field (less tedious but then we have to deal with all of the type errors that haven't been addressed)

Was able to confirm that vendored type resolution is fixed applying either of the above options.

achou11:
Closes #416 

This doesn't include any npm script updates or a step to typecheck the tests. Mostly useful as we're writing tests for editor feedback for now.

Didn't feel inclined to include fixes to the type issues that were found as a result of this PR. Happy to try since I believe most are probably straightforward to address. 

Current state of checks as of now:

```sh
Found 173 errors in 16 files.

Errors  Files
    14  tests/bitfield-rle.js:7
     4  tests/blob-store/blob-store.js:416
     5  tests/blob-store/combine-states.js:40
    12  tests/blob-store/live-download.js:19
    53  tests/core-manager.js:25
     1  tests/core-ownership.js:79
     4  tests/data-type.js:123
     4  tests/datastore.js:29
    23  tests/discovery/dns-sd.js:19
     8  tests/discovery/local-discovery.js:69
     4  tests/fastify-plugins/blobs.js:295
     7  tests/helpers/replication-state.js:3
    13  tests/invite-api.js:33
     6  tests/local-peers.js:220
     9  tests/sync/core-sync-state.js:328
     6  tests/sync/namespace-sync-state.js:111
```

### DIFF
--- a/tests/tsconfig.json
+++ b/tests/tsconfig.json
@@ -1,0 +1,5 @@
+{
+  "extends": "../tsconfig.json",
+  "files": [],
+  "include": ["./**/*", "../types/**/*.d.ts"]
+}
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -22,7 +22,6 @@
       "*": ["./types", "./node_modules/@digidem/types/vendor/*/index.d.ts"]
     }
   },
-  "files": ["tests/schema.js"],
   "include": ["src/**/*", "types/**/*.d.ts", "test-e2e/**/*"],
   "exclude": ["node_modules", "src/generated/*.js"]
 }


gmaclennan (approved):
Looks good to me. I'd be inclined to just put `// @ts-nocheck` at the top of these existing files, I don't think we gain anything from jumping through hoops to make typescript happy, and it's possible that fixing some of the type issues could reduce test coverage. I can see the utility when writing new test files, and then electing to ignore errors and then add `// @ts-nocheck` when done.

achou11:
> I'd be inclined to just put `// @ts-nocheck` at the top of these existing files

Do you prefer I include that as part of this PR? Personally I'd even be fine with not doing that as long as we don't have plans to include a typechecking step for these files. I'm mostly interested in just making it easier to write the tests with the TS-powered assistance

PR #430, opened 2024-01-11
fix: remove stale project reference in manager when leaving a project

achou11:
I _think_ this is needed in order to avoid some bugs related to using a stale project reference. Also have a todo to write an e2e test for leaving and then rejoining, since my early attempt has not been successful 🤔 (which may highlight other issues?)


gmaclennan (approved):
Looks ok to me.

I think it's ok to leave a test for rejoining until later (post-MVP) since we are not supporting that flow yet.

PR #435, opened 2024-01-16
feat: support serving static maps

achou11:
Closes https://github.com/digidem/mapeo-core-next/issues/437

Towards #439 

~Stacked on #440~ 

Combines changes extracted from https://github.com/digidem/mapeo-map-server/pull/101 and https://github.com/digidem/CoMapeo-mobile/pull/155

Notes:

- Implements a generalized Fastify plugin for handling more general maps-related functionality. 
  - declares a dependency on a plugin that implements functionality related to serving the static maps, which is functionally similar to https://github.com/digidem/mapeo-server
  - implements a portion of #439


src/fastify-plugins/static-maps.js
```diff
@@ -0,0 +1,269 @@
...
+ * @param {string} baseDirectory
+ * @param {import('@sinclair/typebox').Static<typeof GetStaticMapTileParamsSchema>} params
+ * @returns {null | { data: Buffer, mimeType: string | null, shouldGzip: boolean}}
+ */
+function getStyleTileInfo(baseDirectory, params) {
+  const { styleId, tileId, z, x, y } = params
+  let { ext } = params
+
+  // TODO: If necessary, need to flip the y value first if the TileJSON source is TMS scheme
+  // Doing this will depend on if we decide that the asar directory structure should only follow XYZ or if it should align with corresponding tilejson spec
```
achou11, line 230: highlighting this. usually this is something that we have to account for, but I think in this case it depends on how the ASAR is created/structured e.g. are people always creating an asar that follows the XYZ scheme, or will they create something that matches the scheme specified in associated the style.json file?

src/media-server.js
```diff
@@ -51,6 +54,13 @@ export class MediaServer {
       getProject,
     })
 
+    if (staticMapsRootDir) {
+      this.#fastify.register(StaticMapsPlugin, {
+        prefix: MAPS_PREFIX,
+        staticRootDir: staticMapsRootDir,
+      })
```
achou11, line 61: for now we register this plugin directly. can imagine in the future where we want to implement more of the general map server, we'd have a more general "maps" wrapper plugin that would register the static maps plugin, and then register the wrapper here instead.

one thing at a time...

src/fastify-plugins/maps/index.js
achou11, line 1: this file serves as the entry point for all maps-related server functionality. this PR introduces just the static custom offline maps, which is a plugin that's implemented separately and then registered here. in the future this will register other plugins related to stable style.json resolution, sqlite-stored maps, etc

src/mapeo-manager.js
```diff
@@ -92,7 +92,12 @@ export class MapeoManager extends TypedEmitter {
    * @param {string} opts.projectMigrationsFolder path for drizzle migrations folder for project database
    * @param {string} opts.clientMigrationsFolder path for drizzle migrations folder for client database
    * @param {string | import('./types.js').CoreStorage} opts.coreStorage Folder for hypercore storage or a function that returns a RandomAccessStorage instance
-   * @param {{ port?: number, logger: import('fastify').FastifyServerOptions['logger'] }} [opts.mediaServerOpts]
+   * @param {Object} [opts.mediaServerOpts]
+   * @param {import('fastify').FastifyServerOptions['logger']} [opts.mediaServerOpts.logger]
+   * @param {Object} [opts.mediaServerOpts.mapsOpts]
+   * @param {string } [opts.mediaServerOpts.mapsOpts.staticMapsFolder]
```
achou11, line 98: not in love with the nesting but wanted the flexibility to add more options under the maps scope in the future

src/media-server.js
```diff
@@ -30,8 +32,10 @@ export class MediaServer {
    * @param {object} params
    * @param {(projectPublicId: string) => Promise<import('./mapeo-project.js').MapeoProject>} params.getProject
    * @param {import('fastify').FastifyServerOptions['logger']} [params.logger]
+   * @param {object} [params.mapsOpts]
+   * @param {string} [params.mapsOpts.staticMapsRootDir] Absolute path to directory containing static maps. If not specified, serving static maps is disabled.
```
achou11, line 36: similar comment about future flexibility provided by having a maps-related scoped opt

src/fastify-plugins/maps/static-maps.js
```diff
@@ -0,0 +1,266 @@
...
+/**
+ * @param {string} baseDirectory
+ * @param {import('@sinclair/typebox').Static<typeof GetStaticMapTileParamsSchema>} params
+ * @returns {null | { data: Buffer, mimeType: string | null, shouldGzip: boolean}}
+ */
+function getStyleTileInfo(baseDirectory, params) {
+  const { styleId, tileId, z, x, y } = params
+  let { ext } = params
+
+  // TODO: If necessary, need to flip the y value first if the TileJSON source is TMS scheme
```
achou11, line 228: this is still a relevant question, just forgot to include in my previous review

src/fastify-plugins/maps/index.js
```diff
@@ -0,0 +1,24 @@
...
+ */
+
+/** @type {import('fastify').FastifyPluginAsync<MapsPluginOpts>} */
+async function mapsPlugin(fastify, opts) {
+  if (opts.staticRootDir) {
+    fastify.register(StaticMapsPlugin, {
+      prefix: opts.prefix,
+      staticRootDir: opts.staticRootDir,
+    })
+  }
```
achou11, line 23: wondering if i should just avoid conditionally registering here for the sake of simplifying the implementation a bit. not sure if there's a case where the end user wouldn't want to support the static map serving functionality

src/fastify-plugins/maps/static-maps.js
```diff
@@ -0,0 +1,266 @@
...
+  // Doing this will depend on if we decide that the asar directory structure should only follow XYZ or if it should align with corresponding tilejson spec
```
achou11, line 229: highlighting this. are we able to assume that the tile urls use z/x/y format and the directories that are stored on the filesystem follow that convention?

src/fastify-plugins/maps/index.js
```diff
@@ -0,0 +1,40 @@
...
+/** @type {import('fastify').FastifyPluginAsync<MapsPluginOpts, import('fastify').RawServerDefault, import('@fastify/type-provider-typebox').TypeBoxTypeProvider>} */
+async function routes(fastify) {
+  fastify.get('/style.json', async (req, rep) => {
+    const serverAddress = await getFastifyServerAddress(req.server.server)
+    // TODO: Figure out how to do interface merging: https://fastify.dev/docs/latest/Reference/TypeScript/#plugins
+    // @ts-expect-error
+    const { data, headers } = await fastify.mapeoStaticMaps.getStyleJsonInfo(
+      'default',
+      serverAddress
+    )
```
achou11, line 35: not sure how to (easily) do interface merging in this case. creating a d.ts file for that plugin seems rather tedious and prone to mistakes, but maybe that's the only way?

src/fastify-plugins/maps/static-maps.js
```diff
@@ -0,0 +1,279 @@
...
+      const { styleId } = req.params
+
+      const serverAddress = await getFastifyServerAddress(req.server.server)
+
+      // TODO: Figure out how to do interface merging: https://fastify.dev/docs/latest/Reference/TypeScript/#plugins
+      // @ts-expect-error
+      const { data, headers } = await fastify.mapeoStaticMaps.getStyleJsonInfo(
+        styleId,
+        serverAddress
+      )
```
achou11, line 197: same comment about doing interface merging

tests/fastify-plugins/maps.js
```diff
@@ -0,0 +1,82 @@
...
+        url: '/maps/style.json',
+      }),
+      'endpoint found at specified prefix'
+    )
+  }
+})
+
+// TODO: Currently fails
+// Requires fixture setup that has default/ in static directory, online proxy, then offline fallback map
+skip('/style.json resolves with expected style.json', async (t) => {
```
achou11, line 52: will enable this test when working on #439

src/fastify-plugins/maps/static-maps.js
```diff
@@ -0,0 +1,284 @@
...
+      }
+
+      rep.headers({
+        'Content-Type': 'application/json; charset=utf-8',
+        'Cache-Control': 'max-age=' + 5 * 60, // 5 minutes
+        'Access-Control-Allow-Headers':
+          'Authorization, Content-Type, If-Match, If-Modified-Since, If-None-Match, If-Unmodified-Since',
+        'Access-Control-Allow-Origin': '*',
+        'Last-Modified': new Date(stats.mtime).toUTCString(),
+        'Content-Length': Buffer.from(styleJson).length,
```
EvanHahn, line 202: tl;dr: this could be improved slightly but is fine. Feel free to ignore the rest of this comment.

---

Nitpick: there are some unnecessary byte-string conversions in this flow:

1. Load *bytes* from disk, convert them to a *string* (using `fs.readFile(..., 'utf-8')`).
1. Convert the *string* to *bytes* to compute the `Content-Length` (this line).
1. Convert the *string* to *bytes* to send the data.

In practice, these conversions probably have minimal impact, and it's unlikely that the file will ever be too long to fit in a string[^0]. But in my deranged mind, this tiny imperfection cannot stand.

<details>
<summary>I implemented a full solution that does all operations on Buffers:</summary>

```diff
diff --git a/src/fastify-plugins/maps/static-maps.js b/src/fastify-plugins/maps/static-maps.js
index e6927ee..96fe965 100644
--- a/src/fastify-plugins/maps/static-maps.js
+++ b/src/fastify-plugins/maps/static-maps.js
@@ -7,6 +7,7 @@ import asar from '@electron/asar'
 import { Mime } from 'mime/lite'
 import standardTypes from 'mime/types/standard.js'
 
+import { bufferReplace } from '../../utils.js'
 import { NotFoundError, getFastifyServerAddress } from '../utils.js'
 
 export const PLUGIN_NAME = 'mapeo-static-maps'
@@ -24,7 +25,7 @@ export const plugin = fp(staticMapsPlugin, {
 
 /**
  * @typedef {object} StaticMapsPluginDecorator
- * @property {(styleId: string, serverAddress: string) => Promise<string>} getResolvedStyleJson
+ * @property {(styleId: string, serverAddress: string) => Promise<Buffer>} getResolvedStyleJson
  * @property {(styleId: string) => Promise<import('node:fs').Stats>} getStyleJsonStats
  */
 
@@ -34,10 +35,11 @@ async function staticMapsPlugin(fastify, opts) {
     async getResolvedStyleJson(styleId, serverAddress) {
       const filePath = path.join(opts.staticRootDir, styleId, 'style.json')
 
-      const data = await fs.readFile(filePath, 'utf-8')
+      const data = await fs.readFile(filePath)
 
-      return data.replace(
-        /\{host\}/gm,
+      return bufferReplace(
+        data,
+        '{host}',
         new URL(`${opts.prefix || ''}/${styleId}`, serverAddress).href
       )
     },
@@ -199,7 +201,7 @@ async function routes(fastify, opts) {
           'Authorization, Content-Type, If-Match, If-Modified-Since, If-None-Match, If-Unmodified-Since',
         'Access-Control-Allow-Origin': '*',
         'Last-Modified': new Date(stats.mtime).toUTCString(),
-        'Content-Length': Buffer.from(styleJson).length,
+        'Content-Length': styleJson.byteLength,
       })
 
       return styleJson
diff --git a/src/utils.js b/src/utils.js
index 8ba6d0e..69d2ce5 100644
--- a/src/utils.js
+++ b/src/utils.js
@@ -123,6 +123,52 @@ export function deNullify(obj) {
   return /** @type {import('./types.js').NullableToOptional<T>} */ (objNoNulls)
 }
 
+/**
+ * @param {Readonly<string | Buffer>} stringOrBuffer
+ * @returns {Buffer}
+ */
+function bufferify(stringOrBuffer) {
+  return Buffer.isBuffer(stringOrBuffer)
+    ? stringOrBuffer
+    : Buffer.from(stringOrBuffer)
+}
+
+/**
+ * @param {Readonly<Buffer>} buffer
+ * @param {Readonly<string | Buffer>} pattern
+ * @param {Readonly<string | Buffer>} replacement
+ * @returns {Buffer}
+ */
+export function bufferReplace(buffer, pattern, replacement) {
+  const patternLength = bufferify(pattern).byteLength
+  replacement = bufferify(replacement)
+
+  /** @type {Buffer[]} */
+  const chunks = []
+  let totalLength = 0
+
+  /** @param {Buffer} chunk */
+  const addChunk = (chunk) => {
+    chunks.push(chunk)
+    totalLength += chunk.byteLength
+  }
+
+  while (buffer.byteLength) {
+    const index = buffer.indexOf(pattern)
+    if (index === -1) {
+      addChunk(buffer)
+      break
+    }
+
+    addChunk(buffer.subarray(0, index))
+    addChunk(replacement)
+
+    buffer = buffer.subarray(index + patternLength)
+  }
+
+  return Buffer.concat(chunks, totalLength)
+}
+
 /**
  * @template {import('@mapeo/schema').MapeoDoc & { forks?: string[] }} T
  * @param {T} doc
diff --git a/tests/utils.js b/tests/utils.js
index e80a195..a3e4871 100644
--- a/tests/utils.js
+++ b/tests/utils.js
@@ -1,6 +1,11 @@
 // @ts-check
 import test from 'brittle'
-import { assert, ExhaustivenessError, setHas } from '../src/utils.js'
+import {
+  assert,
+  ExhaustivenessError,
+  setHas,
+  bufferReplace,
+} from '../src/utils.js'
 
 test('assert()', (t) => {
   t.execution(() => assert(true, 'should work'))
@@ -27,3 +32,16 @@ test('setHas()', (t) => {
   t.ok(setHas(set)(1))
   t.absent(setHas(set)(9))
 })
+
+test('bufferReplace()', (t) => {
+  const fooBarBaz = Buffer.from('foo bar baz')
+
+  const expected = Buffer.from('foo zor zoz')
+  t.alike(bufferReplace(fooBarBaz, 'ba', 'zo'), expected)
+  t.alike(
+    bufferReplace(fooBarBaz, Buffer.from('ba'), Buffer.from('zo')),
+    expected
+  )
+
+  t.alike(bufferReplace(fooBarBaz, 'x', 'ignored'), fooBarBaz)
+})
```
</details>

That's a fairly heavyweight solution, though. Instead, would you move the `Buffer.from(styleJson)` call up so that it can be used to compute the `Content-Length` _and_ for the result? This should remove a needless conversion.

```diff
diff --git a/src/fastify-plugins/maps/static-maps.js b/src/fastify-plugins/maps/static-maps.js
index e6927ee..1b97131 100644
--- a/src/fastify-plugins/maps/static-maps.js
+++ b/src/fastify-plugins/maps/static-maps.js
@@ -192,6 +192,8 @@ async function routes(fastify, opts) {
         throw new NotFoundError(`id = ${styleId}, style.json`)
       }
 
+      const styleJsonBytes = Buffer.from(styleJson)
+
       rep.headers({
         'Content-Type': 'application/json; charset=utf-8',
         'Cache-Control': 'max-age=' + 5 * 60, // 5 minutes
@@ -199,10 +201,10 @@ async function routes(fastify, opts) {
           'Authorization, Content-Type, If-Match, If-Modified-Since, If-None-Match, If-Unmodified-Since',
         'Access-Control-Allow-Origin': '*',
         'Last-Modified': new Date(stats.mtime).toUTCString(),
-        'Content-Length': Buffer.from(styleJson).length,
+        'Content-Length': styleJsonBytes.length,
       })
 
-      return styleJson
+      return styleJsonBytes
     }
   )
```

[^0]: In my cursory testing, the max length of a string is 536,870,888 characters. Not sure what a "character" means in this case.

src/fastify-plugins/maps/static-maps.js
```diff
@@ -0,0 +1,284 @@
...
+
+  // extension check isn't fully necessary since the buffer will only exist if the extension exists
+  // but useful to check for types reasons
+  if (!data || !ext) {
+    return null
+  }
+
+  const mimeType = mime.getType(ext)
+
+  // Set gzip encoding on {mvt,pbf} tiles.
```
EvanHahn, line 280: Nit: I think the code is clear without this comment.

```suggestion
```

src/fastify-plugins/maps/static-maps.js
```diff
@@ -0,0 +1,284 @@
...
+  const shouldGzip = /mvt|pbf$/.test(ext)
```
EvanHahn, line 281: Nitpick: This [will match][0] any string that ends with `pbf` or contains `mvt`. I may be wrong, but I don't think that's what you want.

```suggestion
  const shouldGzip = ext === 'mvt' || ext === 'pbf'
```

[0]: https://regex101.com/r/srscys/2

src/fastify-plugins/maps/static-maps.js
EvanHahn: Not my most helpful comment, but want to explicitly say: I don't know. My hunch is that we should get an answer to this before the MVP, but not before merging this.

src/fastify-plugins/maps/static-maps.js
```diff
@@ -0,0 +1,284 @@
...
+    }
+
+    const { data, mimeType, shouldGzip } = result
+
+    if (mimeType) {
+      rep.header('Content-Type', mimeType)
+    }
+
+    if (shouldGzip) {
+      rep.header('Content-Encoding', 'gzip')
```
EvanHahn, line 103: 1. Does this actually compress the data?
2. I would think we'd want to do content negotiation (i.e. respect the [Accept-Encoding request header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Encoding)) here.

tests/fixtures/maps/streets-sat-style/index.html
EvanHahn, line 1: Is this file necessary?

tests/fastify-plugins/maps.js
```diff
@@ -0,0 +1,81 @@
+import { test } from 'brittle'
```
EvanHahn, line 1: Nitpick:

```suggestion
// @ts-check
import { test } from 'brittle'
```

tests/fastify-plugins/static-maps.js
```diff
@@ -0,0 +1,261 @@
+import fs from 'node:fs'
```
EvanHahn, line 1: Nitpick:

```suggestion
// @ts-check
import fs from 'node:fs'
```

src/fastify-plugins/maps/index.js
```diff
@@ -0,0 +1,64 @@
+import fp from 'fastify-plugin'
```
EvanHahn, line 1: Nitpick:

```suggestion
// @ts-check
import fp from 'fastify-plugin'
```

src/fastify-plugins/maps/static-maps.js
```diff
@@ -0,0 +1,284 @@
+import path from 'path'
```
EvanHahn, line 1: Nitpick:

```suggestion
// @ts-check
import path from 'path'
```

src/fastify-plugins/maps/index.js
```diff
@@ -0,0 +1,64 @@
...
+  dependencies: [MAPEO_STATIC_MAPS],
+})
+
+/**
+ * @typedef {object} MapsPluginOpts
+ * @property {string} [prefix]
+ */
+
+/** @type {import('fastify').FastifyPluginAsync<MapsPluginOpts>} */
+async function mapsPlugin(fastify, opts) {
```
EvanHahn, line 21: Nit: does this need to be async? (Same comment applies for `routes`.)

EvanHahn (changes requested):

EvanHahn:
Had the following thought while I was reviewing: you've split this into multiple Fastify plugins. Why not just have one big plugin? I worry this is a premature abstraction...are there ever cases where you wouldn't include all the plugins?

src/fastify-plugins/maps/index.js
achou11: doesn't need to be but then you'd have to write the plugin using a callback approach, which i think more or less pans out the same way. we've been using the `async` approach for uniformity if anything i guess

src/fastify-plugins/maps/static-maps.js
achou11: same question about whether this is needed if already included by the tsconfig

src/fastify-plugins/maps/static-maps.js
achou11: at the moment, no - can probably remove this for now but we may need it at some point (legacy implementation applied gzip for `mvt` and `pbf` sources), but can be addressed later

> I would think we'd want to do content negotiation

Yeah that makes sense - not necessary if we omit this altogether for now though right?

achou11:
> Had the following thought while I was reviewing: you've split this into multiple Fastify plugins. Why not just have one big plugin? I worry this is a premature abstraction...are there ever cases where you wouldn't include all the plugins?

would could categorize as premature but useful (also easier to build iteratively since they're all rather distinct functionalities). 

the idea was to build this in a way that could potentially be split off into its own package that contais all of these plugins. to summarize the purpose of each plugin:

- `maps`: mostly specific to the "new" maps serving approach for Mapeo. the initial implementation uses a decorator exposed by static maps plugin but that's to support backward compatibility for how maps are stored on a device. eventually, this will primarily use the JS API for the map server dep that you've been working on, and expose the routes necessary to power the Mapeo Map Manager.

- `static-maps`: adding functionality for the "legacy" map serving approach that Mapeo uses i.e. an embedded directory with a specific structure and some metadata files that get served statically. this is isolated from the new approach and but we need the functionality for backwards compat reasons to ease the transition.

- `offline-map`: very specific to how mapeo works and not necessarily needed

src/fastify-plugins/maps/static-maps.js
achou11: looks like i simplify the implementation here in this PR: https://github.com/digidem/mapeo-core-next/pull/450/files#diff-38bb092a0481e2a5e86a9d4af8f6fc9fb9c5c1e7283570ef0d146c97be8836a4

so potentially moot to a certain extent 😄

tests/fastify-plugins/static-maps.js
```diff
@@ -0,0 +1,261 @@
...
+
+      t.not(
+        data,
+        JSON.parse(rawStyleJson),
+        'response data is different from raw style file content'
+      )
+      t.alike(
+        data,
+        JSON.parse(rawStyleJson.replace(/\{host\}/gm, `${address}/${styleId}`)),
+        'response data has correct'
```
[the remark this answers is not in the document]
achou11: addressed via https://github.com/digidem/mapeo-core-next/pull/435/commits/f459f3fdf0a6729d9477e053941f9d7d1498a757

tests/fastify-plugins/static-maps.js
```diff
@@ -0,0 +1,261 @@
...
+  const response = await server.inject({
+    method: 'GET',
+    url: '/',
+  })
+
+  t.is(response.statusCode, 200)
+
+  const data = response.json()
+
+  t.is(data.length, 2, 'data has expected number of items')
```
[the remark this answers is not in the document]
achou11: addressed via https://github.com/digidem/mapeo-core-next/pull/435/commits/f459f3fdf0a6729d9477e053941f9d7d1498a757

tests/fastify-plugins/static-maps.js
```diff
@@ -0,0 +1,261 @@
...
+
+      t.is(response.statusCode, 200)
+
+      const data = response.json()
+
+      t.not(
+        data,
+        JSON.parse(rawStyleJson),
+        'response data is different from raw style file content'
+      )
```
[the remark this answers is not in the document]
achou11: addressed via https://github.com/digidem/mapeo-core-next/pull/435/commits/f459f3fdf0a6729d9477e053941f9d7d1498a757

tests/fastify-plugins/static-maps.js
```diff
@@ -0,0 +1,261 @@
...
+
+  t.is(response.statusCode, 200)
+
+  t.is(
+    response.headers['content-type'],
+    'application/x-protobuf',
+    'content type correct'
+  )
+
+  t.is(response.headers['content-encoding'], 'gzip', 'gzip encoding enabled')
```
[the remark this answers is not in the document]
achou11: addressed via https://github.com/digidem/mapeo-core-next/pull/435/commits/773a3d65489593bfbd93cb9fe10329747cdbb6d9

src/fastify-plugins/maps/static-maps.js
```diff
@@ -0,0 +1,284 @@
...
+              )
+
+              styleJson = JSON.parse(styleJsonContent)
+            } catch (err) {
+              return null
+            }
+
+            const serverAddress = await getFastifyServerAddress(
+              req.server.server
+            )
```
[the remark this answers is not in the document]
achou11: addressed via https://github.com/digidem/mapeo-core-next/pull/435/commits/4f7f0178f23e7d921fe927160de5cae21b9091a9

src/fastify-plugins/maps/static-maps.js
achou11: addressed via https://github.com/digidem/mapeo-core-next/pull/435/commits/01e881761320c06d3da6af1346670804fb7145a5

src/fastify-plugins/maps/static-maps.js
achou11: addressed via https://github.com/digidem/mapeo-core-next/pull/435/commits/01e881761320c06d3da6af1346670804fb7145a5

src/fastify-plugins/maps/static-maps.js
achou11: addressed via https://github.com/digidem/mapeo-core-next/pull/435/commits/773a3d65489593bfbd93cb9fe10329747cdbb6d9

src/fastify-plugins/maps/static-maps.js
```diff
@@ -0,0 +1,279 @@
...
+
+  // extension check isn't fully necessary since the buffer will only exist if the extension exists
+  // but useful to check for types reasons
+  if (!data || !ext) {
+    return null
+  }
+
+  const mimeType = mime.getType(ext)
+
+  const shouldGzip = ext === 'mvt' || ext === 'pbf'
```
EvanHahn, line 276: Should we remove this now that it's unused?

EvanHahn (approved):
Thanks for addressing all my nitpicks and answering my questions!

src/fastify-plugins/maps/static-maps.js
achou11: ah yeah forgot about that - addressed via https://github.com/digidem/mapeo-core-next/pull/435/commits/771f7dffa070a51be7d223124cda1e37fb7521e4

Issue #437, opened 2024-01-18
Add support for serving custom offline maps 


achou11:
## Description

The initial intention was to implement this as part of the map server (see https://github.com/digidem/mapeo-map-server/issues/71) but we decided that for now, the initial implementation should live in this repo. This provides the following benefits:

- avoiding the additional work needed to get the map server dep in a place for easy-enough integration into the mobile app. the functionality for this specific task is not reliant on how the map server dep works in general (which is based on sqlite storage).
- reducing the number of server instances that the app needs to manage. we already have the MediaServer which basically manages a fastify instance for the purpose of serving resources like icons ands blobs. We can take advantage of fastify's plugin architecture and add another plugin to handle this maps feature.
 
## Tasks

- [ ] Create fastify plugin dealing with maps and integrate with Media Server
- [ ] Rename the MediaServer to something more generic. Maybe something like MapeoHttpServer, ResourceServer, AppServer??

PR #440, opened 2024-01-18
chore!: refactor MapeoManager to accept Fastify server instance

achou11:
Towards #437 

As I'm working on #437 and dealing with more Fastify instantiation and plugins, I'm noticing that there's an ergonomic awkwardness caused by the MapeoManager being responsible for _instantiating_ the media server instance. Basically was becoming more and more prone to defining MediaServer-specific options for the MapeoManager constructor, which didn't make much sense since the manager mostly just passes it along to the media server.

The ideal approach is only needing to provide an _instance_ of the MediaServer to instantiate the MapeoProject. I initially thought that this wasn't possible because some of the plugins we register for the media server are dependent on methods provided by the manager. The flow of operations was roughly:

1. Instantiate MapeoManager
2. Manager instantiates MediaServer internally
3. MediaServer internally registers plugins dependent on MapeoManager methods

Then I realized that even if that's the case, those methods are only needed when we _register_ the relevant plugins. With this in mind, this PR updates the MediaServer and MapeoManager such that we can now have the following flow of operations instead:

1. Instantiate MediaServer
2. Instantiate MapeoManager, providing instance of MediaServer
3. MapeoManager _registers_ plugins dependent on MapeoManager methods

This makes it much easier to integrate new plugins. In the context of the maps-related functionality, none of the implementation is reliant on the MapeoManager so it can be registered outside of the MapeoManager more freely, meaning we don't need to specify plugin-specific opts for the manager that are just relayed. E.g.

```ts
const mediaServer = new MediaServer()
// Will register Blobs and Icons plugins internally
const manager = new MapeoManager( { mediaServer, ... })

// Can register manager independent plugins wherever and whenever!
mediaServer.registerPlugin(MapsPlugin, ...)
```

---

Remaining questions:

- ~Do we need `MapeoManager.startMediaServer()` and `MapeoManager.stopMediaServer()` anymore with this change? Those are just exposing the relevant media server methods but if the media server is available outside the context of the manager, could we just defer to using `mediaServer.start()` and `mediaServer.close()`?~ EDIT: no longer needed
- ~Should we rename the MediaServer at this point? Given this note in #437:~ DONE: renamed to `FastifyController` and reduced in scope

> Rename the MediaServer to something more generic. Maybe something like MapeoHttpServer, ResourceServer, AppServer??

Notes:

- Now that the HTTP server  is instantiated separately, consumers of this package will need to instantiate the server themselves and then provide that when instantiating the manager. Should probably update any relevant docs with that


src/mapeo-manager.js
```diff
@@ -654,13 +665,15 @@ export class MapeoManager extends TypedEmitter {
...
    * @param {import('./media-server.js').StartOpts} [opts]
    */
   async startMediaServer(opts) {
     await this.#mediaServer.start(opts)
   }
 
+  // TODO: Can we remove this?
   async stopMediaServer() {
     await this.#mediaServer.stop()
   }
```
achou11, line 679: highlighting these - can these now be removed or is it still helpful to have them?

src/media-server.js
```diff
@@ -27,36 +24,35 @@ export class MediaServer {
...
 
     this.#serverState = new StateMachine({
       start: this.#startServer.bind(this),
       stop: this.#stopServer.bind(this),
     })
   }
 
+  get [kFastify]() {
+    return this.#fastify
+  }
```
achou11, line 45: wasn't sure how useful it is to expose the fastify instance. maybe it is, maybe it isn't 🤷 (currently not being used anywhere)

gmaclennan (commented):
This makes a lot of sense, I like how it simplifies things. I'm tempted to go even further, and just make the parameter for MapeoManager to be a fastify instance. Then the `getServerAddress(fastify.server)` can just be defined in Mapeo code, and what is called MediaServer could just be `FastifyController` or `FastifyStopStart` - could break into a separate module if we want. `getMediaAddress` could just be a private method on MapeoManager, or we could wrap the logic into MapeoProject methods. 

I don't think we need to export `start/stopMediaServer` on the Manager instance - we start and stop the server from the backend, so it doesn't need to be exposed in the API. 

Could make a fastify instance a param of FastifyController too, so code would be:

```js
const fastify = Fastify({ logger })
const fastifyController = new FastifyController({ fastify })
const manager = new MapeoManager({ fastify, ... })
fastifyController.start()
```

A bit verbose maybe, but clear.

src/media-server.js
gmaclennan: Can be useful in unit tests using `fastify.inject()`. We don't have those yet though, and if we create the fastify instance outside the class (see suggestion) then not needed.

src/mapeo-manager.js
gmaclennan: no, can be removed. Needs change to CoMapeo code.

achou11:
@gmaclennan yep makes a lot of sense! will incorporate that suggestion into this PR (won't create a separate module for now)

achou11:
@gmaclennan made the following changes based on your feedback:

- renamed MediaServer to FastifyController
- updated MapeoManager to:
  - accept Fastify instance as constructor param instead of MediaServer instance
  - implement private getMediaAddress() method (used to live in MediaServer)
- update relevant tests (file names and tests)
- renamed and updated docs

docs/guides/mapeo-http-server.md
```diff
@@ -1,28 +1,37 @@
...
+Each Mapeo manager instance requires an adjacent Fastify server instance that is responsible for serving assets over HTTP. Each Fastify instance is responsible for handling requests for assets that can live in any Mapeo project (the URL structure reflects this, as we will show later on).
 
 Some boilerplate for getting started with a Mapeo project:
 
 ```js
+// Create Fastify instance
+const fastify = Fastify()
+
+// Create FastifyController instance for managing the starting and stopping the Fastify server (handles it more gracefully and allows pausing and restarting)
+const fastifyController = new FastifyController({ fastify })
```
gmaclennan, line 12: I think maybe add something to this about this being optional - on Desktop we would only need to call `fastify.listen()`, the use of FastifyController is only necessary for mobile where we need to start and stop the server (because Android doesn't like us running a server while the app is in the background)

src/fastify-controller.js
```diff
@@ -0,0 +1,91 @@
...
+  /**
+   * @param {StartOpts} [opts]
+   */
+  async #startServer({ host = '127.0.0.1', port = 0 } = {}) {
+    this.#host = host
+    this.#port = port
+
+    if (!this.#fastifyStarted) {
+      await this.#fastify.listen({ host: this.#host, port: this.#port })
+      this.#fastifyStarted = true
```
gmaclennan, line 45: I know this was copied across - I realise there is a bug here from my original code. This should avoid a race condition where calling `startServer()` twice before fastify.listen() resolves would throw an error.

```suggestion
      this.#fastifyStarted = true
      await this.#fastify.listen({ host: this.#host, port: this.#port })
```

src/fastify-controller.js
```diff
@@ -0,0 +1,91 @@
...
+      stop: this.#stopServer.bind(this),
+    })
+  }
+
+  /**
+   * @param {StartOpts} [opts]
+   */
+  async #startServer({ host = '127.0.0.1', port = 0 } = {}) {
+    this.#host = host
+    this.#port = port
```
gmaclennan, line 41: Nit: Again, something carried over, but I don't think these need to be instance properties - they are only used within this method.

src/fastify-plugins/utils.js
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * @param {import('node:http').Server} server
+ * @returns {Promise<string>}
+ */
+export async function getFastifyServerAddress(server) {
+  const address = server.address()
+
+  if (!address) {
+    await once(server, 'listening')
```
gmaclennan, line 11: Again, something from before, which I'm only noticing because I'm reviewing it. I think how it is fine and won't cause issues, but this change makes it a bit more "correct".

Wrapping this  function in a pTimeout leaves the `once(server, 'listening')` hanging. Better to integrate timeout into this function. I only just learnt about `AbortSignal.timeout()` but it's helpful here.

```suggestion
 * @param {import('node:http').Server} server
 * @param {{ timeout?: number }} [options]
 * @returns {Promise<string>}
 */
export async function getFastifyServerAddress(server, { timeout } = {}) {
  const address = server.address()

  if (!address) {
    await once(server, 'listening', { signal: timeout ? AbortSignal.timeout(timeout) : undefined })
```

src/mapeo-manager.js
```diff
@@ -199,6 +215,35 @@ export class MapeoManager extends TypedEmitter {
     return this.#replicate(noiseStream)
   }
 
+  /**
+   * @param {'blobs' | 'icons' | 'maps'} mediaType
+   * @returns {Promise<string>}
+   */
+  async #getMediaAddress(mediaType) {
```
gmaclennan, line 222: Nit:
```suggestion
  async #getMediaBaseUrl(mediaType) {
```

src/mapeo-manager.js
```diff
@@ -199,6 +215,35 @@ export class MapeoManager extends TypedEmitter {
...
+        break
+      }
+      default: {
+        throw new Error(`Unsupported media type ${mediaType}`)
+      }
+    }
+
+    const base = await pTimeout(getFastifyServerAddress(this.#fastify.server), {
+      milliseconds: 1000,
+    })
```
gmaclennan, line 242: if implementing the suggestion above. Also maybe 5 second timeout is safer (in case on a slow phone it takes a while for the server to start)

```suggestion
    const base = await getFastifyServerAddress(this.#fastify.server, {
      timeout: 5000,
    })
```

gmaclennan (approved):
This looks great, much neater I think than what we had before. The few things I noticed are actually things from before this PR, it's just that reviewing the code I noticed them. Mainly just nit-picks, but there is one thing that should be addressed - the setting of `isFastifyStarted` which should happen before `fastify.listen()` to avoid the race condition when the server is started twice in quick succession. Other suggestions aren't blocking.

docs/guides/mapeo-http-server.md
achou11: addressed via https://github.com/digidem/mapeo-core-next/pull/440/commits/3fb05ddf439efde39531c72407e0ea6c523a3af8

src/fastify-controller.js
achou11: addressed via https://github.com/digidem/mapeo-core-next/pull/440/commits/f80ac9a6490da9e3a2f23510d911e88f369ebf5a

src/fastify-controller.js
achou11: addressed via https://github.com/digidem/mapeo-core-next/pull/440/commits/a772a425585260f9ffe436707ea95232946dece3

src/mapeo-manager.js
achou11: addressed via https://github.com/digidem/mapeo-core-next/pull/440/commits/3aafbc01327688488971d330e4d98482ac2d93b1

src/fastify-plugins/utils.js
achou11: addressed via https://github.com/digidem/mapeo-core-next/pull/440/commits/3aafbc01327688488971d330e4d98482ac2d93b1

src/mapeo-manager.js
achou11: addressed via https://github.com/digidem/mapeo-core-next/pull/440/commits/29727df112b333a93d57d99834d0f0f178c42a75

PR #441, opened 2024-01-19
chore: add types for nanobench

EvanHahn:
Types for nanobench were [recently added to DefinitelyTyped][0], so we can use them.

The `benchmarks` directory is not currently type-checked, so this change has no impact, but it should make it easier to type-check in the future.

[0]: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/68221


achou11 (approved):
awesome!

PR #443, opened 2024-01-22
chore: prefer `net.isIPv4` over custom function

EvanHahn:
This replaces our custom `isIpv4` function with [`net.isIPv4`][0], which is built into Node.

`net.isIPv4` is stricter than ours (see [Node's source][1]), but it's the same basic thing.

[0]: https://nodejs.org/api/net.html#netisipv4input
[1]: https://github.com/nodejs/node/blob/f820efe0851dd3c64d673156b037ce7e8648fd81/lib/internal/net.js#L31-L36


gmaclennan (approved):
great I was not aware of this function! Thanks.

Issue #439, opened 2024-01-18
Implement style.json resolution strategy a when handling the stable style url


achou11:
## Description

A frontend client is provided a single endpoint to use for retrieving a style.json definition (see #438). This endpoint should be handled by the backend, which performs the following resolution strategy to determine the style.json payload that's returned as the response:

1. Check if there's a offline static map with the style id of `default` (or whatever name we decide on). In practice, this means that there's a file that exists on the filesystem at `__ROOT__/default/style.json`, where `__ROOT__` is a provided directory where static custom maps are stored. If it exists, serve the style.json content as the response

2. if internet access is available, proxy a default online style of our choosing (e.g. using a mapbox style that requires our account's public access token)

3. if internet access is not available, serve an fallback map style that uses statically served local geojson files (e.g. https://github.com/digidem/mapeo-offline-map) 

## Tasks

- [ ] Add route handler for the stable style.json url implemented in #438, which does the described resolution strategy

PR #445, opened 2024-01-23
feat: add method to MapeoManager for getting stable map style json url

### LINKED ISSUE #438: Expose a method for retrieving a stable map style.json url
## Description

Mapping libraries often require a url that points to a [map style definition](https://docs.mapbox.com/mapbox-gl-js/style-spec/), which defines how a map is rendered. In our case, we want a frontend client to be able to get a stable URL that points to the backend, which will do the work of figuring out which resource to send back (see #439).

This makes it much simpler for frontend integration, as it doesn't have to implement any logic to figure out which specific style resource to point to and would roughly look like this:

1. Retrieve the url e.g. `const styleUrl = manager.getStyleJsonUrl()`

2. Set that url for the map e.g. `new mapboxgl.Map({ container: 'map' style: styleUrl })` (plain JS) or `<MapView styleUrl={styleUrl}>...` (React)

## Tasks

- [ ] Implement a method on the `MapeoManager` for retrieving a stable style.json url.

achou11:
Closes #438 

~Currently stacked on #440~

Notes:

- this PR does not actually implement the endpoint that handles the returned url. that will be addressed in #439

### DIFF
--- a/src/mapeo-manager.js
+++ b/src/mapeo-manager.js
@@ -56,6 +56,7 @@ const MAX_FILE_DESCRIPTORS = 768
 // Prefix names for routes registered with http server
 const BLOBS_PREFIX = 'blobs'
 const ICONS_PREFIX = 'icons'
+const MAPS_PREFIX = 'maps'
 
 export const kRPC = Symbol('rpc')
 export const kManagerReplicate = Symbol('replicate manager')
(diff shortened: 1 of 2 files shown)


achou11:
Closing in favor of #457

PR #462, opened 2024-01-31
feat!: rename "Capability" to "Role", attach ID

EvanHahn:
This change makes several improvements to the concept formerly known as "capabilities":

- `Capability` is now called `Role` across the project, resulting in many simple renames.

- `Role`s have an attached `roleId` property, which they didn't before.

- The creator role and "no role" role now have role IDs.[^1]

- The `RoleId` type now includes all possible role types, now including the creator role and the "no role" role. I created narrower types such as `RoleIdAssignableToOthers` and `RoleIdAssignableToAnyone`, which don't include those.

This should help the front-end know what role someone is (for example, <https://github.com/digidem/CoMapeo-mobile/pull/79#discussion_r1470494854>).

[^1]: I generated these with `crypto.randomBytes(8).toString('hex')`.

### DIFF
--- a/src/member-api.js
+++ b/src/member-api.js
@@ -1,15 +1,15 @@
 import { TypedEmitter } from 'tiny-typed-emitter'
 import { InviteResponse_Decision } from './generated/rpc.js'
 import { projectKeyToId } from './utils.js'
-import { DEFAULT_CAPABILITIES } from './capabilities.js'
+import { ROLES } from './roles.js'
 
 /** @typedef {import('./datatype/index.js').DataType<import('./datastore/index.js').DataStore<'config'>, typeof import('./schema/project.js').deviceInfoTable, "deviceInfo", import('@mapeo/schema').DeviceInfo, import('@mapeo/schema').DeviceInfoValue>} DeviceInfoDataType */
 /** @typedef {import('./datatype/index.js').DataType<import('./datastore/index.js').DataStore<'config'>, typeof import('./schema/client.js').projectSettingsTable, "projectSettings", import('@mapeo/schema').ProjectSettings, import('@mapeo/schema').ProjectSettingsValue>} ProjectDataType */
-/** @typedef {{ deviceId: string, name?: import('@mapeo/schema').DeviceInfo['name'], capabilities: import('./capabilities.js').Capability }} MemberInfo */
+/** @typedef {{ deviceId: string, name?: import('@mapeo/schema').DeviceInfo['name'], role: import('./roles.js').Role }} MemberInfo */
 
 export class MemberApi extends TypedEmitter {
   #ownDeviceId
-  #capabilities
+  #roles
   #coreOwnership
   #encryptionKeys
   #projectKey
@@ -19,7 +19,7 @@ export class MemberApi extends TypedEmitter {
   /**
    * @param {Object} opts
    * @param {string} opts.deviceId public key of this device as hex string
-   * @param {import('./capabilities.js').Capabilities} opts.capabilities
+   * @param {import('./roles.js').Roles} opts.roles
    * @param {import('./core-ownership.js').CoreOwnership} opts.coreOwnership
    * @param {import('./generated/keys.js').EncryptionKeys} opts.encryptionKeys
    * @param {Buffer} opts.projectKey
@@ -30,7 +30,7 @@ export class MemberApi extends TypedEmitter {
    */
   constructor({
     deviceId,
-    capabilities,
+    roles,
     coreOwnership,
     encryptionKeys,
     projectKey,
@@ -39,7 +39,7 @@ export class MemberApi extends TypedEmitter {
   }) {
     super()
     this.#ownDeviceId = deviceId
-    this.#capabilities = capabilities
+    this.#roles = roles
     this.#coreOwnership = coreOwnership
     this.#encryptionKeys = encryptionKeys
     this.#projectKey = projectKey
@@ -51,7 +51,7 @@ export class MemberApi extends TypedEmitter {
    * @param {string} deviceId
    *
    * @param {Object} opts
-   * @param {import('./capabilities.js').RoleId} opts.roleId
+   * @param {import('./roles.js').RoleIdAssignableToOthers} opts.roleId
    * @param {string} [opts.roleName]
    * @param {string} [opts.roleDescription]
    * @param {number} [opts.timeout]
@@ -59,7 +59,7 @@ export class MemberApi extends TypedEmitter {
    * @returns {Promise<import('./generated/rpc.js').InviteResponse_Decision>}
    */
   async invite(deviceId, { roleId, roleName, roleDescription, timeout }) {
-    if (!DEFAULT_CAPABILITIES[roleId]) {
+    if (!ROLES[roleId]) {
       throw new Error('Invalid role id')
     }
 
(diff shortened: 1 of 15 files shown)


src/roles.js
```diff
@@ -344,8 +377,8 @@ export class Capabilities {
 /**
  *
  * @param {string} roleId
- * @returns {roleId is keyof DEFAULT_CAPABILITIES}
+ * @returns {roleId is keyof ROLES}
  */
 function isKnownRoleId(roleId) {
-  return roleId in DEFAULT_CAPABILITIES
+  return roleId in ROLES
```
EvanHahn, line 383: Highlighting this: `DEFAULT_CAPABILITIES` didn't have the creator or "no role" capabilities before. `ROLES` _does_ have those. I _think_ this is fine based on usage but want to make sure.

src/types.ts
```diff
@@ -104,11 +104,6 @@ export type KeyPair = {
   publicKey: PublicKey
   secretKey: SecretKey
 }
-export type RoleDetails = {
-  name: string
-  capabilities: string[]
-}
-export type AvailableRoles = RoleDetails[]
```
EvanHahn, line 111: These seemed completely unused (even [across the org][0]). I removed them rather than renaming them.

[0]: https://github.com/search?q=org%3Adigidem+availableroles&type=code

src/roles.js
gmaclennan: I think we should avoid allowing the creator role id, maybe not in this function but instead where it is used on line 288? We will later validate role assignments, but I think it's safer that we don't allow a role record to define a role to be the project creator, since we're not allowing the project creator role to be assigned, just implicit.

src/types.ts
gmaclennan: Yes I think this is from older code, fine to remove.

src/member-api.js
```diff
@@ -51,15 +51,15 @@ export class MemberApi extends TypedEmitter {
...
+   * @param {import('./roles.js').RoleIdAssignableToOthers} opts.roleId
    * @param {string} [opts.roleName]
    * @param {string} [opts.roleDescription]
    * @param {number} [opts.timeout]
    *
    * @returns {Promise<import('./generated/rpc.js').InviteResponse_Decision>}
    */
   async invite(deviceId, { roleId, roleName, roleDescription, timeout }) {
-    if (!DEFAULT_CAPABILITIES[roleId]) {
+    if (!ROLES[roleId]) {
```
gmaclennan, line 62: I think this needs to be stricter now, since we do not want to allow inviting someone as a project creator. This should be the "assignable to others" role list, although we should probably exclude the blocked role from that, but maybe as a to-do because I think some tests invite as blocked, just because I was lazy.

src/roles.js
```diff
@@ -17,25 +19,51 @@ export const LEFT_ROLE_ID = '8ced989b1904606b'
...
+ */
+
+/**
+ * @typedef {Extract<RoleId,
+ *   typeof COORDINATOR_ROLE_ID |
+ *   typeof MEMBER_ROLE_ID |
+ *   typeof BLOCKED_ROLE_ID |
+ *   typeof LEFT_ROLE_ID
+ * >} RoleIdAssignableToAnyone
+ */
```
gmaclennan, line 56: I think better to have both of these as runtime checks to e.g.

```js
const ROLE_ID_ASSIGNABLE_TO_ANYONE = /** @type {const} */ ([
  COORDINATOR_ROLE_ID,
  MEMBER_ROLE_ID,
  BLOCKED_ROLE_ID,
  LEFT_ROLE_ID,
])
/** @typedef {typeof ROLE_ID_ASSIGNABLE_TO_ANYONE} RoleIdAssignableToAnyone */
```

and when used as a type we should also add an assertion.

src/member-api.js
```diff
@@ -163,10 +163,10 @@ export class MemberApi extends TypedEmitter {
 
   /**
    * @param {string} deviceId
-   * @param {import('./capabilities.js').RoleId} roleId
+   * @param {import('./roles.js').RoleIdAssignableToOthers} roleId
    * @returns {Promise<void>}
    */
   async assignRole(deviceId, roleId) {
```
gmaclennan, line 169: Can we add a TODO note here or open an issue to have stricter runtime checks here.
I also wonder if "assigning a blocked role" is confusing for the API consumer, and we should expose that separately as `$member.block(deviceId)` and limit `assignRole()` to just assignable roles within the project (currently participant and coordinator).

gmaclennan: Also in the future we might want to add options to how we block users (which internally we might implement as different roles, or additional fields on membership records - maybe the latter better for many scenarios), for example:

1. Removed with "self-destruct", e.g. when a device discovers it is removed it deletes all data including its own. Would be useful for compromised phones.
2. Removed with removed encryption keys and keeping own data (but deleting others) - useful for when removing a device that we know has important data on it, that we want to sync off it in the future, without allowing that device to access it. Device would need to self-delete its encryption keys when it finds out it is blocked this way.
3. Removed with ignore all data from device - would re-index and ignore all edits or created data from the removed device.

Having a separate API to remove devices would be helpful for implementing these.

src/roles.js
```diff
@@ -17,25 +19,51 @@ export const LEFT_ROLE_ID = '8ced989b1904606b'
  */
 
 /**
- * @typedef {object} Capability
+ * @typedef {object} Role
+ * @property {RoleId} roleId
```
gmaclennan, line 23:
```suggestion
 * @template {RoleId} [T=RoleId]
 * @typedef {object} Role
 * @property {T} roleId
```

See below

src/roles.js
```diff
@@ -81,9 +110,11 @@ export const NO_ROLE_CAPABILITIES = {
   },
 }
 
-/** @type {Record<RoleId, Capability>} */
-export const DEFAULT_CAPABILITIES = {
+/** @type {Record<RoleId, Role>} */
```
gmaclennan, line 113:
```suggestion
/** @type {{ [K in RoleId]: Role<K> }} */
```

To give us a bit more type safety to avoid silly bugs in the future. Requires CREATOR_ROLE to have type `Role<typeof CREATOR_ROLE_ID>` and similarly for NO_ROLE.

gmaclennan (approved):
Looks good. Happy for my comments to be addressed in a follow-up so that this can get merged sooner, since it touches quite a few files and I don't want to create complex merge operations in the future.

src/roles.js
EvanHahn: This is great; done in b8742e47508cdd69d24aeea067b38c7ee8ac3a91. Had to add 885071e93f21a8316e6d0ec17dc34d55ef3024ef as a followup to fix a type error.

src/roles.js
EvanHahn: Done in b8742e47508cdd69d24aeea067b38c7ee8ac3a91.

src/roles.js
EvanHahn: Done in f24c85e3d807ce58e3f4b99a148f6ee1c9d345b6.

src/member-api.js
EvanHahn: Done in f24c85e3d807ce58e3f4b99a148f6ee1c9d345b6.

src/roles.js
EvanHahn: Done in f24c85e3d807ce58e3f4b99a148f6ee1c9d345b6.

src/member-api.js
EvanHahn: Added a stricter runtime check in f24c85e3d807ce58e3f4b99a148f6ee1c9d345b6. Let me know if you still want me to add a TODO/issue.

EvanHahn:
I believe I've addressed all comments. I've requested another review but will plan to merge tomorrow barring objections.

PR #469, opened 2024-02-06
feat: delete lots of data from indexers when leaving project

### LINKED ISSUE #427: Delete indexed data when leaving a project
## Description

Follow-up item based on discussions around https://github.com/digidem/mapeo-core-next/pull/410. Implementing it requires updating the multi-core and sqlite indexers first, and then updating the `IndexWriter` and `MapeoProject.leave()` to take advantage of the changes.

## Tasks

- [ ] Update [multi-core-indexer](https://github.com/digidem/multi-core-indexer/issues/26)
- [ ] Update [mapeo-sqlite-indexer](https://github.com/digidem/mapeo-sqlite-indexer/issues/22)
- [ ] Add a method to `IndexWriter` to call the relevant method for each sqlite indexer instance
- [ ] Update `MapeoProject.leave()` implementation to use the new methods. Also have to keep in mind that the order in which they're called is important:

  1. Delete core data (already implemented)
  2. Reset multicore indexer state
  3. Delete data from index tables

EvanHahn:
This is a draft because it depends on vendored versions of `multi-core-indexer` and `@mapeo/sqlite-indexer`, but it's otherwise ready to go. Once this is approved, I think we should release new versions of those libraries.

When you leave a project, we now delete all indexed data outside of the `auth` namespace. Fixes #427.

Co-Authored-By: Andrew Chou <andrewchou@fastmail.com>


src/capabilities.js
```diff
@@ -299,13 +299,13 @@ export class Capabilities {
...
       )
     }
-    const ownCapabilities = await this.getCapabilities(this.#ownDeviceId)
 
     if (roleId === LEFT_ROLE_ID) {
       if (deviceId !== this.#ownDeviceId) {
         throw new Error('Cannot assign LEFT role to another device')
       }
     } else {
+      const ownCapabilities = await this.getCapabilities(this.#ownDeviceId)
```
EvanHahn, line 308: While I was developing, I ran into issues with this being in its original location. I think this is a better spot for it anyway, so I kept it.

src/index-writer/index.js
```diff
@@ -52,8 +57,11 @@ export class IndexWriter {
     }
   }
 
+  /**
+   * @returns {Iterable<SchemaName>}
```
EvanHahn, line 61: I changed the contract for this, but I don't think it was ever used previously so I think that's fine.

src/utils.js
```diff
@@ -66,6 +66,28 @@ export async function openedNoiseSecretStream(stream) {
   return /** @type {OpenedNoiseStream | DestroyedNoiseStream} */ (stream)
 }
 
+/**
```
EvanHahn, line 69: This is lifted from #462, along with its tests.

src/mapeo-project.js
```diff
@@ -626,9 +628,28 @@ export class MapeoProject extends TypedEmitter {
...
+          await dataStore.unlink()
+        })
+    )
+
     // 3.2 Clear indexed data
+    const isIndexedSchema = setHas(new Set(this.#indexWriter.schemas))
+    await Promise.all(
+      Object.values(this.#dataTypes)
+        .filter((dataType) =>
+          namespacesWithoutAuth.includes(dataType.namespace)
```
gmaclennan, line 647: nitpick, but this being different from L#635 bothers my OCD!

```suggestion
          dataType.namespace !== 'auth'
```

src/mapeo-project.js
```diff
@@ -626,9 +628,28 @@ export class MapeoProject extends TypedEmitter {
...
+        )
+        .map((dataType) => dataType.schemaName)
+        .filter(isIndexedSchema)
+        .map((schemaName) => this.#indexWriter.deleteSchema(schemaName))
+    )
 
     // 4. Assign LEFT role for device
     await this.#roles.assignRole(this.#deviceId, LEFT_ROLE_ID)
```
gmaclennan, line 655: I don't think this should work after the core is closed (line 625) and I don't think it should resolve after the indexer is closed.

src/mapeo-project.js
```diff
@@ -626,9 +628,28 @@ export class MapeoProject extends TypedEmitter {
...
+        .map(async (dataStore) => {
+          await dataStore.close()
+          await dataStore.unlink()
+        })
+    )
+
     // 3.2 Clear indexed data
+    const isIndexedSchema = setHas(new Set(this.#indexWriter.schemas))
+    await Promise.all(
+      Object.values(this.#dataTypes)
```
gmaclennan, line 645: This feels like the wrong way to access what needs to be deleted. It requires an unspoken contract of "any schema you add to the indexer must be a dataType in the this.#dataType map". I'm worried this is an easy route to a bug by a future maintainer. How about iterating the schemas from the index writer, and using the map `NAMESPACE_SCHEMAS` from DataStore to figure out which are non-auth store schemas?

gmaclennan (commented):
Reading through this and related code I have some thoughts:

- We need to close cores before we unlink/purge them - this should cancel replication requests and closes file descriptors
- We should probably unreplicate cores that we are deleting before deleting them
- We should write the updated role for leaving a project before deleting anything
- We should probably wait for (pre)sync to complete before deleting a project - ensuring connected peers get all the data from the departing device, including the role record that they have left.
- I'm not entirely clear about the state of a project instance after leaving - not everything is closed, but I'm not sure if the instance will work any more, or just have undefined behaviour?
- I'm not sure about deleting schemas from the indexer separately from deleting the multi-core-indexer state - what happens if we don't delete a schema in a namespace but we do delete the multi-core-indexer?
- I'm worried we're introducing complicated states where the indexes could be out-of-sync with the underlying cores. Would it be safer to just delete all indexes and index reader states and then re-index the cores we have left?
- I think we need to maintain some kind of left project state, so that when a new project instance is created for a project that has been left, it doesn't try to re-sync everything and re-create the writer cores.

Sorry for inconclusive review - this PR raises a lot of questions / highlights some potential issues.

src/datastore/index.js
```diff
@@ -229,6 +229,14 @@ export class DataStore extends TypedEmitter {
   async close() {
     await this.#coreIndexer.close()
   }
+
+  /**
+   * Unlink all index files. This should only be called after `close()` has resolved.
```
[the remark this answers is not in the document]
EvanHahn: Done in https://github.com/digidem/multi-core-indexer/pull/42.

EvanHahn:
> We need to close cores before we unlink/purge them - this should cancel replication requests and closes file descriptors

I think we're doing that here:

https://github.com/digidem/mapeo-core-next/blob/cdebfafa772958729f5ade5e64c04010b160b710/src/mapeo-project.js#L624-L629

> We should probably unreplicate cores that we are deleting before deleting them

I'm not super familiar with its internals but [it looks like Hypercore might do this for us](https://github.com/holepunchto/hypercore/blob/16375ac22e87907b212c03d0e4a8e6b0a38f8c76/index.js#L440-L442)?

> We should write the updated role for leaving a project before deleting anything

Done in ab4d9c1194a1b286ce52d85a9f25f988857cb8c2.

> I'm not sure about deleting schemas from the indexer separately from deleting the multi-core-indexer state - what happens if we don't delete a schema in a namespace but we do delete the multi-core-indexer?

At worst, we might've left a project but not cleaned up one or both of the indexers due to a crash or other failure.

We could add something to check this at startup. Pseudocode:

```javascript
onStartup(() => {
  for (const leftProject in getLeftProjects()) {
    leftProject.makeSureCoresAndIndexesAreDestroyed()
  }
})
```

> I'm worried we're introducing complicated states where the indexes could be out-of-sync with the underlying cores. Would it be safer to just delete all indexes and index reader states and then re-index the cores we have left?

I defer to your expertise here but I *think* we're just clearing the indexes and closing/deleting the cores. I suppose cores could have data in them that hasn't made it to the indexes and never will, but I think that's fine?

This feels like one of a few questions we haven't answered about leaving projects. Others include:

> - We should probably wait for (pre)sync to complete before deleting a project - ensuring connected peers get all the data from the departing device, including the role record that they have left.
> - I'm not entirely clear about the state of a project instance after leaving - not everything is closed, but I'm not sure if the instance will work any more, or just have undefined behaviour?
> - I think we need to maintain some kind of left project state, so that when a new project instance is created for a project that has been left, it doesn't try to re-sync everything and re-create the writer cores.

I think these are separate tasks, but I agree.

src/mapeo-project.js
EvanHahn: Moved in ab4d9c1194a1b286ce52d85a9f25f988857cb8c2.

src/mapeo-project.js
EvanHahn: Done in 0b6aec3d3a9e726d6d5422d763dd73313589f20f.

src/mapeo-project.js
EvanHahn: Removed in 0b6aec3d3a9e726d6d5422d763dd73313589f20f.

EvanHahn:
I'm going to file the following issues if that seems reasonable to you:

- **At startup, clean up left projects that failed to delete**

  If we start to leave a project and the app crashes, we may fail to remove data from indexers. At startup, we should find left projects and remove the "orphaned" data.

- **Test leaving a project while offline, then connecting and syncing**

  We don't test what happens when someone leaves a project while offline, then syncs that to others. This test may uncover bugs.

- **Test project instance after leaving**

  We don't test what happens when someone leaves a project, then tries to perform certain operations on it. This test may uncover bugs or missing features.

Do those three tasks seem reasonable?

EvanHahn:
Just filed #583,#584, and #585 based on the above.

gmaclennan (approved):
This seems ok to me, although I admit I can't quite get my head around possible edge-cases, but what is here looks like it should work. It could be worth spending some time writing out some more detailed tests of how we expect devices that have left a project to behave when connecting to others. I think the other aims of project leaving are:
- Remove others data to free up disk space - can we test for this?
- Keep own data so that it can be recovered later - can we test for this?

EvanHahn:
> - Remove others data to free up disk space - can we test for this?
> - Keep own data so that it can be recovered later - can we test for this?

Good idea. I'll plan to do this in #585.

PR #471, opened 2024-02-07
feat: load default config when creating a project

### LINKED ISSUE #402: Pre-populate the database with default config when a project is first created

tomasciccola:
this should close #402


tomasciccola:
Working on this I found that I had an error on the default config that is on the project. So - despite being out of scope - I'll probably add a test for the default config...

src/mapeo-manager.js
```diff
@@ -381,6 +383,13 @@ export class MapeoManager extends TypedEmitter {
     // TODO: Close the project instance instead of keeping it around
     this.#activeProjects.set(projectPublicId, project)
 
+    // load default config
+    // TODO: see how to expose warnings to frontend
+    /* eslint-disable no-unused-vars */
+    const warnings = await project.importConfig({
+      configPath: path.resolve(DEFAULT_CONFIG_PATH),
```
EvanHahn, line 390: Nit: should this resolve relative to `__dirname`?

EvanHahn (approved):
LGTM!

src/mapeo-manager.js
EvanHahn: Oh yeah, that's better. I'd love that!

EvanHahn (approved):
LGTM assuming CI passes.

tomasciccola:
> LGTM assuming CI passes.

Yeah, I see that E2E tests are failing on project-crud.js. Gotta see whats that all about but its hard to pinpoint since the output is kinda huge ...

tomasciccola:
So, tests are failing on `test-e2e/manager-basic.js` because expectations about how a project looks after creation changed, basically because when we create a project we implicitly load the default config. So now theres a bunch of assertions like so:
```js
st.alike(project, {name: 'project1', defaultPresets: undefined})
```
which fails because `defaultPresets` is not undefined anymore.
One option is to change the assertion to:
```js
st.is(project.name, 'project1')
```
Another option would not to run `importConfig` when creating a project implicitly.
A third option would be change the assertion to contemplate the loaded presets
@gmaclennan @EvanHahn  wdyt??

gmaclennan:
I think this test is the best place to check that the config has been imported when a project is created. I suggest:

- read the config in the test
- check defaults are correct after creating the project
- check fields and presets in the project match what is in the config file
- check icons are there too.

tomasciccola:
> I think this test is the best place to check that the config has been imported when a project is created. I suggest:
> 
>     * read the config in the test
> 
>     * check defaults are correct after creating the project
> 
>     * check fields and presets in the project match what is in the config file
> 
>     * check icons are there too.

the config is already read when doing `manager.createProject`. Do you think we should also import and call `readConfig` with the path to the default config and compare it to the config loaded when creating the project??

gmaclennan:
> the config is already read when doing `manager.createProject`. Do you think we should also import and call `readConfig` with the path to the default config and compare it to the config loaded when creating the project??
> I think calling `readConfig` in the tests and then comparing with `project.{defaultSettings,presets,icons,fields}` makes sense since, if not, we're basically repeating the tests of `importConfig`

As far as I can see, we don't yet have any tests for `project.importConfig()`. What we have tests for are unit tests for `readConfig`. I think we need two e2e tests:

1. If you create a new project, then `const { defaults } = await project.getProjectSettings()` and `presets.getMany()` and `fields.getMany()` should all return the expected data in the default config - for which you need to read the config in the test to get the expected data (which is ok because you are unit testing readConfig to more-or-less make sure it's working correctly).
2. If you `project.importConfig(newConfig)` then you need to test the same things, but updated with the new config data.

Should probably also check icons are correctly there in both scenarios too. A test helper function to get `expectedConfig` might be helpful, to DRY iterating through the config reader, at the expense of memory use, which is ok in tests I think.

test-e2e/manager-basic.js
```diff
@@ -56,19 +58,86 @@ test('Managing created projects', async (t) => {
...
+
+      const projectPresets = await project1.preset.getMany()
+
+      st.is(
+        settings1.defaultPresets?.point.length,
+        defaultPresets.length,
+        'the default presets loaded is equal to the number of presets in the default config'
+      )
+
+      st.alike(
```
tomasciccola, line 86: this asssertion - and the following in this group - are failing. From what I can seem it seems that calling `project1.preset.getMany()` returns the presets corresponding to the config that I load later, which is weird.

tomasciccola: It seems that is related to assertions running in an unexpected order

tomasciccola:
from what I can gather looking at the logs, it seems that there are a bunch of tests failing given that now there are a bunch new records loaded from the default config. 
So working on that is gonna take a while...
@gmaclennan does it makes sense to fix those tests here?

EvanHahn:
Sorry I'm just now reviewing this!

> does it makes sense to fix those tests here, or make another PR that fixes this directly?

I think we should fix the tests in this PR. I like Gregor's idea of a test helper function.

Does that seem reasonable?

tomasciccola:
After a short sync with gregor, we decided to add a `defaultConfigPath` opt to `mapeo-manager`.  If no path is passed, we skip loading the config. This allows us to decide if a default config should be loaded and where should it live, which removes overhead from the tests.

src/mapeo-manager.js
```diff
@@ -389,6 +393,16 @@ export class MapeoManager extends TypedEmitter {
     // TODO: Close the project instance instead of keeping it around
     this.#activeProjects.set(projectPublicId, project)
 
+    // load default config
+    // TODO: see how to expose warnings to frontend
+    /* eslint-disable no-unused-vars */
+    let warnings
+    if (this.#defaultConfigPath) {
+      warnings = await project.importConfig({
+        configPath: new URL(this.#defaultConfigPath, import.meta.url).pathname,
```
gmaclennan, line 402: I think `defaultConfigPath` should be an absolute path, no need to support relative paths here.

src/mapeo-project.js
```diff
@@ -721,8 +721,8 @@ function extractEditableProjectSettings(projectDoc) {
 /** @param {DataType<any,any,any,any,any>} dataType */
 async function deleteAll(dataType) {
   const deletions = []
-  for (const { versionId } of await dataType.getMany()) {
-    deletions.push(dataType.delete(versionId))
+  for (const { docId } of await dataType.getMany()) {
```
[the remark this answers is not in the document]
gmaclennan: This suggests we were lacking test coverage here. Can you create an issue to create an e2e test for changing config, to ensure that old config is correctly deleted.

test-e2e/manager-basic.js
```diff
@@ -56,20 +59,90 @@ test('Managing created projects', async (t) => {
...
+    async (st) => {
+      const settings1 = await project1.$getProjectSettings()
+      const settings2 = await project2.$getProjectSettings()
+
+      const {
+        presets: defaultPresets,
+        fields: defaultFields,
+        icons: defaultIcons,
+      } = await getExpectedConfig(
+        new URL('../config/defaultConfig.mapeoconfig', import.meta.url).pathname
```
gmaclennan, line 73: define this once maybe, at top of the test.

test-e2e/manager-basic.js
```diff
@@ -56,20 +59,90 @@ test('Managing created projects', async (t) => {
...
+    'initial settings and default config from project instances',
+    async (st) => {
+      const settings1 = await project1.$getProjectSettings()
+      const settings2 = await project2.$getProjectSettings()
+
+      const {
+        presets: defaultPresets,
+        fields: defaultFields,
+        icons: defaultIcons,
+      } = await getExpectedConfig(
```
gmaclennan, line 72: Naming suggestion: call this `const expected = await getExpectedConfig()`
Then in the asserts below you can do `t.alike(settings1.defaultpresets.point, expected.presets)` etc.

test-e2e/manager-basic.js
```diff
@@ -56,20 +59,90 @@ test('Managing created projects', async (t) => {
...
+        new URL('../config/defaultConfig.mapeoconfig', import.meta.url).pathname
+      )
+
+      st.is(settings1.name, undefined, 'undefined name for project1')
+      st.is(settings2.name, 'project 2', 'matched name for project2')
+
+      const projectPresets = await project1.preset.getMany()
+
+      st.is(
+        settings1.defaultPresets?.point.length,
```
gmaclennan, line 82: Better to `@ts-ignore` here and let it throw, rather than using `?`, since it could obscure a bug in the test.

test-e2e/manager-basic.js
```diff
@@ -56,20 +59,90 @@ test('Managing created projects', async (t) => {
...
+        defaultPresets.length,
+        'the default presets loaded is equal to the number of presets in the default config'
+      )
+
+      st.alike(
+        projectPresets.map((preset) => preset.name),
+        defaultPresets.map((preset) => preset.value.name),
```
gmaclennan, line 89: sort order is not guaranteed, so best to compare sets e.g. `new Set([...projectPresets.map((preset) => preset.name)])`.
Also: I think you have these the wrong way around - `projectPresets` is what has `.value` properties.
Also: test the whole thing e.g. `st.alike(new Set([...projectPresets.map(p => p.value)]), new Set([...expected.presets]))`

test-e2e/manager-basic.js
```diff
@@ -56,20 +59,90 @@ test('Managing created projects', async (t) => {
...
+        'project is loading the default presets correctly'
+      )
+
+      const projectFields = await project1.field.getMany()
+
+      st.alike(
+        projectFields.map((field) => field.tagKey),
+        defaultFields.map((field) => field.value.tagKey),
+        'project is loading the default fields correctly'
+      )
```
gmaclennan, line 99: These are the wrong way around I think.

test-e2e/manager-basic.js
```diff
@@ -56,20 +59,90 @@ test('Managing created projects', async (t) => {
...
+  )
 
-    st.alike(settings2, {
-      name: 'project 2',
-      defaultPresets: undefined,
-    })
-  })
+  await t.test(
+    'load different config and check if correctly loaded',
+    async (st) => {
```
gmaclennan, line 112: I think things can get confusing when you have subtests that mutate the project, maybe move to a separate test and a separate project instances? I don't think this needs to be part of the same test.

test-e2e/manager-basic.js
```diff
@@ -56,20 +59,90 @@ test('Managing created projects', async (t) => {
...
+      st.alike(
+        projectPresets.map((preset) => preset.name),
+        loadedPresets.map((preset) => preset.value.name),
+        'project presets explicitly loaded match expected config'
+      )
+
+      const projectFields = await project1.field.getMany()
+      st.alike(
+        projectFields.map((field) => field.tagKey),
+        loadedFields.map((field) => field.value.tagKey),
```
gmaclennan, line 134: As above, aren't these the wrong way around?

test-e2e/manager-basic.js
```diff
@@ -56,20 +59,90 @@ test('Managing created projects', async (t) => {
...
+
+      const projectFields = await project1.field.getMany()
+
+      st.alike(
+        projectFields.map((field) => field.tagKey),
+        defaultFields.map((field) => field.value.tagKey),
+        'project is loading the default fields correctly'
+      )
+
+      const projectIcons = await project1[kDataTypes].icon.getMany()
```
gmaclennan, line 101: maybe read icons from presets?

test-e2e/manager-basic.js
```diff
@@ -56,20 +59,90 @@ test('Managing created projects', async (t) => {
...
+      // TODO: since we don't delete icons, this wouldn't match
+      // const projectIcons = await project1[kDataTypes].icon.getMany()
+      // st.alike(
+      //   projectIcons.map((icon) => icon.name),
+      //   loadedIcons.map((icon) => icon.name)
+      // )
+    }
+  )
 
   t.test('after updating project settings', async (st) => {
```
gmaclennan, line 147: needs await

tests/config-import.js
```diff
@@ -208,3 +208,38 @@ test('config import - presets', async (t) => {
...
+    'valid config file'
+  )
+
+  let config = await readConfig('./config/defaultConfig.mapeoconfig')
+  let nFields = 0
+  /* eslint-disable-next-line */
+  for (const field of config.fields()) {
+    nFields++
+  }
+  t.is(nFields, 2, 'correct number of fields in default config')
```
gmaclennan, line 224: This doesn't seem right

tests/config-import.js
```diff
@@ -208,3 +208,38 @@ test('config import - presets', async (t) => {
...
   t.is(config.warnings.length, 0, `no warnings on the file`)
 })
+
+test('config import - load default config', async (t) => {
+  t.ok(
+    await readConfig('./config/defaultConfig.mapeoconfig'),
+    'valid config file'
+  )
+
+  let config = await readConfig('./config/defaultConfig.mapeoconfig')
```
gmaclennan, line 218: why read twice?

gmaclennan (changes requested):
I think there are bugs in the tests that need resolved.

gmaclennan:
Also lots the subtests in the manager-basic are missing `await` - code was created prior to this PR, but the fact they are missing means subtests are running out-of-order.

tests/config-import.js
gmaclennan: No that would be it, ok to resolve in a follow up

test-e2e/manager-basic.js
tomasciccola: How would I do that?
I mean, I can do `project1.presets.getMany()` and from that get all the `iconId`s, but that wouldn't be of much help since the `expected.icons` are not records on the db, with a `docId` that I can match

gmaclennan (changes requested):
Looks good. However just now checked out the code on my laptop and could see the whole tests in context - it doesn't seem to make sense to add the default config and config import tests as subtests to the "Managing created projects" test. That test deals with testing project settings with two separate projects. The additional tests for config seem completely different, and I think should be a separate test.

Other than that, I think it's ok to merge and address issues in a follow-up. Ideally this would be an end-to-end test without delving into implementation details for icons, but I think fine to add that later, since it's a bit complicated to do (I think you would need to iterate through actual presets in the project, and match them with expected presets, then look up which icon variants to expect, then attempt to getUrl for each, and read the returned blob, then compare with expected data).

Also ideally we would test complete values e.g.

```
      st.alike(
        new Set([...projectFields.map((field) => stripUndef(valueOf(field)))]),
        new Set([...expectedDefault.fields.map((field) => field.value)]),
        'project is loading the default fields correctly'
      )
```

Also putting them in a Set makes the test more robust for order changes. However doing this for presets also involves removing/ignoring fieldIds and iconId, so again just a bit too much work and not a priority for getting the MVP out the door.

So only change requested is just moving the config tests to a separate test, not subtests as the currently are.

tomasciccola:
> Looks good. However just now checked out the code on my laptop and could see the whole tests in context - it doesn't seem to make sense to add the default config and config import tests as subtests to the "Managing created projects" test. That test deals with testing project settings with two separate projects. The additional tests for config seem completely different, and I think should be a separate test.
> 
> Other than that, I think it's ok to merge and address issues in a follow-up. Ideally this would be an end-to-end test without delving into implementation details for icons, but I think fine to add that later, since it's a bit complicated to do (I think you would need to iterate through actual presets in the project, and match them with expected presets, then look up which icon variants to expect, then attempt to getUrl for each, and read the returned blob, then compare with expected data).
> 
> Also ideally we would test complete values e.g.
> 
> ```
>       st.alike(
>         new Set([...projectFields.map((field) => stripUndef(valueOf(field)))]),
>         new Set([...expectedDefault.fields.map((field) => field.value)]),
>         'project is loading the default fields correctly'
>       )
> ```
> 
> Also putting them in a Set makes the test more robust for order changes. However doing this for presets also involves removing/ignoring fieldIds and iconId, so again just a bit too much work and not a priority for getting the MVP out the door.
> 
> So only change requested is just moving the config tests to a separate test, not subtests as the currently are.

Yeah, I agree. I also wanted to test whole values - instead of just the name - but it seemed a little to much work. 
I addressed moving the subtests to their separate test, so I think this should be ready to merge...

gmaclennan (approved):

Issue #466, opened 2024-02-06
Translation API


tomasciccola:
## Description

Translation API that wraps the dataType instance. The only big difference to note is that when creating a new doc of type translation, the `docId` is hashed from `recordType, recordId, fieldRef, languageCode`

PR #473, opened 2024-02-12
feat: Add `lang` option to DataType read operations

### LINKED ISSUE #472: Add `lang` option to DataType read operations
## Description

We need a way to translate fields of Mapeo records to the device language, e.g. Preset and Field record types could have multi-lingual names, labels, hints etc.

The proposed [Translations API #466](https://github.com/digidem/mapeo-core-next/issues/466) allows for storage of translated fields. This issue is how to translate fields when reading records.

Option A: Translate in the front-end. When reading a record, for each field, use the translation API to lookup a translation.
Option B: Translate in the back-end. Front end can pass a `lang` option when reading a record, and the backend will return the record with translated fields.

I think Option B is far preferable, because Option A adds a lot of complexity to front-end code, and will also have a performance impact because of multiple requests to the backend. Language matching is also complicated, so Option B keeps this logic in the backend, and also allows for optimization by caching which records and fields have translations, to avoid unnecessary lookups.

## Implementation idea

Add an extra paramater `opts.lang` to the DataType read methods (`dataType.getByDocId()`, `dataType.getByVersionId` and `dataType.getMany()`), which is a IETF BCP-47 language string (using `opts.lang` to mirror the usage of the `lang` tag in HTML).

In the Translation API, keep an in-memory cache of which record types and which field refs have been parsed, as a performance optimization, e.g. `$translation.translations: Map<MapeoDoc['schemaName'], Set<FieldRef>>`. This should be ok to be in-memory because it is limited to the number of record types x number of text fields per record, and will likely be much smaller.

Within the DataType read methods, because this is potentially an expensive operation, first check whether there are any indexed translations for the record type (`$translation.translations.has(recordType)`), then for each field in the read data, check if there are any translations indexed for that field ref (`$translation.translations.get(recordType).has(fieldRef)`).

If a translation might exist, then we look it up:

1. Parse `opts.lang` using https://github.com/wooorm/bcp-47 (can be memoized, because this will likely always be the same)
2. Map the primary language subtag to ISO 639-3 equivalent (it could be ISO 639-1 or (unlikely) ISO 639-2)
3. Try to read the translation `$translation.get({ languageCode, recordType, recordId, fieldRef })` (return array)
4. If more than one item is in the array, match the closest region
5. Replace field value with translation before returning.

## Questions

1. Should we support an array for `opts.lang`, since a user could have multiple preferred languages e.g. a user could have their primary language as Quechua, but Spanish might be a suitable fallback, rather than the default of English.
    I think this is maybe a follow-up issue for post-MVP.
2. How do we handle writes? Should writes always be in English, and then translations written separately via the API, or should write methods also take `opts.lang` and internally write to the translation API?
    Since we don't currently have any UI for writing any of the docs that we would translate (Presets, Fields), then I think the former is ok - write in English, then manually write translations - because this can be done within the config import code.

## Tasks

- [ ]

gmaclennan:
Fixes #472 

Depends on:

- [x] https://github.com/digidem/mapeo-schema/issues/165
- [x] #466

TODO:

- [ ] Add tests once dependencies are ready

The implementation in this draft PR tries to minimize the overhead for looking up translations.

Ready for an initial review.


gmaclennan:
@EvanHahn non-urgent, but it would be great to have your eyes on this this week to check this approach makes sense. A full review can maybe wait until we get the dependencies in place, but the approach here might affect how we design the translation API.

tomasciccola:
TranslationApi need `dataType` in its constructor, but `DataType` needs `translationApi` in its constructor. This means that we usually need to do:
```js
let translationApi;
const dataType = new DataType({
  dataStore,
  table,
  db,
  translation: translationApi
})
translationApi = new TranslationApi({
  dataType, 
 table
})
```
This kinda feels weird but it seems to work. Could there be edge cases where this wouldn't work? Is there a pattern that we could have were we avoid this?
A possibility would be to have an `init` function in, lets say, `DataType` so that we pass `translationApi` there, after instancing it...

src/datatype/index.js
```diff
@@ -172,26 +177,65 @@ export class DataType extends TypedEmitter {
...
+  /**
+   * @param {MapeoDoc} doc
+   * @param {{ lang?: string }} [opts]
+   */
+  async #translate(doc, { lang } = {}) {
+    if (!lang) return doc
+
+    const { language, region } = parseBCP47(lang)
+    if (!language) return doc
+    // const translatedDoc = JSON.parse(JSON.stringify(doc))
```
tomasciccola, line 207: I commented this line because I don't the reason for it. Everything seems to work fine, but I may be missing smth

src/datatype/index.js
```diff
@@ -172,26 +177,73 @@ export class DataType extends TypedEmitter {
 
   /**
    * @param {string} docId
+   * @param {{ lang?: string }} [opts]
    */
-  async getByDocId(docId) {
+  async getByDocId(docId, { lang } = {}) {
     await this.#dataStore.indexer.idle()
-    const result = this.#sql.getByDocId.get({ docId })
+    const result = /** @type {MapeoDoc} */ (this.#sql.getByDocId.get({ docId }))
```
EvanHahn, line 184: Nitpick: this type should include `undefined`, because we don't know (until the next line) that the doc is found.

```suggestion
    const result = /** @type {undefined | MapeoDoc} */ (
      this.#sql.getByDocId.get({ docId })
    )
```

src/datatype/index.js
```diff
@@ -172,26 +177,73 @@ export class DataType extends TypedEmitter {
...
     const rows = includeDeleted
       ? this.#sql.getManyWithDeleted.all()
       : this.#sql.getMany.all()
-    return rows.map((doc) => deNullify(doc))
+    return await Promise.all(
+      rows.map(
+        async (doc) =>
+          await this.#translate(
+            // @ts-ignore - too complicated to type this
+            deNullify(doc),
```
EvanHahn, line 242: Does this work?

```suggestion
            deNullify(/** @type {MapeoDoc} */ (doc)),
```

src/mapeo-project.js
```diff
@@ -206,51 +207,61 @@ export class MapeoProject extends TypedEmitter {
         dataStore: this.#dataStores.data,
         table: observationTable,
         db,
+        translation: this.$translation,
```
EvanHahn, line 210: In my testing, `this.$translation` is `undefined` at this point. I think we need a way to avoid this circular reference.

src/mapeo-project.js
```diff
@@ -206,51 +207,61 @@ export class MapeoProject extends TypedEmitter {
...
         dataStore: this.#dataStores.config,
         table: iconTable,
         db,
+        translation: this.$translation,
       }),
       translation: new DataType({
         dataStore: this.#dataStores.config,
         table: translationTable,
         db,
+        translation: this.$translation,
```
EvanHahn, line 264: Can we prevent people from using translations for translations, which would create an infinite loop?

src/datatype/index.js
```diff
@@ -172,26 +177,73 @@ export class DataType extends TypedEmitter {
...
+      docIdRef: doc.docId,
+      regionCode: region !== null ? region : undefined,
+    }
+    let translations = await this.#translation.get(value)
+    // if passing a region code returns no matches,
+    // fallback to matching only languageCode
+    if (translations.length === 0) {
+      value.regionCode = undefined
+    }
+    translations = await this.#translation.get(value)
```
EvanHahn, line 221: This does an extra lookup. To avoid that:

```suggestion
    if (translations.length === 0 && value.regionCode) {
      value.regionCode = undefined
      translations = await this.#translation.get(value)
    }
```

tests/data-type.js
```diff
@@ -167,6 +171,67 @@ test('delete()', async (t) => {
...
+
+  t.is(
+    translation.message,
+    getProperty(
+      await dataType.getByDocId(doc.docId, { lang: 'es' }),
+      translation.fieldRef
+    ),
+    `we get a valid translated field`
+  )
+  t.is(
```
EvanHahn, line 210: Nit: could you also test passing `es-AR`?

```js
t.is(
  translation.message,
  getProperty(
    await dataType.getByDocId(doc.docId, { lang: 'es-AR' }),
    translation.fieldRef
  ),
  'we get a valid translation'
)
```

src/datatype/index.js
```diff
@@ -172,26 +177,73 @@ export class DataType extends TypedEmitter {
...
+
+    for (let translation of translations) {
+      if (hasProperty(doc, translation.fieldRef)) {
```
EvanHahn, line 224: I worry about overwriting a non-string here. Could we do this instead?

```suggestion
      if (typeof getProperty(doc, translation.fieldRef) === 'string') {
```

Requires importing `getProperty` up above.

EvanHahn (changes requested):
Could you add an e2e test too?

src/mapeo-project.js
tomasciccola: How can be do that? could be make `translationApi` a noop if inside a `DataType` of translation? Or check every call to the api and avoid it if in translation `DataType` ?

src/datatype/index.js
EvanHahn: I think this is to clone `doc`...probably worth keeping in some form?

src/mapeo-project.js
EvanHahn: Done in #604.

tomasciccola: merged that!

test-e2e/translation-api.js
```diff
@@ -142,6 +142,57 @@ test('translation api - put() and get() fields', async (t) => {
...
+      docIdRef,
+      { lang: 'es-CO' }
+    )
+    t.is(
+      fallbackRegionCodeTranslatedField[fieldRef],
+      message,
+      `passing 'lang' with untranslated 'regionCode' returns a fallback translated 'fieldRef' matching 'languageCode'`
+    )
+    // console.log(fieldRef, message)
+    // const fields = await project.field.getMany({ lang: 'en' })
```
EvanHahn: Nitpick: can we remove these?

```suggestion
```

src/mapeo-project.js
tomasciccola: With your merged PR this would be solved, right?

test-e2e/translation-api.js
```diff
@@ -142,6 +142,97 @@ test('translation api - put() and get() fields', async (t) => {
...
+    })
+
+    t.is(
+      translatedField[fieldRef],
+      message,
+      `passing 'lang' returns the correct translated field`
+    )
+    const untranslatedField = await project.field.getByDocId(docIdRef)
+    t.not(
+      untranslatedField[fieldRef],
```
tomasciccola, line 179: Any ts incantation that would allow me to use the `fieldRef` as key without complains @EvanHahn ??

src/mapeo-project.js
EvanHahn: Yup!

test-e2e/translation-api.js
EvanHahn: Done in #619.

tomasciccola:
@EvanHahn do you feel there are e2e-tests that we can add? The ones I added are pretty similar to the translationApi tests with the added layer of going through `dataType.getByDocId(docId, {lang})` instead of using `translationApi.get()`

EvanHahn:
Nah, I think those are good enough!

I'll review the rest now.

EvanHahn (approved):
LGTM! Feel free to merge when CI passes.

Thanks for all the back-and-forth here.

PR #475, opened 2024-02-12
chore: update bonjour-service to v1.2.1

EvanHahn:
This version integrates all of our patches except the one for the `probe` option when starting a `Service`. That version also has some trickiness around types for `start` and `stop`, which are also fixed here.

I submitted these patches [upstream] but haven't heard back as of this writing.

[upstream]: https://github.com/onlxltd/bonjour-service/pull/48


src/discovery/dns-sd.js
```diff
@@ -119,7 +119,7 @@ export class DnsSd extends TypedEmitter {
         host: this.#name + '.local',
         port,
         type: SERVICE_NAME,
-        disableIpv6: this.#disableIpv6,
+        disableIPv6: this.#disableIpv6,
```
EvanHahn, line 122: Our old patch used a slightly different name than the "real" version upstream.

patches/bonjour-service+1.2.1.patch
EvanHahn, line 1: Again, I [submitted these changes upstream](https://github.com/onlxltd/bonjour-service/pull/48).

achou11 (approved):

PR #476, opened 2024-02-13
Fix: sort icons to group variants

tomasciccola:
given that this bug wasn't caught by tests, I use this PR to add a couple more tests of icons


src/config-import.js
```diff
@@ -65,10 +57,12 @@ export async function readConfig(configPath) {
       /** @type {IconData | undefined} */
       let icon
 
-      iconsZip = await yauzl.open(configPath)
-      const entries = await iconsZip.readEntries(MAX_ENTRIES)
-      for (const entry of entries) {
-        if (!entry.filename.match(/^icons\/([^/]+)$/)) continue
+      // we sort the icons by filename so we can group variants together
+      const iconEntries = entries
+        .filter((entry) => entry.filename.match(/^icons\/([^/]+)$/))
```
gmaclennan, line 62:
```suggestion
        .filter((entry) => entry.filename.startsWith('icons/')
```

Fine as it is, just not sure the regex is necessary

src/config-import.js
```diff
@@ -65,10 +57,12 @@ export async function readConfig(configPath) {
...
+        .sort((icon, nextIcon) => (icon.filename > nextIcon.filename ? 1 : -1))
```
gmaclennan, line 63:
```suggestion
        .sort((icon, nextIcon) => icon.filename.localeCompare(nextIcon.filename))
```
Not a necessary change, but just sharing this string method if you don't know about it - can be handy for this kind of logic. However note that your current implementation is missing the case when the name is equal, which should return `0`, but this is ok given that it should not be possible for this array to include two entries with the same filename - just pointing out that you would normally need this to ensure a stable sort order I think.

tests/config-import.js
```diff
@@ -131,12 +131,12 @@ test('config import - icons', async (t) => {
...
-      icon.variants.length,
-      9,
-      '9 variants of icons (dot product of density and size)'
-    )
+    // t.is(icon.name, 'plant', 'icon name is `plant`')
+    if (icon.name === 'plant') {
+      t.is(icon.variants.length, 3, '3 variants of plant icons')
+    } else if (icon.name === 'tree') {
+      t.is(icon.variants.length, 9, '9 - all - variants of tree icons')
+    }
```
gmaclennan, line 139: This test would pass if the fixture happened to not include icons with those names. The test should check that these assertions have been called - normally done with a `t.plan()`.

gmaclennan (changes requested):
Just a couple of nit-picks on the code, which don't actually need any change, since they do work fine as-is.

The test though needs a check to make sure that your new assertions are actually being called.

src/config-import.js
tomasciccola: The difference is that with the regex you only match files in the icon folder, where with the other option you will also match with 'icons/' (so the listing of the directory itself) which we don't want. Instead, we could do:
```js
.filter(
          (entry) =>
            entry.filename.startsWith('icons/') && !entry.filename.endsWith('/')
        )

```
but I dunno...

src/config-import.js
tomasciccola: uhh I like local compare! prefer it much more that comparing strings as if they were 'numbers'

src/config-import.js
gmaclennan: Yeah, just keep the regex

PR #477, opened 2024-02-14
chore: update @mapeo/schema to get new features

EvanHahn:
This updates `@mapeo/schema` to `3.0.0-next.14`, which pulls in two significant features:

1. New `Translation` schema
2. New `deviceType` property on `DeviceInfo`

Neither of these are useful right now but will unblock upcoming changes.


drizzle/project/0001_curved_korvac.sql
EvanHahn, line 1: Note that I added a new migration here. Because we haven't released I could pile this into the existing migration.

drizzle/project/meta/_journal.json
achou11, line 1: We should override the existing schema instead of introducing an actual migration

achou11 (changes requested):
The drizzle changes currently reflect a proper migration, but in this case we should just override the existing schema files

EvanHahn:
> The drizzle changes currently reflect a proper migration, but in this case we should just override the existing schema files

Done in fdbdcd34982dc990e9901b9b2f6eaede17b72c85.

package.json
```diff
@@ -111,7 +111,7 @@
     "@fastify/type-provider-typebox": "^3.3.0",
     "@hyperswarm/secret-stream": "^6.1.2",
     "@mapeo/crypto": "1.0.0-alpha.10",
-    "@mapeo/schema": "3.0.0-next.13",
+    "@mapeo/schema": "^3.0.0-next.14",
```
achou11, line 114: This should stay pinned

```suggestion
    "@mapeo/schema": "3.0.0-next.14",
```

package-lock.json
achou11, line 1: need to update after pinning issue is addressed

achou11 (changes requested):
Another note about deps pinning

package.json
EvanHahn: Done in 3b6069776123445b959de85c610c473785af016f.

achou11 (approved):
sorry for not getting to this sooner - thought i had already approved it!

EvanHahn:
No worries!

Issue #447, opened 2024-01-29
Track invites and allow cancelling invites


gmaclennan:
A peer might have sent an invite and then want to expire it manually, or change the invite (e.g. re-invite) with different role positions. This would require the receiver to:

A) Override a previously received invite with the new one
B) Include some kind of identifier in the response to the the invitor know _which_ invite is being responded to

gmaclennan:
After reviewing the implementation of invites in the front-end, it seems like we need a way to cancel invites.

If a user invites another device, they are currently left waiting for a response, and need a way to cancel if they do not receive a reply within an expected time-frame. The cancellation should ensure that when a response is received, it is ignored (currently _whenever_ a response to an invite is received, the invited device is automatically added to the project).

The cancellation should also send a wire message to the invitee, so that they can see that their invite has been cancelled (and this should result in the project key to be deleted, and the project instance to be deleted if it has not already been).

This points to perhaps the need for an additional step in the invite flow, of the invitee receiving a final message back about whether they were actually added to a project.

EvanHahn:
Here's a proposal for how we'd do this.

# tl;dr

Currently, two messages are sent:

1. Inviter sends an invite message, which has lots of info
1. Invitee sends ACCEPT, REJECT, or ALREADY
1. Inviter adds them to the project locally

I propose that _three_ messages be sent:

1. Inviter sends an invite message, which has limited info
1. Invitee sends ACCEPT, REJECT, or ALREADY
1. Inviter responds with project details (project key, encryption keys, etc) if they receive an ACCEPT, and add them to the project

Made a little diagram:

```mermaid
sequenceDiagram
  Inviter->>Invitee: Invite (limited info)
  Invitee->>Inviter: Accept
  Inviter->>Invitee: Project details (key, encryption keys, etc)
```

# Goals

Here are the goals of the invite system (I think):

- Users can invite multiple peers (though this may be hidden in the frontend)
- Users can receive multiple invites from other peers, possibly conflicting
- Users can respond to invites by accepting, rejecting, or saying they're already in the project
- Users can cancel outbound invites
- Users can re-invite peers for any reason
- Invites are culled and do not grow indefinitely
- Outbound invites can have a TTL (e.g., a timeout)
- Messages may be lost in transit
- Devices may have bugs (e.g., accepting invites twice) or be malicious
- *New:* when the inviter cancels the invite, the invitee should know (and not join the project)
[...]

PR #480, opened 2024-02-16
chore: invite API's membership check can be synchronous

EvanHahn:

This function doesn't do anything async so we can make it synchronous.  I think this is a useful change on its own but will make upcoming changes (#447) a bit simpler.


gmaclennan:
There is one thing missing from the existing implementation that is asynchronous: project membership also has a "joining" state, eg a user has accepted an invite but has not finished adding the project, eg if they do not receive back the project join details, or the add project method fails for some reason. Ideally other invites would await the joining state rather than treat it as an "already joined" state. I think it's ok to ignore that state for now, but making this asynchronous just leaves more work when we try to add it back. I'm ok with that as long as we don't get ourselves into a corner where we make it really hard to deal with this edge case in the future.

EvanHahn:
Good point. I'll make sure to do that in my upcoming work (#447).

In any case, I still think it makes sense for this to be synchronous because I don't think the "joining" state would be stored here.

gmaclennan (approved):
Great, looks good.

PR #481, opened 2024-02-16
chore!: remove "already part of project" invite response

EvanHahn:
For simplicity, we decided to remove the "ALREADY" invite response and use "REJECT" in its place. We can add it back in the future if needed, but this makes the code simpler on the front-end and back-end.

### DIFF
--- a/proto/rpc.proto
+++ b/proto/rpc.proto
@@ -18,7 +18,6 @@ message InviteResponse {
   enum Decision {
     REJECT = 0;
     ACCEPT = 1;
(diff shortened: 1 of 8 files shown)


gmaclennan:
How will the inviter know though without it? Do you plan to add a "reason" field to the reject response? I think we need to maintain the user experience for the inviter of knowing when inviting a device that it has "succeeded", in that they are part of the project.

EvanHahn:
After some discussion, we decided to abandon this.

PR #484, opened 2024-02-19
refactor!: use more steps for invites

EvanHahn:
_I recommend reviewing this with whitespace changes disabled._

See #447 for details.

I plan to add invite cancelations in an upcoming PR.

### DIFF
--- a/proto/rpc.proto
+++ b/proto/rpc.proto
@@ -3,25 +3,29 @@ syntax = "proto3";
 import "keys.proto";
 
 message Invite {
-  // Project info that is displayed to the user receiving the invite
-  message ProjectInfo {
-    optional string name = 1;
-  }
-  bytes projectKey = 1;
-  EncryptionKeys encryptionKeys = 2;
-  optional ProjectInfo projectInfo = 3;
-  string roleName = 4;
+  bytes inviteId = 1;
+  string projectPublicId = 2;
+  string projectName = 3;
+  optional string roleName = 4;
   optional string roleDescription = 5;
-  string invitorName = 6;
+  optional string invitorName = 6;
 }
+
 message InviteResponse {
   enum Decision {
     REJECT = 0;
     ACCEPT = 1;
     ALREADY = 2;
   }
-  bytes projectKey = 1;
+  bytes inviteId = 1;
   Decision decision = 2;
+}
+
+message ProjectJoinDetails {
+  bytes inviteId = 1;
+  bytes projectKey = 2;
+  EncryptionKeys encryptionKeys = 3;
+  string projectName = 4;
 }
 
 message DeviceInfo {
--- a/src/invite-api.js
+++ b/src/invite-api.js
@@ -1,56 +1,205 @@
 // @ts-check
 import { TypedEmitter } from 'tiny-typed-emitter'
 import { InviteResponse_Decision } from './generated/rpc.js'
-import { projectKeyToId, projectKeyToPublicId } from './utils.js'
+import { assert, keyToId } from './utils.js'
+import timingSafeEqual from './lib/timing-safe-equal.js'
+import { promiseWithResolvers } from './ponyfills.js'
+
+/**
+ * Internally, we typically use the `Invite` type from the protobuf. We also use
+ * an external type for public consumers.
+ *
+ * @internal
+ * @typedef {import('./generated/rpc.js').Invite} InviteInternal
+ */
+
+/**
+ * @internal
+ * @typedef {import('./generated/rpc.js').ProjectJoinDetails} ProjectJoinDetails
+ */
+
+/**
+ * @typedef {object} Invite
+ * @prop {string} inviteId
+ * @prop {string} projectName
+ * @prop {string} [roleName]
+ * @prop {string} [roleDescription]
+ * @prop {string} [invitorName]
+ */
+
+/**
+ * Manage pending invite state.
+ */
+class PendingInvites {
+  /**
+   * @internal
+   * @typedef {object} PendingInvite
+   * @prop {string} peerId
+   * @prop {InviteInternal} invite
+   * @prop {boolean} isAccepting
+   */
+
+  /** @type {Map<string, PendingInvite>} */
+  #byInviteId = new Map()
+
+  /** @type {Map<string, PendingInvite[]>} */
+  #byProjectPublicId = new Map()
+
+  /**
+   * @returns {Iterable<PendingInvite>} the pending invites, in insertion order
+   */
+  invites() {
+    return this.#byInviteId.values()
+  }
+
+  /**
+   * @param {PendingInvite} pendingInvite
+   * @throws if adding a duplicate invite ID
+   * @returns {void}
+   */
+  add(pendingInvite) {
+    const { invite } = pendingInvite
+    const { inviteId, projectPublicId } = invite
+
+    const inviteIdString = keyToId(inviteId)
+    assert(!this.#byInviteId.has(inviteIdString), 'Added duplicate invite')
+    this.#byInviteId.set(inviteIdString, pendingInvite)
+
+    const existingForProject = this.#byProjectPublicId.get(projectPublicId)
+    if (existingForProject) {
+      existingForProject.push(pendingInvite)
+    } else {
+      this.#byProjectPublicId.set(projectPublicId, [pendingInvite])
+    }
+  }
+
+  /**
+   * @param {Buffer} inviteId
+   * @returns {void}
+   */
+  markAccepting(inviteId) {
+    const inviteIdString = keyToId(inviteId)
+
+    const oldPendingInvite = this.#byInviteId.get(inviteIdString)
+    assert(!!oldPendingInvite, `Couldn't find invite for ${inviteIdString}`)
+
+    const newPendingInvite = { ...oldPendingInvite, isAccepting: true }
+
+    this.#byInviteId.set(inviteIdString, newPendingInvite)
+
+    const {
+      invite: { projectPublicId },
+    } = newPendingInvite
+    const newPendingInvitesForThisProject = (
+      this.#byProjectPublicId.get(projectPublicId) ?? []
+    ).map((pendingInvite) =>
+      pendingInvite.invite.inviteId.equals(inviteId)
+        ? newPendingInvite
+        : pendingInvite
+    )
+    this.#byProjectPublicId.set(
+      projectPublicId,
+      newPendingInvitesForThisProject
+    )
+  }
+
+  /**
+   * @param {Buffer} inviteId
+   * @returns {boolean}
+   */
+  hasInviteId(inviteId) {
+    const inviteIdString = keyToId(inviteId)
+    return this.#byInviteId.has(inviteIdString)
+  }
+
+  /**
+   * @param {string} projectPublicId
+   * @returns {boolean}
+   */
+  isAcceptingForProject(projectPublicId) {
+    return (this.#byProjectPublicId.get(projectPublicId) ?? []).some(
+      (pendingInvite) => pendingInvite.isAccepting
+    )
+  }
+
+  /**
+   * @param {Buffer} inviteId
+   * @returns {undefined | PendingInvite}
+   */
+  getByInviteId(inviteId) {
+    const inviteIdString = keyToId(inviteId)
+    return this.#byInviteId.get(inviteIdString)
+  }
+
+  /**
+   * @param {Buffer} inviteId
+   * @returns {boolean} `true` if an invite existed and was deleted, `false` otherwise
+   */
+  deleteByInviteId(inviteId) {
+    const inviteIdString = keyToId(inviteId)
+
+    const pendingInvite = this.#byInviteId.get(inviteIdString)
+    if (!pendingInvite) return false
+
+    this.#byInviteId.delete(inviteIdString)
+
+    const {
+      invite: { projectPublicId },
+    } = pendingInvite
+    const newPendingInvitesForThisProject = (
+      this.#byProjectPublicId.get(projectPublicId) ?? []
+    ).filter((pendingInvite) => pendingInvite.invite.inviteId.equals(inviteId))
+    if (newPendingInvitesForThisProject.length) {
+      this.#byProjectPublicId.set(
+        projectPublicId,
+        newPendingInvitesForThisProject
+      )
+    } else {
+      this.#byProjectPublicId.delete(projectPublicId)
+    }
+
+    return true
+  }
+
+  /**
+   * @param {string} projectPublicId
+   * @returns {PendingInvite[]} the pending invites that were deleted
+   */
+  deleteByProjectPublicId(projectPublicId) {
+    const pendingInvitesDeleted =
+      this.#byProjectPublicId.get(projectPublicId) ?? []
+
+    for (const { invite } of pendingInvitesDeleted) {
+      const inviteIdString = keyToId(invite.inviteId)
+      this.#byInviteId.delete(inviteIdString)
+    }
+
+    this.#byProjectPublicId.delete(projectPublicId)
+
+    return pendingInvitesDeleted
+  }
+}
 
 /**
  * @typedef {Object} InviteApiEvents
- *
- * @property {(info: { projectId: string, projectName?: string, peerId: string, roleName: string, roleDescription: string | undefined, invitorName: string }) => void} invite-received
+ * @property {(invite: Invite) => void} invite-received
+ * @property {(invite: Invite) => void} invite-removed
  */
-
-/**
- * @template K
- * @template V
- * @extends {Map<K, Set<V>>}
- */
-class MapOfSets extends Map {
-  /** @param {K} key */
-  get(key) {
-    const existing = super.get(key)
-    if (existing) return existing
-    /** @type {Set<V>} */
-    const set = new Set()
-    super.set(key, set)
-    return set
-  }
-}
 
 /**
  * @extends {TypedEmitter<InviteApiEvents>}
  */
 export class InviteApi extends TypedEmitter {
-  // Maps project public id -> project id
-  /** @type {Map<string, string>} */
-  #projectIdsMapping = new Map()
-
-  // Maps project id -> set of device ids
-  /** @type {MapOfSets<string, string>} */
-  #peersToRespondTo = new MapOfSets()
-
-  // Maps project id -> { invite, fromPeerId }
-  /** @type {Map<string, { fromPeerId: string, invite: import('./generated/rpc.js').Invite }>} */
-  #pendingInvites = new Map()
-
   #isMember
   #addProject
+  #pendingInvites
 
   /**
    * @param {Object} options
    * @param {import('./local-peers.js').LocalPeers} options.rpc
    * @param {object} options.queries
-   * @param {(projectId: string) => Promise<boolean>} options.queries.isMember
-   * @param {(invite: import('./generated/rpc.js').Invite) => Promise<void>} options.queries.addProject
+   * @param {(projectId: string) => boolean} options.queries.isMember
+   * @param {(projectDetails: Pick<ProjectJoinDetails, 'projectKey' | 'encryptionKeys' | 'projectName'>) => Promise<unknown>} options.queries.addProject
    */
   constructor({ rpc, queries }) {
     super()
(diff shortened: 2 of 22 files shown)


src/member-api.js
```diff
@@ -54,42 +75,148 @@ export class MemberApi extends TypedEmitter {
...
+   * )>}
    */
-  async invite(deviceId, { roleId, roleName, roleDescription, timeout }) {
+  async invite(
+    deviceId,
+    {
+      roleId,
+      roleName = ROLES[roleId]?.name,
+      roleDescription,
+      timeout = 1_800_000,
```
EvanHahn, line 90: This is 30 minutes. Is that a reasonable timeout?

src/invite-api.js
```diff
@@ -1,236 +1,440 @@
...
 import { InviteResponse_Decision } from './generated/rpc.js'
-import { projectKeyToId, projectKeyToPublicId } from './utils.js'
+import { assert, keyToId } from './utils.js'
+import timingSafeEqual from './lib/timing-safe-equal.js'
+import { promiseWithResolvers } from './ponyfills.js'
 
 /**
- * @typedef {Object} InviteApiEvents
+ * Internally, we typically use the `Invite` type from the protobuf. We also use
+ * an external type for public consumers.
```
EvanHahn, line 10: I don't love this, but it's the best of three options I evaluated. The main issue: we want to use `Buffer`s over the wire, but `string`s in the public API.

The three options I evaluated:

1. Always use strings, even in the protos
2. Always use strings _except_ in the protos; we'd convert to/from buffers when sending/receiving the bytes
3. Only use strings for the public API, and use `Buffer`s everywhere else

Better ideas are welcome!

src/local-peers.js
EvanHahn, line 1: Note that this file no longer has any invite state.

src/mapeo-manager.js
```diff
@@ -149,17 +149,14 @@ export class MapeoManager extends TypedEmitter {
     this.#invite = new InviteApi({
       rpc: this.#localPeers,
       queries: {
-        isMember: async (projectId) => {
-          const projectExists = this.#db
+        isMember: (projectPublicId) =>
```
EvanHahn, line 152: I split this into a smaller change in #480, which can be reviewed/merged separately.

src/member-api.js
```diff
@@ -54,42 +75,148 @@ export class MemberApi extends TypedEmitter {
...
+        case InviteResponse_Decision.REJECT:
+          return inviteResponse.decision
+        case InviteResponse_Decision.UNRECOGNIZED:
+          return InviteResponse_Decision.REJECT
+        case InviteResponse_Decision.ACCEPT:
+          handleAborted()
+
+          // We should assign the role locally *before* sharing the project details
+          // so that they're part of the project even if they don't receive the
+          // project details message.
```
EvanHahn, line 202: Calling this out, as I'm not sure it's correct.

src/invite-api.js
```diff
@@ -1,236 +1,440 @@
...
+    } else {
+      this.#byProjectPublicId.set(projectPublicId, [pendingInvite])
+    }
+  }
+
+  /**
+   * @param {Buffer} inviteId
+   * @returns {void}
+   */
+  markAccepting(inviteId) {
```
gmaclennan, line 80: I'm often tempted to just mutate objects for something like this, but I should probably ignore those evil thoughts

src/invite-api.js
```diff
@@ -1,236 +1,440 @@
...
+      projectPublicId,
+      newPendingInvitesForThisProject
+    )
+  }
+
+  /**
+   * @param {Buffer} inviteId
+   * @returns {boolean}
+   */
+  hasInviteId(inviteId) {
```
gmaclennan, line 110: A helper could end up be handy here. Not sure if it's warranted/needed.

```js
const BufferMap extends Map {
  get(key) { return super.get(keyToId(key)) }
  set(key, value) { return super.set(keyToId(key), value) }
  has(key) { return super.get(keyToId(key)) }
  delete(key) { return super.delete(keyToId(key)) }
}
```

src/invite-api.js
```diff
@@ -1,236 +1,440 @@
...
     this.#isMember = queries.isMember
     this.#addProject = queries.addProject
 
-    this.rpc.on('invite', (peerId, invite) => {
-      this.#handleInvite(peerId, invite).catch(() => {
-        /* c8 ignore next */
-        // TODO: Log errors, but otherwise ignore them, but can't think of a reason there would be an error here
-      })
-    })
+    this.rpc.on('invite', this.#handleInvite)
```
gmaclennan, line 210: I think thrown errors in #handleInvite are no longer being caught, but they should be, to avoid uncaught errors.

src/invite-api.js
```diff
@@ -1,236 +1,440 @@
...
-    const peerIds = this.#peersToRespondTo.get(projectId)
-    peerIds.add(peerId)
-
-    if (this.#pendingInvites.has(projectId)) return
+    const hasAlreadyReceivedThisInvite = this.#pendingInvites.hasInviteId(
+      invite.inviteId
+    )
+    if (hasAlreadyReceivedThisInvite) {
+      return
+    }
```
gmaclennan, line 231: Ah, I see you are doing the check here that you throw for in the pending invites state. I think we still need to catch errors in #handleInvite

src/invite-api.js
```diff
@@ -1,236 +1,440 @@
...
+   * @prop {string} peerId
+   * @prop {InviteInternal} invite
+   * @prop {boolean} isAccepting
+   */
+
+  /** @type {Map<string, PendingInvite>} */
+  #byInviteId = new Map()
+
+  /** @type {Map<string, PendingInvite[]>} */
+  #byProjectPublicId = new Map()
```
gmaclennan, line 46: Having this map feels like a premature optimization - I think it's just an index of `#byInviteId`, but having it requires a lot of additional code to keep it in sync with `#byInviteId`. Is it necessary?

src/invite-api.js
```diff
@@ -1,236 +1,440 @@
...
+        // This peer ID check is probably superfluous because the invite ID
+        // should be impossible to guess, but is easy to add for correctness.
+        timingSafeEqual(projectDetailsPeerId, peerId) &&
+        timingSafeEqual(inviteId, details.inviteId)
+      if (!isDetailsForThisInvite) return
+
+      gotProjectDetails(details)
+      this.rpc.off('got-project-details', onProjectDetails)
     }
+    this.rpc.on('got-project-details', onProjectDetails)
```
gmaclennan, line 309: I don't think this needs to be a deferred promise, and I've tried to avoid them when I can, purely based on [this random advice](https://github.com/petkaantonov/bluebird/wiki/Promise-anti-patterns#the-deferred-anti-pattern) that I don't know if it's correct. But I think this is maybe easier to read as a regular promise, something like:

```js
const projectDetailsPromise = new Promise(resolve => {
  const { rpc } = this
  rpc.on('got-project-defailts', function onProjectDetails(projectDetailsPeerId, details) {
    const isDetailsForThisInvite =
      // This peer ID check is probably superfluous because the invite ID
      // should be impossible to guess, but is easy to add for correctness.
      timingSafeEqual(projectDetailsPeerId, peerId) &&
      timingSafeEqual(inviteId, details.inviteId)
    if (!isDetailsForThisInvite) return
    rpc.off('got-project-details', onProjectDetails)
    resolve(details)
  })
})
```

gmaclennan (commented):
I ran out of time today to complete the review - there is a lot of code here to process. Some initial comments here, which might be resolved in other parts of the code that I have not read yet.

proto/rpc.proto
```diff
@@ -3,27 +3,31 @@ syntax = "proto3";
...
-  bytes projectKey = 1;
+  bytes inviteId = 1;
   Decision decision = 2;
 }
 
+message ProjectJoinDetails {
+  bytes inviteId = 1;
+  bytes projectKey = 2;
+  EncryptionKeys encryptionKeys = 3;
+  string projectName = 4;
```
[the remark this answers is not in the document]
EvanHahn: Not sure. I assumed that the invitee would want to see the project name. If they don't, we can omit it here.

src/invite-api.js
```diff
@@ -1,236 +1,440 @@
...
+/**
+ * @internal
+ * @typedef {import('./generated/rpc.js').ProjectJoinDetails} ProjectJoinDetails
  */
 
 /**
- * @template K
- * @template V
- * @extends {Map<K, Set<V>>}
+ * @typedef {object} Invite
```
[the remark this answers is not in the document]
EvanHahn: Cool, I didn't know you could do that! I think we want to make two changes to the external type:

1. The external type uses strings, not buffers (solved by your solution)
2. The external type does not have all fields. Notably, it's missing `projectPublicId`

I feel like this type adds a little repetition (bad) but is probably simpler than a fancy type that accomplishes both of those goals.

src/invite-api.js
```diff
@@ -1,236 +1,440 @@
...
+  /**
+   * @param {PendingInvite} pendingInvite
+   * @throws if adding a duplicate invite ID
+   * @returns {void}
+   */
+  add(pendingInvite) {
+    const { invite } = pendingInvite
+    const { inviteId, projectPublicId } = invite
+
+    const inviteIdString = keyToId(inviteId)
```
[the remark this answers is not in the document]
EvanHahn: I named this so _of course_ it seems reasonable to me...but `inviteKey` implies cryptographic properties to me. I think `inviteId` is the clearest, but I'm happy to be swayed.

src/invite-api.js
EvanHahn: I considered that too...the siren call of mutation beckons us both. I opted to avoid mutation here because these objects are used outside this class, but I don't feel strongly.

proto/rpc.proto
```diff
@@ -3,27 +3,31 @@ syntax = "proto3";
...
-  EncryptionKeys encryptionKeys = 2;
-  optional ProjectInfo projectInfo = 3;
-  string roleName = 4;
+  bytes inviteId = 1;
+  string projectPublicId = 2;
+  string projectName = 3;
+  optional string roleName = 4;
   optional string roleDescription = 5;
-  string invitorName = 6;
+  optional string invitorName = 6;
```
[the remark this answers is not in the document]
EvanHahn: Done in e1ebe54cd3f5562f01e0de3c536f7de408720287.

src/invite-api.js
EvanHahn: Made a generic `HashMap` class in 8ad29aa2237cda365cc7b1e2653d94ca62e9ed6.

src/invite-api.js
EvanHahn: Addressed in d07b4a52ec3663f818aabdade188b1ed8bbe3a75.

src/invite-api.js
EvanHahn: Good point. Removed in 539916f976eb3927cd1b955e8b1fcaca20233f87.

src/invite-api.js
EvanHahn: I think it needs to work this way because we want to remove the listener if the accept fails for any reason. I thought I'd done that but I guess not! Done in f71351d14c6c7ca297bf8f2846c9076797b3e3b3.

src/local-peers.js
```diff
@@ -27,6 +28,7 @@ const DEDUPE_TIMEOUT = 1000
 const MESSAGE_TYPES = {
   Invite: 0,
   InviteResponse: 1,
+  ProjectJoinDetails: 3,
```
EvanHahn, line 31: Based on the comment above, I didn't reorder this and just added this as message type 3. Let me know if I should reorder given that we haven't released.

src/invite-api.js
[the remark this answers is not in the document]
gmaclennan: Yeah, I agree. We had a convention earlier on to use `key` for buffers, and `id` for their string representation, but we didn't stick to it, and I think the implication of cryptographic properties is confusing. It would be good to adopt a naming convention however that does clearly identify is something is a buffer or a string representation of the same thing.

src/invite-api.js
gmaclennan: That seems like a good reason to me.

src/local-peers.js
gmaclennan: not sure which comment you are referring to? I think for something like this maybe either re-number to `2` if not released, or else keep the removed type in the list with a comment explaining it's deprecated, otherwise someone in the future looks at this and spends a while scratching their head about how it ended up like this.

proto/rpc.proto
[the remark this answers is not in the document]
gmaclennan: I think it's necessary in the invite, which is what the invitee sees when they make a decision whether to join, but I hadn't thought there is any UI associated with receiving ProjectJoinDetails?

src/invite-api.js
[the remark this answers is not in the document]
EvanHahn: Same!

proto/rpc.proto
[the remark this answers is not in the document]
EvanHahn: Ah, sorry, I thought this was a comment for another part of the proto.

I don't think there's any UI associated with receiving `ProjectJoinDetails`. This was here in case the project's name changed after the initial invite was sent. (But maybe we don't want that?)

src/local-peers.js
EvanHahn: Reordered in 34465832688eb641bb8c37483b85469d2b244a84 (leaving a space for invite cancellations).

src/local-peers.js
```diff
@@ -517,30 +470,38 @@ export class LocalPeers extends TypedEmitter {
     switch (type) {
       case 'Invite': {
         const invite = Invite.decode(value)
-        assertInviteHasKeys(invite)
+        if (!isInviteValid(invite)) {
+          this.#l.log('Received invalid invite')
+          return
```
gmaclennan, line 475: I think we should throw here, since `Invite.decode()` will also throw for invalid invites (e.g. can't parse the protobuf), which makes me realise that we need to catch unhandled errors when calling this within `onmessage` of protomux.

src/mapeo-manager.js
```diff
@@ -368,6 +365,7 @@ export class MapeoManager extends TypedEmitter {
       projectId,
       projectPublicId,
       projectKeys: keys,
+      projectInfo: { name: settings.name },
```
gmaclennan, line 368: The project name (and other info about a project) is indexed to the `projectSettingsTable` once a project is synced. Do we need to replicate it here? Maybe... for a project that hasn't synced yet, but I don't think we consider a project joined until we actually complete that first sync?

src/member-api.js
gmaclennan: This seems too long maybe. My gut says 5 mins, but should check with Jen and others on the team who have done trainings, but we can adjust in the future. Can you move this to a `const DEFAULT_INVITE_TIMEOUT` at the top of the file?

src/member-api.js
```diff
@@ -54,42 +75,148 @@ export class MemberApi extends TypedEmitter {
...
+      timeout = 1_800_000,
+    }
+  ) {
     assert(isRoleIdForNewInvite(roleId), 'Invalid role ID for new invite')
+    assert(
+      !this.#deviceIdsWithPendingInvites.has(deviceId),
+      'Already inviting this device ID'
+    )
+    assert(timeout > 0, 'Timeout should be a positive number')
+    assert(timeout < 2 ** 32, 'Timeout is too large')
```
gmaclennan, line 99: ha! but 50 days is ok? (no need to change)

src/member-api.js
```diff
@@ -54,42 +75,148 @@ export class MemberApi extends TypedEmitter {
...
-   *
-   * @returns {Promise<import('./generated/rpc.js').InviteResponse_Decision>}
+   * @returns {Promise<(
+   *   typeof InviteResponse_Decision.ACCEPT |
+   *   typeof InviteResponse_Decision.REJECT |
+   *   typeof InviteResponse_Decision.ALREADY
+   * )>}
    */
-  async invite(deviceId, { roleId, roleName, roleDescription, timeout }) {
+  async invite(
```
gmaclennan, line 84: Not for this PR, but a note for the future: I don't think it's a great API design having these long-running promises. It means we need special handling on the RPC bridge, because normally we would consider a promise taking this long to resolve an error. At some point I think we should change this to an events-based API e.g. `on('invite-accepted', (inviteId) => {})`

src/member-api.js
gmaclennan: I think this is ok, although also not entirely sure. If we wait until after we don't know whether they have maybe received the join details, but we lost the connection before we learned that, which would leave them "in the project" but without a role, which would be a bad experience, so I think the order here is the safest thing to do.

src/member-api.js
```diff
@@ -54,42 +75,148 @@ export class MemberApi extends TypedEmitter {
...
+      if (!projectName)
+        throw new Error('Project must have a name to invite people')
+      handleAborted()
 
-    if (response === InviteResponse_Decision.ACCEPT) {
-      await this.#roles.assignRole(deviceId, roleId)
-    }
+      const inviteId = crypto.randomBytes(32)
+
+      const {
```
gmaclennan, line 152: Again I think there is a better pattern than deferred promises, but it's ok for now.

gmaclennan (commented):
Didn't review the tests, but went through all the other code, and it all looks good other than a couple of small changes. Opinions about deferred promises can be ignored for now.

A thought looking at this, a utility function like this would be helpful:

```js
async function onceSatisfied(emitter, eventName, check, { signal }) {
  return new Promise((res, rej) => {
    signal.once('abort', () => {
       emitter.off(eventName, onEvent)
    })
    emitter.on(eventName, function onEvent (...args) {
        const satisfied = checked.apply(null, args)
        if (!satisfied) return
        emitter.off(eventName, onEvent)
        res.apply(null, args)
    })
  })
```

We've got quite a few places where we want a `await once()`, but waiting for an event with specific arguments. A helper like that would make writing the code without the deferred promise a bit easier.

src/lib/hashmap.js
```diff
@@ -0,0 +1,93 @@
+// @ts-check
```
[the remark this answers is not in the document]
EvanHahn: Good idea. Added to our "long-term tech improvements" list.

proto/rpc.proto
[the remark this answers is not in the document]
EvanHahn: Removed in f7e6864a45fad76dc142c367e96f63e2703878e8.

src/invite-api.js
```diff
@@ -1,231 +1,396 @@
...
+        timingSafeEqual(projectDetailsPeerId, peerId) &&
+        timingSafeEqual(inviteId, details.inviteId)
+      if (!isDetailsForThisInvite) return
+
+      gotProjectDetails(details)
+      this.rpc.off('got-project-details', onProjectDetails)
     }
+    this.rpc.on('got-project-details', onProjectDetails)
+
+    assert(
```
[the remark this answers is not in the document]
EvanHahn: Done in 941ca2c4b2780ed6020f43b4b0f2a792a136a084.

src/invite-api.js
```diff
@@ -1,231 +1,396 @@
...
       return
     }
 
-    const pendingInvite = this.#pendingInvites.get(projectId)
-
-    if (!pendingInvite) {
-      throw new Error(
-        `Cannot find invite for project with ID ${projectPublicId}`
-      )
+    const { promise: projectDetailsPromise, resolve: gotProjectDetails } =
```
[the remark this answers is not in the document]
EvanHahn: Updated this in 0a625010c5117cf285103fd9d64951280405bc71.

src/member-api.js
EvanHahn: Addressed in 0a625010c5117cf285103fd9d64951280405bc71.

src/member-api.js
EvanHahn: Lowered to 5 minutes and moved to a variable in 210ab8bca6ed1f415daf144396b64f8354a8d56d.

EvanHahn: I'll follow up on this number ~next week.

src/invite-api.js
```diff
@@ -1,231 +1,391 @@
...
+      )(
+        this.rpc,
+        'got-project-details',
+        (projectDetailsPeerId, details) =>
+          // This peer ID check is probably superfluous because the invite ID
+          // should be impossible to guess, but is easy to add for correctness.
+          timingSafeEqual(projectDetailsPeerId, peerId) &&
+          timingSafeEqual(inviteId, details.inviteId),
+        { signal: projectDetailsAbortController.signal }
+      ).then((args) => args?.[1])
```
gmaclennan, line 266: I think this is needed to avoid an uncaught error if `#sendAcceptResponse` below throws, since this could throw in a tick after that once this function context no longer exists.

```suggestion

projectDetailsPromise.catch(noop)
```

gmaclennan: (we do want `projectDetailsProject` to be able to throw, hence this is on a separate line)

src/member-api.js
```diff
@@ -54,42 +77,140 @@ export class MemberApi extends TypedEmitter {
...
+          onceSatisfied
+        )(
+          this.#rpc,
+          'invite-response',
+          (peerId, inviteResponse) =>
+            timingSafeEqual(peerId, deviceId) &&
+            timingSafeEqual(inviteId, inviteResponse.inviteId),
+          { signal: receiveInviteAbortController.signal }
+        ).then((args) => args?.[1])
+
```
gmaclennan, line 169:
```suggestion

receiveInviteResponsePromise.catch(noop)
```

src/member-api.js
```diff
@@ -54,42 +77,140 @@ export class MemberApi extends TypedEmitter {
...
-    // since we are always getting #ownDeviceId,
-    // this should never throw (see comment on getById), but it pleases ts
-    if (!deviceName)
-      throw new Error(
-        'Internal error trying to read own device name for this invite'
-      )
+    try {
+      let handleAborted = noop
+      const timeoutId = setTimeout(() => {
+        handleAborted = () => {
```
gmaclennan, line 122: I think you need to abort the abortSignal in this timeout?

gmaclennan (changes requested):
Just reviewed the onceSatisfied changes and left some comments - I think these need fixed.
Do you know why CI is failing?

EvanHahn:
> Do you know why CI is failing?

Yes, silly reason: I canceled it. I did this because I knew I was gonna keep working, so I didn't want to waste GitHub's time.

src/invite-api.js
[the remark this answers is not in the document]
EvanHahn: Done in 1c1e3c1e2672444438a33144953fe223dfa5329b.

src/invite-api.js
```diff
@@ -1,231 +1,396 @@
...
+  if (internal.roleName) {
+    result.roleName = internal.roleName
+  }
+  if (internal.roleDescription) {
+    result.roleDescription = internal.roleDescription
+  }
+  if (internal.invitorName) {
+    result.invitorName = internal.invitorName
   }
+  return result
```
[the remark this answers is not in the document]
EvanHahn: Done in 1c1e3c1e2672444438a33144953fe223dfa5329b.

src/local-peers.js
EvanHahn: Made this throw in 572d09cf9d030596f3b37a5ea4542321e629bb76, which required catching the error in `onmessage` as you mentioned. Filed #514 in case there's more we want to do around this in general.

src/invite-api.js
```diff
@@ -1,231 +1,396 @@
...
-    }
-    for (const peerId of this.#peersToRespondTo.get(projectId)) {
-      await this.#sendRejectResponse({ peerId, projectId })
-    }
+    const pendingInvite = this.#pendingInvites.getByInviteId(inviteId)
+    assert(!!pendingInvite, `Cannot find invite ${inviteId}`)
+
+    const { peerId, invite } = pendingInvite
+
+    this.#sendRejectResponse({ peerId, inviteId: invite.inviteId })
```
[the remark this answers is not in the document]
EvanHahn: Addressed in ac786170b93cf2008d98df73ad33502656a979e3.

src/invite-api.js
```diff
@@ -1,231 +1,396 @@
...
    */
-  #handleInvite = (peerId, invite) => {
-    const projectId = projectKeyToId(invite.projectKey)
-    const isAlreadyMember = this.#isMember(projectId)
-
+  #handleInvite(peerId, invite) {
+    const isAlreadyMember = this.#isMember(invite.projectPublicId)
     if (isAlreadyMember) {
-      this.#sendAlreadyResponse({ peerId, projectId })
+      this.#sendAlreadyResponse({ peerId, inviteId: invite.inviteId })
```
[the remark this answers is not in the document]
EvanHahn: Addressed in ac786170b93cf2008d98df73ad33502656a979e3.

src/invite-api.js
EvanHahn: I can easily change this, but `onceSatisfied` should never reject, even if aborted—it will resolve `null`. Therefore, I don't _think_ this needs to be caught. Happy to change any of that, including my understanding.

src/member-api.js
EvanHahn: See [my comment above](https://github.com/digidem/mapeo-core-next/pull/484#discussion_r1520359400)—I don't think this promise will ever reject.

src/member-api.js
EvanHahn: That's done here:

https://github.com/digidem/mapeo-core-next/blob/210ab8bca6ed1f415daf144396b64f8354a8d56d/src/member-api.js#L155-L157

src/mapeo-manager.js
EvanHahn: Not sure why I added this! Removed in 6ff40d284c4f271b3efe3416c5e1b4ee770a8c42.

EvanHahn:
I believe I've addressed all comments. Thanks for the thorough review!

src/invite-api.js
gmaclennan: ah, right, I hadn't fully parsed the implementation of `onceSatisfied`. I see it won't reject, but notice it can result in an uncaught error, if the `check` function throws, e.g. in this case it would throw if somehow `projectDetailsPeerId` or other params were a `number` or `undefined` (`timingSafeEqual` would throw a `TypeError`).

I think it's always best to assume that async functions throw, and make sure that they are caught, rather than relying on the assumption that they won't throw, because down the line it's easy to add some code that results in them throwing, and it's hard to retain the memory of "this function must never throw".

In a case like this I'm not a huge fan of returning `null` to indicate abort or an error from a helper like this, although I think that's just a personal preference. I prefer to throw, and if I want to use `null` as my check then I can do `fnThatMightThrow().catch(() => null)` or I can choose to handle the error a different way.

gmaclennan: TL;DR, `onceSatisfied` needs to wrap `check()` in a try...catch, and either reject or return `null` if it throws. My preference is to throw, but fine with keeping `res(null)` if preferred. I think we should have the `.catch(noop)` anyway, so anyone refactoring / changing code doesn't have to remember which async functions can throw and which can't.

src/member-api.js
gmaclennan: yeah but if we never receive an invite response, `await receiveInviteResponsePromise` will never resolve, and the cleanup functions will never be called. Or am I missing something?

EvanHahn: I _think_ that sequence of events would be:

1. We set up the receive invite abort controller.
   https://github.com/digidem/mapeo-core-next/blob/25d3c073cb9992ea66f0ecf381bcf5efb2819cea/src/member-api.js#L154
1. We immediately add it to our list of cleanup functions.
   https://github.com/digidem/mapeo-core-next/blob/25d3c073cb9992ea66f0ecf381bcf5efb2819cea/src/member-api.js#L155-L157
1. The timeout fires and `handleAborted` now throws.
   https://github.com/digidem/mapeo-core-next/blob/25d3c073cb9992ea66f0ecf381bcf5efb2819cea/src/member-api.js#L121-L125
1. `handleAborted` is called, which throws an error.
   https://github.com/digidem/mapeo-core-next/blob/25d3c073cb9992ea66f0ecf381bcf5efb2819cea/src/member-api.js#L179
1. The error is caught and cleaned up.
   https://github.com/digidem/mapeo-core-next/blob/25d3c073cb9992ea66f0ecf381bcf5efb2819cea/src/member-api.js#L211-L213

More broadly: even if this code is technically correct[^0], it's hard to follow. I spent a good amount of time trying to make this clearer but I know I missed at least one good idea: `onceSatisfied`. I'll try to think of further ways to simplify this, as it's hard to understand.

[^0]: which it may not be

src/member-api.js
EvanHahn: It now might reject! Updated in 759e5c5cf90daf00d346c34df35ec3768420b4e4.

src/invite-api.js
EvanHahn: `onceSatisfied` now throws. I added `.catch(noop)` to its calls. See 759e5c5cf90daf00d346c34df35ec3768420b4e4.

EvanHahn:
Addressed the next round of comments!

src/member-api.js
gmaclennan: The code that is going to 
potentially take a long time and we want to timeout is awaiting the invite to be accepted on line 181. If that never resolves, then I don't see how the timeout can abort it, since the abort only happens if the code keeps executing.

EvanHahn: You're totally right. Fixed in dbfb6a915912aca53dde9a0a3225c511d5408ac5.

PR #488, opened 2024-02-20
chore: fix flaky local peer device info test

EvanHahn:
`waitForPeers()` has a `waitForDeviceInfo` option. When enabled, this helper should resolve only after device info is populated.

It had a tricky bug that only manifested some of the time (see [example CI failure 1][1], [2]). In some cases, the function could _think_ device info was populated when it wasn't.[^CI]

To fix this, I reworked the helper. It now does redundant work but it's now stateless-ish, which should avoid this problem.

[^CI]: Well, that's what I could reproduce locally. The errors on CI are unhelpful ([which I hope to fix in Brittle][brittle PR]). It's possible a different problem exists on CI that I couldn't repro locally.

[1]: https://github.com/digidem/mapeo-core-next/actions/runs/7925562783/job/21731451900
[2]: https://github.com/digidem/mapeo-core-next/actions/runs/7965492933/job/21745121763
[brittle PR]: https://github.com/holepunchto/brittle/pull/44

### DIFF
--- a/test-e2e/utils.js
+++ b/test-e2e/utils.js
@@ -118,50 +118,37 @@ export async function invite({
  *
  * @param {readonly MapeoManager[]} managers
  * @param {{ waitForDeviceInfo?: boolean }} [opts] Optionally wait for device names to be set
+ * @returns {Promise<void>}
  */
-export async function waitForPeers(
-  managers,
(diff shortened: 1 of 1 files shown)


gmaclennan (approved):
great! thanks for tracking that down

PR #489, opened 2024-02-21
chore: restore bitCount32 test

EvanHahn:
_This change is not urgent._

bitCount32's test is skipped because it is slow, and it's slow because it iterates over all integers between 0 and 2<sup>32</sup>.

This keeps that test skipped but adds a new test that we can run, which only tests a few inputs. It also speeds up the test-only comparison function by about 9x in my testing.


gmaclennan (approved):
Great. Was initially hesitant because the idea of the test/comparison bit count is something so simple it doesn't need a test itself, and bitwise functions scare me, but re-reading the `slowBitCount` function, I think it's obvious enough to be used as a test comparison case.

Looks good to me.

PR #494, opened 2024-02-21
chore: enable useUnknownInCatchVariables TypeScript option

EvanHahn:
This enables the [useUnknownInCatchVariables] TypeScript option, bringing a little more type safety.

I fixed a few violations, in part by moving `isRecord` to its own file with tests.

[useUnknownInCatchVariables]: https://www.typescriptlang.org/tsconfig/#useUnknownInCatchVariables


src/config-import.js
```diff
@@ -253,18 +254,23 @@ function validatePresetsFile(presetsFile) {
...
  */
 function hasOwn(obj, prop) {
   return Object.prototype.hasOwnProperty.call(obj, prop)
 }
+
+/**
+ * @param {unknown} value
+ * @returns {Error}
+ */
+function errify(value) {
```
EvanHahn, line 269: Probably overkill...

src/lib/object.js
```diff
@@ -0,0 +1,9 @@
+// @ts-check
+
+/**
+ * @param {unknown} value
+ * @returns {value is Record<string, unknown>}
+ */
+export function isRecord(value) {
+  return value !== null && typeof value === 'object' && !Array.isArray(value)
+}
```
achou11, line 9: mixed feelings on introducing this. can `Record<string, unknown>` can be used for class instance types? 

maybe not important given the usage - maybe it's a matter of naming? my understanding is that this is basically narrowing the type to something that can be indexed by a string for the purposes of basic property access

src/blob-store/live-download.js
```diff
@@ -241,7 +242,7 @@ export class DriveLiveDownload extends TypedEmitter {
       const [blobs] = await once(this.#drive, 'blobs', { signal: this.#signal })
       return blobs.core
     } catch (e) {
-      if (e.name === 'AbortError') return
+      if (isRecord(e) && e.name === 'AbortError') return
```
achou11, line 245: instead of relying on the helper, maybe something like this would suffice?

```suggestion
      if (e instanceof Error && e.name === 'AbortError') return
```

src/config-import.js
```diff
@@ -253,18 +254,23 @@ function validatePresetsFile(presetsFile) {
...
+  if (value instanceof Error) return value
+  if (typeof value === 'string') return new Error(value)
+  if (isRecord(value) && typeof value.message === 'string') {
```
achou11, line 272: again not sure if the helper is really necessary. maybe i just prefer inlined verbosity 😄

achou11 (approved):
nothing blocking but added some thoughts

src/lib/object.js
EvanHahn: Yeah, that's the idea. I named it to match TypeScript's built-in `Record` type (a name I don't personally love).

src/config-import.js
EvanHahn: Simplified in 3febedcba0e512880841eee3eded8ef68072b7f0.

achou11 (changes requested):
Looks like the `isRecord()` implementation can be completely removed now

achou11 (approved):
my bad, didn't realize isRecord() existed for other parts of the codebase prior to this PR

EvanHahn:
> Looks like the `isRecord()` implementation can be completely removed now

It looks like it's still needed but only by one place now, so I've moved it back in aaf0af5604bc2e43b75c34adddcb2d75d612d031.

PR #498, opened 2024-02-22
fix: force-closing local discovery TCP server could always time out

EvanHahn:
We have a way to force-close the local discovery TCP server. This was susceptible to the following race condition bug:

1. A connection opens and is established.

2. A duplicate connection opens. We decide to keep the newer one. We start to close the old connection and enqueue a "promotion" of the new one.

   https://github.com/digidem/mapeo-core-next/blob/5e05e35200297435f4ebd33df2c5b1444ed4ef1b/src/discovery/local-discovery.js#L229-L232

4. We close the TCP server. This does not immediately halt the server; it ["stops the server from accepting new connections and keeps existing connections"][docs]. In other words, it closes the doors but doesn't kick people out.

   We iterate through all existing connections and destroy them...

   https://github.com/digidem/mapeo-core-next/blob/5e05e35200297435f4ebd33df2c5b1444ed4ef1b/src/discovery/local-discovery.js#L278-L280

   ...but the duplicate connection is not yet in this list so it is never destroyed and the server will never close.

5. The old connection is closed and ["promotion" of the new connection completes][4]. It is never destroyed.

This caused [intermittent CI failures][ci] and should be fixed by this change.

[docs]: https://nodejs.org/docs/latest-v18.x/api/net.html#serverclosecallback
[ci]: https://github.com/digidem/mapeo-core-next/actions/runs/7996074898/job/21837815901
[2]: https://github.com/digidem/mapeo-core-next/blob/5e05e35200297435f4ebd33df2c5b1444ed4ef1b/src/discovery/local-discovery.js#L229-L232
[3]: https://github.com/digidem/mapeo-core-next/blob/5e05e35200297435f4ebd33df2c5b1444ed4ef1b/src/discovery/local-discovery.js#L278-L280
[4]: https://github.com/digidem/mapeo-core-next/blob/5e05e35200297435f4ebd33df2c5b1444ed4ef1b/src/discovery/local-discovery.js#L173


src/discovery/local-discovery.js
```diff
@@ -233,6 +233,16 @@ export class LocalDiscovery extends TypedEmitter {
...
     }
+
+    // Bail if the server has already stopped by this point. This should be
+    // rare but can happen during connection swaps if the new connection is
+    // "promoted" after the server's doors have already been closed.
+    if (!this.#server.listening) {
+      this.#log('server stopped, destroying connection %h', remotePublicKey)
+      conn.destroy()
+      return
+    }
```
EvanHahn, line 244: I'm confident[^hubris] in my understanding of the problem...less so in my solution. Do we want to do this if the server is not force-closed, for example?

[^hubris]: this might be hubris

gmaclennan (commented):
I have been unable to get my head deep enough into this code to fully understand the problem/solution. I think we are still missing a piece around gracefully closing connections to peers. Currently I think peers stay connected until the connection errors or they are force-terminated. Ideally what would happen is:

1. Peer starts intention to close connections (e.g. when going into the background)
2. Peer immediately stops accepting new connections
3. For each connection to a peer, if all data is "in sync", close the connection
4. Else, wait peers to be "in sync", then close the connection
5. If after a timeout, peers are still not "in sync", then force close the connection

This comment probably belongs somewhere else, just thought I'd initially drop it here since I think it's relevant to the logic here.

EvanHahn:
Moving to draft because we might delete a lot of this code.

PR #508, opened 2024-03-06
feat: send device info changes to connected peers

### LINKED ISSUE #502: Inform connected peers of name changes
## Description

@ErikSin brought this up when reviewing the front-end implementation of invites.

Currently when two devices connect over the local network, they share their "device name" as it is defined at the moment that they connect. The device name can be changed after the devices connect, and the connected device should be informed of this change. This is particularly relevant because of the order of operations in a real-life app:

1. User opens Mapeo for the first time
2. Device starts connecting to other Mapeo devices on the local network
3. Device connects and doesn't share a name because it's not defined
4. User enters device name in intro screen
5. Connected devices do not see the device as named until they disconnect and reconnect

## Tasks

- [ ] Send device name to connected peers when it changes
- [x] Emit a new `peers` event when a device receives a new name for a device.

EvanHahn:
_I recommend [reviewing this with whitespace changes disabled](https://github.com/digidem/mapeo-core-next/pull/508/files?w=1)._

Closes #502.

### DIFF
--- a/src/mapeo-manager.js
+++ b/src/mapeo-manager.js
@@ -257,10 +257,16 @@ export class MapeoManager extends TypedEmitter {
    */
   #replicate(noiseStream) {
     const replicationStream = this.#localPeers.connect(noiseStream)
-    Promise.all([this.getDeviceInfo(), openedNoiseSecretStream(noiseStream)])
-      .then(([{ name }, openedNoiseStream]) => {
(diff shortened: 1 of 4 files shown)


achou11 (commented):
one question to get some clarification on before approving but otherwise changes look good!

achou11 (approved):

PR #509, opened 2024-03-06
chore: update Undici to latest version

EvanHahn:
This updates Undici to v6.7.0.

This is a test-only dependency so I think this is safe to merge as long as tests pass.


achou11 (approved):
Just noting that #459 will move undici to a direct dependency - don't think it's a problem to upgrade but just wanted to surface that

PR #515, opened 2024-03-11
feat: Translation Api

### LINKED ISSUE #466: Translation API

tomasciccola:
this should close #466


tomasciccola:
I can't remember whats the rationale for omitting the actual translated value (called `message`) when hashing the doc. For now I'm using the whole doc for hashing, but there was probably a good reason for it

gmaclennan:
> I can't remember whats the rationale for omitting the actual translated value (called `message`) when hashing the doc. For now I'm using the whole doc for hashing, but there was probably a good reason for it

Yes the reason is so that when you edit the translated value, the `docId` does not change, and so that you can know the `docId` without knowing the translated value (when reading a translated value, when you have the other fields)

src/translation-api.js
```diff
@@ -0,0 +1,80 @@
...
+      .all()
+  }
+
+  /**
+   * @param {import('@mapeo/schema').TranslationValue} doc
+   */
+  index(doc) {
+    let translatedSchemas = this.#translatedLanguageCodeToSchemaNames.get(
+      doc.languageCode
+    )
```
gmaclennan, line 73:
```suggestion
    )
    if (!translatedSchemas) {
      translatedSchemas = new Map()
      this.#translatedLanguageCodeToSchemaNames.set(doc.languageCode, translatedSchemas)
    }
```

src/translation-api.js
```diff
@@ -0,0 +1,80 @@
...
+  async get(value) {
+    return (await this.dataType[kSelect]())
+      .where(
+        and(
+          eq(this.table.docIdRef, value.docIdRef),
+          eq(this.table.fieldRef, value.fieldRef),
+          eq(this.table.schemaNameRef, value.schemaNameRef),
+          eq(this.table.languageCode, value.languageCode),
+          // TODO: should we fallback to the first matching languageCode if regionCode doesn't match?
+          eq(this.table.regionCode, value.regionCode)
```
gmaclennan, line 60: yeah,  the code in [`dataType.#translate()`](https://github.com/digidem/mapeo-core-next/pull/473/files#diff-8ec3fd148a344d7225e902402078968d7097813f81c60db2fb161681b81c923dR219-R246) expects to get back all matching region codes. It also expects all fields to be returned - I think it's more efficient to return all fields and region codes in this `get()`, then parse them in memory.

However for this API standalone, maybe we can make `fieldRef` and `regionCode` optional, e.g.

```js
const filters = [
  eq(this.table.docIdRef, value.docIdRef),
  eq(this.table.schemaNameRef, value.schemaNameRef),
  eq(this.table.languageCode, value.languageCode)
]
if (value.fieldRef) {
  filters.push(eq(this.table.fieldRef, value.fieldRef))
}
if (value.regionCode) {
  filters.push(eq(this.table.regionCode, value.regionCode))
}
return (await this.dataType[kSelect]())
       .where(and.apply(null, filters))
```

src/translation-api.js
```diff
@@ -0,0 +1,80 @@
...
+  async put(value) {
+    // const { message, ...identifiers } = value
+    const docId = hashObject(value)
+    let existing
+    try {
+      existing = await this.dataType.getByDocId(docId)
+    } catch (e) {
+      existing = await this.dataType[kCreateWithDocId](docId, value)
+    }
+    if (existing) {
```
gmaclennan, line 37: `existing` will always be truthy here, maybe this should be in the `try` block?

src/mapeo-project.js
```diff
@@ -413,6 +426,17 @@ export class MapeoProject extends TypedEmitter {
...
         if (schemaName === 'projectSettings') {
           projectSettingsEntries.push(entry)
+        } else if (schemaName === 'translation') {
+          const doc = decode(entry.block, {
+            coreDiscoveryKey: entry.key,
+            index: entry.index,
+          })
+
+          // TODO: this is so that I can cast the doc as a `TranslationValue`
+          // without needing to decode the doc beforehand
```
gmaclennan, line 436: I think this is fine like this

src/translation-api.js
tomasciccola: The problem with this is that `getDocById` will throw if it doesn't find anything, instead of returning `null` or `undefined`

gmaclennan: Is this reply in the right place?

src/translation-api.js
tomasciccola: oops, yeah, you're right

src/translation-api.js
tomasciccola: how can `fieldRef` be optional? if that's the case then when requesting a translation for the key 'monitor-name' in a `Field`, you'll get every translated key instead of just what you want

gmaclennan: In the actual usage [dataType.#translate()](https://github.com/digidem/mapeo-core-next/pull/473/files#diff-8ec3fd148a344d7225e902402078968d7097813f81c60db2fb161681b81c923dR219-R246) you can see that in order to translate a record, we need to check whether each field has a translation. I thought it was more efficient to make a single query for all fields for a particular docId, since we will need all of them, rather than make a query for each field.

gmaclennan: It's because we need to check every field for every record for a translation, for every read of the database. The two strategies in the implementation to limit the performance impact of this are:

1. Index which datatypes have any translations to a particular languageCode
2. If, from the index, we do know that e.g. `preset` datatype has at least one translation to `es` languageCode, then we do one query for each document, by `docIdRef`, `schemaIdRef` and `languageCode`, and then translate any fields from the returned rows.

tomasciccola: Yeah, that makes a lot of sense

src/translation-api.js
tomasciccola: nop, it wasn't :|

tomasciccola:
I'm wondering how are we actually using the cache (when calling `translationApi.index(doc)`. If, lets say, I translate a `field` to spanish (by calling `translationApi.put(translation)`, the index would look like
```js
Map(1) { 'es' => Set(1) { 'field' } }
```
Which signals that basically we have fields translated to spanish. I would think that we would use this Map on subsequent calls to `translationApi.get()`, but I don't actually know how we would use that?? (I would imagine this would remove some filters from the sql query, but I don't actually know...)

tomasciccola:
Now that I think about it, shouldn't the  `#translatedLanguageCodeToSchemaNames`, should be smth like
```js
Map<languageCode,Map<schemaName,Map<fieldRef,message>>>
```
I know that it looks convoluted, but that way we're actually storing the translated values in memory that we can retrieve

gmaclennan:
> Now that I think about it, shouldn't the `#translatedLanguageCodeToSchemaNames`, should be smth like
> 
> ```js
> Map<languageCode,Map<schemaName,Map<fieldRef,message>>>
> ```
> 
> I know that it looks convoluted, but that way we're actually storing the translated values in memory that we can retrieve

In `translationApi.get()`:

```js
get(opts) {
  const docTypeIsTranslatedToLanguage = this.#translatedLanguageCodeToSchemaNames.get(opts.languageCode)?.has(opts.schemaNameRef)
  if (! docTypeIsTranslatedToLanguage) return []
```

The vast majority of records in the database will not have any translation. This provides a shortcut to avoid an unnecessary database query, with a fairly small memory footprint (the size of this map is limited to the number of languages the user has translated for, by the number of schema types they have translated, so fairly small).

tomasciccola:
owww ok yeah, that makes sense; its not to cache records, is to avoid queries for things we don't have translation for. I kinda knew that after our sync,  but I guess I just forgot

tomasciccola:
Shouldn't the `index` method be called every time we do a `put` ? So, in that sense, it should basically be a private method of the api? 
Is there a case where we want to `put` a translation but not index it??

tomasciccola:
I took the liberty of setting `index` as a private method and running it when calling `put`. There's smth that I may be missing, so that change could be reverted.
I'll probably create e2e tests for this, so that we can test creating translations for real records in a project (probably using the configs that we use for testing), but in terms of unit tests I dunno if there's stuff missing (there may be...)

gmaclennan:
You're missing that there is more than one user/writer in a project 🙂

tomasciccola:
> You're missing that there is more than one user/writer in a project 🙂

HEh! That's what I was missing. I'll revert that and update the tests

src/translation-api.js
```diff
@@ -0,0 +1,107 @@
...
+    const filters = [
+      eq(this.table.docIdRef, value.docIdRef),
+      eq(this.table.schemaNameRef, value.schemaNameRef),
+      eq(this.table.languageCode, value.languageCode),
+    ]
+    if (value.fieldRef) {
+      filters.push(eq(this.table.fieldRef, value.fieldRef))
+    }
+
+    if (value.regionCode) {
```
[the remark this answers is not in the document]
EvanHahn: I think so.

tomasciccola:
So, I think this is ready to review. Welcoming ideas on how to improve the e2e tests. I basically wrote manual translations on a fixture and translated a whole bunch of presets and tested over that. I could do the same thing with fields.
But this also may make sense to expand upon in the[ follow up PR](https://github.com/digidem/mapeo-core-next/pull/473)

EvanHahn:
I plan to review next week.

src/translation-api.js
```diff
@@ -0,0 +1,107 @@
...
+    } catch (e) {
+      existing = await this.dataType[kCreateWithDocId](docId, value)
+    }
+    return existing.docId
+  }
+
+  /**
+   * @param {import('type-fest').SetOptional<
+   * Omit<import('@mapeo/schema').TranslationValue,'schemaName' | 'message'>,
+   * 'fieldRef' | 'regionCode'>} value - We omit the schemaName and message, and we make fieldRef and RegionCode optional
```
EvanHahn, line 52: Nitpick: I think it's clear without this comment, and this comment isn't really for consumers of this function. I'd remove.

```suggestion
   * 'fieldRef' | 'regionCode'>} value
```

src/translation-api.js
```diff
@@ -0,0 +1,107 @@
...
+   */
+  async get(value) {
+    const docTypeIsTranslatedToLanguage =
+      this.#translatedLanguageCodeToSchemaNames
+        .get(value.languageCode)
```
EvanHahn, line 57: I think this will be empty on app startup, which means that we'll never be able to get docs that were added on a different launch. For example:

1. I start the app and add a translation, which gets indexed into the cache.
2. I restart the app. The cache is wiped.
3. I call `TranslationApi.prototype.get`. Because the cache is empty, we won't look up the document.

What if we just removed the cache entirely?

src/mapeo-project.js
```diff
@@ -420,6 +434,18 @@ export class MapeoProject extends TypedEmitter {
...
+          const doc = decode(entry.block, {
+            coreDiscoveryKey: entry.key,
+            index: entry.index,
+          })
+
+          // this is so that I can cast the doc as a `TranslationValue`
+          // without needing to decode the doc beforehand
+          if (doc.schemaName === 'translation') {
+            this.#translationApi.index(doc)
+          }
```
EvanHahn, line 447: 1. Don't we decode it on the lines above?
2. Shouldn't the schema name always be `translation`?

Could we import `assert` from `utils.js` and do this?

```suggestion
          assert(doc.schemaName === 'translation', 'expected a translation doc')
          this.#translationApi.index(doc)
```

EvanHahn (changes requested):
Looks good other than a few comments.

src/translation-api.js
```diff
@@ -0,0 +1,107 @@
...
+   *   typeof import('./schema/project.js').translationTable,
+   *   'translation',
+   *   import('@mapeo/schema').Translation,
+   *   import('@mapeo/schema').TranslationValue
+   * >}  opts.dataType
+   * @param {typeof import('./schema/project.js').translationTable} opts.table
+   */
+  constructor({ dataType, table }) {
+    this.dataType = dataType
+    this.table = table
```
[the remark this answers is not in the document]
tomasciccola: merged!

src/translation-api.js
tomasciccola: Yeah, what you say makes sense. I'm wondering how hard would it be to populate the cache when restarting the app. I _think_ it would be smth like calling `this.#datatype.getMany()` on the `TranslationApi` constructor and using that to populate the cache?  I would need to be able to set up tests differently to allow running them with a persistent db - so that I can test this specific thing - . Doing that may be a little more involved that what I'm thinking. Buut it may be worth it. wdyt?

EvanHahn: Is the cache primarily for performance? If so: how slow is it without the cache?

I wonder if we can remove it because the "no cache" performance is fine.

tomasciccola: I did a quick (and pretty non-exhaustive) perf test where we query 150_000 docs with non-existent `languageCodes`. With the cache it runs in 1.2 secs and without it it runs in 15.36 secs. 
On the other side I'm wondering how many combinations of `languageCode` and `schemaName` a sole user would have to justify having a cache.
I wonder what @gmaclennan thinks.

EvanHahn: I wonder if you could recoup some of that performance with a SQLite index...

gmaclennan: The idea of the cache is for read performance of records that _do not have_ translations. Currently the only docs that might have translations are presets and fields, and there will never be many of these. So the performance issue we are worried about is looking up and not finding translations for observations really. The index is minimal - I don't think it will ever be large - so maybe we could just store it as json somewhere, maybe in SQLite? It won't change much and will only need read once on app startup.

src/translation-api.js
```diff
@@ -0,0 +1,110 @@
...
+    let existing
+    try {
+      existing = await this.#dataType.getByDocId(docId)
+      if (existing) {
+        await this.#dataType.update(existing.versionId, value)
+      }
+    } catch (e) {
+      existing = await this.#dataType[kCreateWithDocId](docId, value)
+    }
+    return existing.docId
```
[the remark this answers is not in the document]
tomasciccola: You're right. I think I copied behavior from `iconApi` which returns `docId` (we should maybe change that also?)

src/translation-api.js
tomasciccola: In any reason, we would want to test the cache right? Which means having some sort of persistence on subsequent runs of a test? I think I'm going to need help with that.

gmaclennan: > If there's valid reasons for keeping the cache, and its going to be pretty minimal, shouldn't rebuilding it from the saved cores when the app starts be enough??

To rebuild it from saved cores, then we need to re-index all of the cores, which is an expensive operation. I think we need to persist it somehow, interested in @EvanHahn 's thoughts though.

gmaclennan: > In any reason, we would want to test the cache right? Which means having some sort of persistence on subsequent runs of a test? I think I'm going to need help with that.

Yes, and we actually don't have enough e2e tests that include closing the instance and then re-opening it. Maybe @EvanHahn can help with setting up some helpers for that.

tomasciccola: [this PR](https://github.com/digidem/mapeo-core-next/pull/580/files) adds a helper to have a persistent project between runs, until it is merged I commented the test and the population of the cache when instancing the `TranslationApi`([here](https://github.com/digidem/mapeo-core-next/blob/06da1a0cccab406cc10434c44ffe197b14fae428/src/translation-api.js#L30)). 
Regarding saving the cache on a JSON or creating an index in SQL lite, it may be worth it, but maybe not? (I should test how much time it takes to re-build the cache, but we know that this cache is not going to be pretty big)

tomasciccola:
The last commit adds a test of the cache after subsequent restarts. I needed to do two _not ideal_ things:
* Expose the cache through a getter, since we can't serialize a Symbol (so we can't pass it to a child process)
* I needed to wait 500ms to test the cache. I think this is because the instantiation of the translation api can't be awaited, so the promise that rebuilds the cache is run in parallel to it

src/translation-api.js
```diff
@@ -0,0 +1,120 @@
...
+    const docId = hashObject(identifiers)
+    try {
+      const doc = await this.#dataType.getByDocId(docId)
+      return await this.#dataType.update(doc.versionId, value)
+    } catch (e) {
+      // @ts-ignore how can this be improved (we can maybe set "useUnknownInCatchVariables": false)
+      if (e.message !== 'Not found')
+        throw new Error(`Error on translation ${e}`)
+      // TODO: throw if the error is different from 'Not found'
+      return await this.#dataType[kCreateWithDocId](docId, value)
```
EvanHahn, line 53: These lines have several issues, which (I think) I fixed in #590.

src/translation-api.js
```diff
@@ -0,0 +1,120 @@
...
+    }
+  }
+
+  /**
+   * @param {import('type-fest').SetOptional<
+   * Omit<import('@mapeo/schema').TranslationValue,'schemaName' | 'message'>,
+   * 'fieldRef' | 'regionCode'>} value
+   */
```
EvanHahn, line 61: Nitpick: This currently returns `any`. Could we add a return type?

```suggestion
   * @returns {Promise<import('@mapeo/schema').Translation[]>}
   */
```

EvanHahn (changes requested):

src/translation-api.js
```diff
@@ -0,0 +1,120 @@
...
+  async get(value) {
```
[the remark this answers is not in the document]
tomasciccola: Yeah, thanks for catching this!

EvanHahn (approved):
LGTM. Thanks for addressing all of my comments!

PR #517, opened 2024-03-14
chore: improve ExpectionAssertion type in Brittle TypeScript definitions

achou11:
Noticed a couple of minor issues:

- types were not handling providing a promise (note, not a function returning a promise) as the first argument
- was noticing some misleading "unnecessary awaits" warnings in my editor, even though the provided function was async. narrowed it down to needing to adjust the order of the overload so that it was ordered from most specific to least specific (i.e. `() => Promise<unknown>` is a subset of `() => unknown`)

### DIFF
--- a/types/brittle.d.ts
+++ b/types/brittle.d.ts
@@ -9,15 +9,28 @@ declare module 'brittle' {
   type AnyErrorConstructor = new () => Error
 
   interface ExceptionAssertion {
+    (
+      fn: Promise<unknown> | (() => Promise<unknown>),
+      message?: string
+    ): Promise<void>
+    (
+      fn: Promise<unknown> | (() => Promise<unknown>),
+      error?: RegExp | AnyErrorConstructor,
+      message?: string
+    ): Promise<void>
     (fn: () => unknown, message?: string): void
     (
       fn: () => unknown,
       error?: RegExp | AnyErrorConstructor,
       message?: string
     ): void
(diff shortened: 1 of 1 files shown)


types/brittle.d.ts
```diff
@@ -9,15 +9,28 @@ declare module 'brittle' {
   type AnyErrorConstructor = new () => Error
 
   interface ExceptionAssertion {
+    (
+      fn: Promise<unknown> | (() => Promise<unknown>),
```
achou11, line 13: technically `fn` is kind of a misleading parameter name in the case of providing a promise, but didn't think that creating another overload made sense vs using a union

types/brittle.d.ts
```diff
@@ -57,7 +64,10 @@ declare module 'brittle' {
 
   export interface TestInstance extends Assertion {
     plan(n: number): void
-    teardown(fn: () => unknown | Promise<unknown>, options?: { order?: number }): void
+    teardown(
+      fn: () => unknown | Promise<unknown>,
+      options?: { order?: number }
+    ): void
```
achou11, line 70: unintentional formatting fix - can revert if desired

EvanHahn: All good, love a formatting fix.

EvanHahn (approved):

EvanHahn:
Thanks for fixing this!

PR #518, opened 2024-03-14
fix: emit sync state on sync.{start,stop}

### LINKED ISSUE #510: Emit `sync-state` event after calling `SyncApi.start()` and `SyncApi.stop()`?
## Description

Been working with the sync api in application code and I'm noticing that calling the `start()` and `stop()` methods will not trigger a `sync-state` event if they meaningfully change the `state.data.syncing` (i.e. not in the idempotent case). The payload for that event is directly affected by mutations in those methods so I would expect the event to fire to notify listeners of the updated value.

The mutations of interest in each method:

https://github.com/digidem/mapeo-core-next/blob/30e637ba74168d106e17d7b3de349abd97f51c74/src/sync/sync-api.js#L137
https://github.com/digidem/mapeo-core-next/blob/30e637ba74168d106e17d7b3de349abd97f51c74/src/sync/sync-api.js#L125

Part of the payload is derived by the mutations above:

https://github.com/digidem/mapeo-core-next/blob/30e637ba74168d106e17d7b3de349abd97f51c74/src/sync/sync-api.js#L74

A more practical example of what I would expect:

```ts
// 1. Add listener
project.$sync.on('sync-state', (state) => {
  console.log(state.data.syncing)
})

// 2. Start/enable data sync
// `state.data.syncing` changes from false to true, should cause listener to log `true` 
project.$sync.start()

// Idempotent, should not emit event
project.$sync.start()

// 3. Stop/disable data sync
// `state.data.syncing` changes from true to false, should cause listener to log `false`
project.$sync.stop()

// Idempotent, should not emit event
project.$sync.stop()
```

Without this behavior, I have to implement some complicated code to set up a wrapped version of `start()` and `stop()` that eventually call `getState()` afterwords in order to get the most updated state and trigger a reactive change

gmaclennan:
fixes #510

### DIFF
--- a/test-e2e/sync.js
+++ b/test-e2e/sync.js
@@ -1,5 +1,6 @@
 // @ts-check
 import { test } from 'brittle'
+import { setTimeout as delay } from 'timers/promises'
 import {
   connectPeers,
   createManagers,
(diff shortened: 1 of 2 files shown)


test-e2e/sync.js
```diff
@@ -259,3 +260,40 @@ test('no sync capabilities === no namespaces sync apart from auth', async (t) =>
 
   await Promise.all(projects.map((p) => p.close()))
 })
+
+test('Sync state emitted when starting and stopping sync', async function (t) {
```
achou11, line 264: potential suggestion: add some assertions to test idempotency of `start()` and `stop()` i.e. events are not emitted for subsequent calls that do not meaningfully change sync state

maybe that's diving too much into internals though...

achou11 (approved):
Added a non-blocking suggestion but lgtm - thanks for the quick fix!

EvanHahn (approved):
Gave this a 30-second scan and it looks good to me. I suspect Andrew's review was more thorough but hoping my 👍🏻 helps!

PR #522, opened 2024-03-14
chore: replace onceSatisfied with p-event

EvanHahn:
Lets us remove a bunch of code!


gmaclennan (approved):
Yay to removing code!

PR #523, opened 2024-03-18
fix: sync state prior to full sync should be correct

gmaclennan:
Before starting sync of data, the sync state should accurately show the data that needs to be synced. There were bugs in the code that meant the "data to sync" was not accurate until the user had already started to sync the data. This PR fixes things so that after doing an "initial" sync, the sync state should accurately show how much data there is to be synced before the user starts syncing.

This PR fixes several things, because each fix revealed other bugs:

- Updates `RemoteBitfield` with changes from [upstream](https://github.com/holepunchto/hypercore/blob/fa23b5bcf5745dddaf607d92f8bccaa46a8642b1/lib/bitfield.js) to fix `findLast` and a few other things (data was not being added to the Quickbit index when it was inserted from a UInt32Array (as opposed to being set bit-by-bit).
- Fix `CoreSyncState` to correctly calculate `want` `wanted` `have` for cores that have not yet been replicated, based on preHave messages. We don't get the core length without replication, so we estimate it from the preHave messages. The length will be incorrect if the connected peer is missing blocks at the end of their core (because we use the last set bit in the bitfield for the length), but this is ok, it just means that `missing` might be inaccurate before a core is actually replicated.
- Fix to ignore remote states for blocked peers - without this, because a blocked peer potentially has data which will not sync, then sync state never considers sync to be complete for a blocked peer.
- Fix to test to ignore `missing` when comparing states of sync with a blocked peer, because different peers connected to a blocked peer can have different impressions of what is missing, based on whether they have tried to sync with them before blocking them.

(Note: I don't really like how this has been fixed by passing around peer sync controllers and making `deriveState` more complex because it relies on a more complicated state.... but all I can think of for now, so hopefully "good enough").

### DIFF
--- a/src/sync/core-sync-state.js
+++ b/src/sync/core-sync-state.js
@@ -140,13 +140,6 @@ export class CoreSyncState {
       this.#preHavesLength,
       peerState.preHavesBitfield.lastSet(start + bitfield.length * 32) + 1
     )
-    console.log(
-      'insertPreHaves',
-      peerId.slice(0, 5),
-      this.#core?.discoveryKey?.toString('hex').slice(0, 5),
-      bitfield.slice(0, 5),
-      this.#preHavesLength
-    )
     this.#update()
   }
 
@@ -164,6 +157,14 @@ export class CoreSyncState {
       peerState.setWantRange({ start, length })
     }
     this.#update()
+  }
+
+  /**
+   * @param {PeerId} peerId
+   */
+  addPeer(peerId) {
+    if (this.#remoteStates.has(peerId)) return
+    this.#remoteStates.set(peerId, new PeerState())
   }
 
   /**
@@ -201,12 +202,6 @@ export class CoreSyncState {
     // A peer can have a pre-emptive "have" bitfield received via an extension
     // message, but when the peer actually connects then we switch to the actual
     // bitfield from the peer object
-    // console.log(
-    //   'setting haves',
-    //   peerId.slice(0, 5),
-    //   this.#core?.discoveryKey?.toString('hex').slice(0, 5),
-    //   peer.remoteBitfield.firstUnset(0)
-    // )
     peerState.setHavesBitfield(peer.remoteBitfield)
     this.#update()
 
--- a/src/sync/core-sync-state.js
+++ b/src/sync/core-sync-state.js
@@ -58,6 +58,7 @@ export class CoreSyncState {
   #localState = new PeerState()
   /** @type {DerivedState | null} */
   #cachedState = null
+  #preHavesLength = 0
   #update
 
   /**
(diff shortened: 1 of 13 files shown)


tests/remote-bitfield.js
```diff
@@ -2,10 +2,24 @@
...
+  t.is(b.findLast(true, 2_000_000), 1_000_000)
+})
+
+test('remote bitfield - findLast, findFirst from insert', function (t) {
+  // Regression test for bug in RemoteBitfield which was not updating the index
+  // when inserting a new bitfield.
+  const b = new RemoteBitfield()
+
+  const size = 100
+  const arr = new Uint32Array(Array(size).fill(2 ** 32 - 1))
```
EvanHahn, line 20: Nitpick: slightly shorter:

```suggestion
  const arr = new Uint32Array(size).fill(2 ** 32 - 1)
```

test-e2e/sync.js
```diff
@@ -253,7 +255,17 @@ test('no sync capabilities === no namespaces sync apart from auth', async (t) =>
       t.is(inviteeState[ns].coreCount, 2)
       t.is(blockedState[ns].coreCount, 1)
     }
-    t.alike(invitorState[ns].localState, inviteeState[ns].localState)
+
+    // "Invitor" knows blocked peer is blocked from the start, so never connects
+    // and never creates a local copy of the blocked peer cores, but "Invitee"
+    // does connect initially, before it realized the peer is blocked, and
+    // creates a local copy of the blocked peers cores, but never downloads
```
EvanHahn, line 262: Nitpick:

```suggestion
    // creates a local copy of the blocked peer's cores, but never downloads
```

src/sync/core-sync-state.js
```diff
@@ -134,6 +147,10 @@ export class CoreSyncState {
   insertPreHaves(peerId, start, bitfield) {
     const peerState = this.#getPeerState(peerId)
     peerState.insertPreHaves(start, bitfield)
+    this.#preHavesLength = Math.max(
+      this.#preHavesLength,
+      peerState.preHavesBitfield.lastSet(start + bitfield.length * 32) + 1
```
EvanHahn, line 152: It looks like `32` was changed to `128` upstream. Just double checking: is this `32` okay here? (Broadly, I'm a little surprised that we need to do this multiplication at all, but I think it's fine.)

EvanHahn (approved):
LGTM other than a few small nitpicks.

src/sync/core-sync-state.js
gmaclennan: It's slightly confusing, in that here `bitfield` refers to a Uint32Array, rather than an instance of RemoteBitfield. For the core length, we need the length in bits (because one bit = one block in the bitfield), and since each item in the array is 32 bits (because it's a Uint32Array), then we need the `length * 32`.

tests/remote-bitfield.js
gmaclennan: oh cool, didn't realize this also had the fill() method.

PR #524, opened 2024-03-20
chore: remove unused cache code from SyncState

EvanHahn:
*This change is not urgent.*

This change should have no impact on functionality.

`SyncState` had some caching code for performance, but we never actually used it. This removes that "dormant" code.


EvanHahn:
After some discussion, I'm going to do the opposite of this: fix the caching.

PR #525, opened 2024-03-20
feat!: invite cancelations

### LINKED ISSUE #447: Track invites and allow cancelling invites

EvanHahn:
*I recommend reviewing this PR one commit at a time.*

This PR adds invite cancelations. Notable changes:

- `InviteCancel` RPC messages are now sent and received.
- `project.$member.invite()` takes an `AbortSignal`, used for cancelations. It replaces the timeout option.
- The `invite-removed` event now has a "reason" parameter which the frontend can use.

Closes #447.

### DIFF
--- a/src/member-api.js
+++ b/src/member-api.js
@@ -177,6 +177,10 @@ export class MemberApi extends TypedEmitter {
    * @param {AbortSignal} signal
    */
   async #sendInviteAndGetResponse(deviceId, invite, signal) {
+    const inviteAbortedError = new Error('Invite aborted')
+
+    if (signal.aborted) throw inviteAbortedError
+
     const abortController = new AbortController()
 
     const responsePromise =
@@ -207,7 +211,7 @@ export class MemberApi extends TypedEmitter {
       return await responsePromise
     } catch (err) {
       if (err instanceof Error && err.name === 'AbortError') {
-        throw new Error('Invite aborted')
+        throw inviteAbortedError
       } else {
         throw err
       }
--- a/src/member-api.js
+++ b/src/member-api.js
@@ -217,13 +217,14 @@ export class MemberApi extends TypedEmitter {
   }
 
   /**
-   * Cancel an outbound invite, if it exists. No-op if we weren't inviting this
-   * device.
+   * Attempt to cancel an outbound invite, if it exists.
+   *
+   * No-op if we weren't inviting this device.
    *
    * @param {string} deviceId
    * @returns {void}
    */
-  cancelInvite(deviceId) {
+  requestCancelInvite(deviceId) {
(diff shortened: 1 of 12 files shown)


EvanHahn:
Reworked this to avoid using `AbortSignal`, instead using a `cancelInvite()` API.

Doesn't change the fact that invites create a long-running Promise, but I think that's a separate change in any case.

src/member-api.js
```diff
@@ -187,26 +187,47 @@ export class MemberApi extends TypedEmitter {
...
 
+  /**
+   * Attempt to cancel an outbound invite, if it exists.
+   *
+   * No-op if we weren't inviting this device.
+   *
+   * @param {string} deviceId
+   * @returns {void}
+   */
+  requestCancelInvite(deviceId) {
```
gmaclennan, line 227: Suggestion: return a boolean indicating if the invite has been cancelled. Because an inprogress invite can be non-cancellable in its final stages, there is a window of time before `invite()` resolves when a user might call `requestCancelInvite()` and the invite might still resolve - having the boolean response might help client/front-end implementation.

src/member-api.js
```diff
@@ -162,22 +167,17 @@ export class MemberApi extends TypedEmitter {
...
-   * received within the timeout.
-   *
    * @param {string} deviceId
    * @param {Invite} invite
-   * @param {object} options
-   * @param {number} options.timeout
+   * @param {AbortSignal} signal
    */
-  async #sendInviteAndGetResponse(deviceId, invite, { timeout }) {
+  async #sendInviteAndGetResponse(deviceId, invite, signal) {
```
gmaclennan, line 179: maybe check if signal is already aborted before continuing?

gmaclennan (approved):
great, looks good, just a couple of non-blocking comments

src/member-api.js
```diff
@@ -162,22 +167,17 @@ export class MemberApi extends TypedEmitter {
...
     const abortController = new AbortController()
```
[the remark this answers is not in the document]
EvanHahn: I think so. When this function finishes, it hits the `finally` which should remove the `'invite-response'` listener. This shouldn't (necessarily) abort the `signal` argument.

src/member-api.js
EvanHahn: I like this idea but don't think we should implement it unless the frontend needs it.

src/member-api.js
EvanHahn: Done in c38b70823f1540dfce29e7276c4bdcc7b646ff04.

PR #530, opened 2024-03-28
chore: test jsonSchemaToDrizzleColumns

EvanHahn:
This is a test-only change (other than tweaking a comment).

Issue #150, opened 2023-08-08
Handling floating point precision in sqlite


gmaclennan:
## Description

Writing a floating point number (e.g. non-decimal) to sqlite and reading it again results in a different value due to floating point conversion errors. This is noticeable when storing lat/lon coordinates. We are currently using the SQLite **REAL** data type for storing lat and lon. More about how SQLite handles floating point here: https://www.sqlite.org/floatingpoint.html

This might also be related to encoding of floating point numbers in Protobuf, which is also involved in the round-trip of writing-reading data in Mapeo Core.

I think the issue is compounded by Javascript and Sqlite using different ways to store floating points.

It's not ideal that when storing data with lat/lon coordinates, the values that come back are different to those that went in. I wonder that maybe we switch to using an integer field for lat and lon, multiplying as needed for a fixed precision. e.g. we could support 6 decimal places of precision by storing lat / lon as `value * Math.pow(10,6)`.

I think this is probably preferable to getting different values back than what you write, which is not what people would expect.

Suggestions about the best way forwards?

achou11:
> I wonder that maybe we switch to using an integer field for lat and lon, multiplying as needed for a fixed precision. e.g. we could support 6 decimal places of precision by storing lat / lon as value * Math.pow(10,6)

I think this would make sense! A quick search online seems to confirm that this is a reasonable approach (albeit annoying)

> I think this is probably preferable to getting different values back than what you write, which is not what people would expect.

Agreed!

gmaclennan:
I'm thinking maybe 7 decimal places for lat/lon is good. This gives precision of at least 0.011112m (1.1cm) - with greater precision for longitude away from the equator. This is way more accurate than most GPS, which have a max precision of about 5m, however survey GPS and even [RTK](https://learn.sparkfun.com/tutorials/what-is-gps-rtk/all) can get down to millimeter precision (although the most common standard for surveys is Trimble devices, which have about 2-3cm precision I think).

The advantage of 7 decimal places is that we can fit them in a 32-bit signed integers (decoding 64-bit integers to Javascript can be complicated). Min, max of sint32 is -2,147,483,647 to 2,147,483,647.

7 decimal precision lat/lon is going to be -1,800,000,000 to 1,800,000,000

We could go 8 decimal places for millimeter precision, but I think this is overkill for Mapeo, given the overhead of then needing 64-bit integers.

gmaclennan:
[...]

PR #531, opened 2024-03-28
chore: save schema integers in INTEGER SQLite columns

EvanHahn:
_Depends on #530._

If we use the ["integer" JSON Schema type][0], we should save that as an INTEGER column, not a REAL column. Both will work for many numbers, but larger integers (like `2**63`) can only be safely stored in INTEGERs.  This should be a bit more reliable.

As of this patch, we don't have any integer types in our schemas, so this has no effect. However, that may change as we address [upcoming issues][1].

[0]: https://json-schema.org/understanding-json-schema/reference/numeric#integer
[1]: https://github.com/digidem/mapeo-core-next/issues/150


gmaclennan (approved):
must have missed that column type when implementing. Thanks.

PR #536, opened 2024-04-01
chore: shorten `getExpectedConfig`

EvanHahn:
This is a test-only change that should have no impact.

`config.presets()`, `config.fields()`, and `config.icons()` are all iterables that we can convert to arrays.


tomasciccola (approved):
merging!

PR #538, opened 2024-04-01
chore: update Drizzle

EvanHahn:
This updates drizzle-orm and drizzle-kit to the latest versions.

Notably, [a drizzle-orm change][0] required a workaround to avoid resolving a promise with something awaitable.

I think this is a useful change on its own, but it also blocks upgrading TypeScript to the latest version.

[0]: https://github.com/drizzle-team/drizzle-orm/commit/c063144dc08726cc15323582fe377210329e579e


src/datatype/index.js
```diff
@@ -239,12 +239,17 @@ export class DataType extends TypedEmitter {
     return this.getByDocId(docId)
   }
 
-  /**
-   * @param {Parameters<import('drizzle-orm/better-sqlite3').BetterSQLite3Database['select']>[0]} fields
-   */
```
EvanHahn, line 244: This parameter (1) isn't in our type definitions (2) isn't used, so I removed it for simplicity.

src/datatype/index.js
```diff
@@ -239,12 +239,17 @@ export class DataType extends TypedEmitter {
...
-    return this.#db.select(fields).from(this.#table)
+    const result = this.#db.select().from(this.#table)
+    // [The result of `from()` is awaitable.][0] We don't want this because we
+    // want to be able to await the result of this method and then call
+    // `.where()` on the result.
+    //
+    // As a workaround, we remove `.then` from the result.
+    //
+    // [0]: https://github.com/drizzle-team/drizzle-orm/commit/c063144dc08726cc15323582fe377210329e579e
+    return removeThen(result)
```
EvanHahn, line 252: I would love a better way of doing this...

achou11: spent a brief moment looking into this and there isn't really a great way from what I can tell (unless we want to change the way we write queries 😬)

do other promise-related fields need to be removed as well?

achou11 (approved):
Couple of non-blocking suggestions:

1. Update to latest drizzle-orm (v0.30.8)

2. Getting a deprecation warning about the `placeholder` export (think the notice was introduced somewhere in the v0.29.x):

    image

    Might be worth including an update to the import in this PR, as there are only a couple of callsites.

EvanHahn:
> Update to latest drizzle-orm (v0.30.8)

Done in 0f22b0d82e4810b310d2e7aa0ec48fd92b4dd889.

> Getting a deprecation warning about the `placeholder` export (think the notice was introduced somewhere in v0.29.x):

Fixed in 3a78872df776c31b890741e93043b61c236f846b.

src/datatype/index.js
EvanHahn: Good idea. Done in 5c249cd208e0ace06cf72f7638c28e6fc334e953.

Issue #540, opened 2024-04-01
Fix flaky "Creator can leave project if another coordinator exists" test


EvanHahn:
[The "Creator can leave project if another coordinator exists" test][0] is very flaky, especially on CI.

We should fix this.

[0]: https://github.com/digidem/mapeo-core-next/blob/d5bf57063f055314ea71945034dbe4987f4b926b/test-e2e/project-leave.js#L115

PR #541, opened 2024-04-01
chore: disable flaky project leave test

EvanHahn:
*This might be a controversial change. Feel free to close!*

We should fix this, but I think it's disruptive enough that we should disable it for now. See #540 for a task to properly fix this.


EvanHahn:
Fixed in eb7d1f8feb2866fd4707bd58cba6ab0247620eaa.

PR #543, opened 2024-04-04
Feat: default config as dependency

### LINKED ISSUE #496: remove bundled config from repo
## Description
For now we're bundling the default config (which actually is based from an incorrect repo) as a zip in the repo
The idea would be to remove the bundled config from the repo, add the mapeo-settings-builder as a devDep, have a build step which basically pulls the mapeo-default-config repo, builds it using mapeo-settings-builder and uses the resulting bundled config for tests.

## Tasks

- [x] remove default config
- [x] add mapeo-default-settings as devDep
- [ ] update tests to reflect the correct config

tomasciccola:
should close #496 . Depends on [this issue](https://github.com/digidem/mapeo-default-config/issues/39) to be fixed so that tests can pass. 

My approach to vendoring the config was:
1. add `@mapeo/default-config` as a devDependency
1. create a script in `scripts/config.js` that
    1. looks for the default config file in `@mapeo/default` by parsing its package.json to get the version (so that we don't hardcode the filename)
    2. copies it into `tests/fixtures/config/`
2. add that script as `build:config` and add it to `test` 

I'm not sure if the default config should be published with core or not (I kinda remember that it shouldn't and that the default config should be a dependency of the frontend separately...)


src/config-import.js
```diff
@@ -108,14 +108,14 @@ export async function readConfig(configPath) {
     *fields() {
       const { fields } = presetsFile
       for (const [name, field] of Object.entries(fields)) {
-        if (!isRecord(field) || !hasOwn(field, 'key')) {
+        // is the hasOwn 'tagKey' check necessary?
+        if (!isRecord(field) || !hasOwn(field, 'tagKey')) {
```
EvanHahn, line 112: I don't think so. It's handled by `validate` below.

```suggestion
        if (!isRecord(field)) {
```

EvanHahn (changes requested):
LGTM other than my comments.

EvanHahn (approved):
LGTM once we use a "real" `@mapeo/default-config`.

gmaclennan (approved):
I made a couple of changes and published @mapeo/default-config, so this is ready to merge.

- When reading dependencies from disk, it's best practice to avoid building the disk path e.g. `node_modules/module-name`, because there is no guarantee that a dependency is always in the same location, e.g. when this module is installed as a dependency, then any dependencies of this module are normally installed directly in the parent `node_modules` folder, rather than in the `node_modules` of this package. As written it probably would have been ok to use what was here, because it was only a dev dep, but to avoid problems in the future I've changed it to use `require.resolve`.
- I changed the artifact (built config) filename published by `@mapeo/default-config`, to  remove the version name from the filename. No need to make life difficult for ourselves.
- I made the artifact the `main` of `@mapeo/default-config`, so that we don't need to use the filename at all here - any `require` of `@mapeo/default-config` will resolve to the build config file.

EvanHahn (approved):

tomasciccola:
>     * I changed the artifact (built config) filename published by `@mapeo/default-config`, to  remove the version name from the filename. No need to make life difficult for ourselves.
:laughing: 

>     * I made the artifact the `main` of `@mapeo/default-config`, so that we don't need to use the filename at all here - any `require` of `@mapeo/default-config` will resolve to the build config file.

Hey this is so cool, I didn't know you could do that with arbitrary assets!

PR #550, opened 2024-04-11
feat!: frontend should handle peer discovery

### LINKED ISSUE #474: Improve local device discovery on different networks
In our testing, we've found some issues where peers aren't discovered properly. We should fix that.

(This issue isn't super fleshed out because we had issues on particular wifi networks, and it's hard to describe those in a useful way here.)

EvanHahn:
_I recommend reviewing this one commit at a time._

Our peer discovery system is unreliable on some networks. We'd like to replace it with [a native local service discovery system][react-native-zeroconf].

To do this, the frontend needs to tell the backend about peers it finds. In other words, we need to "invert" peer discovery.

Here's pseudocode for what the frontend needs to do:

```javascript
const { name, port } = await mapeoApi.startLocalPeerDiscoveryServer()
zeroconf.publishService({ name, port })
zeroconf.on('resolved', (peer) => {
  mapeoApi.connectPeer(peer)
})
```

This implements that by removing our mDNS system and replacing it with these public APIs.

Closes #474.

[react-native-zeroconf]: https://www.npmjs.com/package/react-native-zeroconf


src/discovery/local-discovery.js
```diff
@@ -74,29 +66,29 @@ export class LocalDiscovery extends TypedEmitter {
...
+   * @param {object} peer
+   * @param {string} peer.address
+   * @param {number} peer.port
+   * @param {string} peer.name
+   * @returns {void}
    */
-  #handleServiceUp({ address, port, name }) {
+  connectPeer({ address, port, name }) {
+    if (this.#name === name) return
     this.#log('serviceUp', name.slice(0, 7), address, port)
```
achou11, line 92: nit: rename the log here?

src/discovery/local-discovery.js
```diff
@@ -25,10 +25,10 @@ export const ERR_DUPLICATE = 'Duplicate connection'
  */
 export class LocalDiscovery extends TypedEmitter {
   #identityKeypair
+  #name = randomBytes(8).toString('hex')
```
achou11, line 28: nit: do you think it will it ever be useful to provide a custom name? can consider a constructor opt for this if so

achou11 (approved):
changes seem good to me!

src/discovery/local-discovery.js
EvanHahn: I think it might be useful in future but not right now, so I'm going to skip it for now.

PR #552, opened 2024-04-11
chore: testing improvements

EvanHahn:
This change makes several test-only improvements:

- `prettier --check` should be run. (I had to reconfigure Prettier a bit, and format `CHANGELOG.md`.)

- ESLint should be run with the `--cache` option.

- Test code should be run in parallel, for performance. _⚠️ This is controversial!_

- `lint-staged` should run Prettier on all files, not just some extensions.

- `lint-staged` should only check files, not fix them. This addresses a
  race condition (search [lint-staged's docs for "race condition"][0]).

[0]: https://www.npmjs.com/package/lint-staged


EvanHahn:
Made a mistake here. I also think it's better to split this into smaller PRs, which I'll do now.

PR #568, opened 2024-04-16
chore: use `size` helper in config import tests

EvanHahn:
Iterpal's `size` function can save us a few lines and avoid a few `eslint-disable`s.


tomasciccola (approved):
what a self plug!
JK, iterpal is neat!

EvanHahn:
😈

PR #569, opened 2024-04-16
chore: update TypeScript

EvanHahn:
This updates TypeScript to its latest version, 5.4.5. It also required updating `typedoc`.

As far as I could tell, this made no meaningful changes to `dist/` and should be safe to do.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -96,9 +96,9 @@
         "streamx": "^2.15.1",
         "tempy": "^3.1.0",
         "ts-proto": "^1.156.7",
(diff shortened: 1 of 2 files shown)


achou11 (approved):
confirmed locally that the changes in the output for `npm run:build-types` are not significant

potential follow-up task: remove `typedoc` and anything related to it. think it'd make maintenance in the short-term slightly easier and also allows us to work with a clean slate if/when we want to introduce it again

EvanHahn:
Removed Typedoc in #574.

PR #571, opened 2024-04-16
chore: invites should use special invite ID, not project public key

### LINKED ISSUE #559: Invites shouldn't use project public key, but instead project invite ID

EvanHahn:
A sophisticated bad actor who knows a project's public ID could spam invites to determine whether someone was in a project.

Now, they need the harder-to-get "project invite ID", which is a separate generated value.

Closes #559.


src/mapeo-manager.js
```diff
@@ -538,10 +547,10 @@ export class MapeoManager extends TypedEmitter {
    * @param {{ waitForSync?: boolean }} [opts] For internal use in tests, set opts.waitForSync = false to not wait for sync during addProject()
    * @returns {Promise<string>}
    */
-  async addProject(
+  addProject = async (
```
gmaclennan, line 550: Any reason for binding this method to `this`?

src/utils.js
```diff
@@ -1,8 +1,15 @@
 import b4a from 'b4a'
+import sodium from 'sodium-universal'
 import { keyToPublicId } from '@mapeo/crypto'
 import { createHash } from 'node:crypto'
 import stableStringify from 'json-stable-stringify'
 
+const PROJECT_INVITE_ID_SALT = Buffer.from(
```
gmaclennan, line 7: I have previously used human-readable words (as a buffer) for this. This is based on what I have seen Hypercore do. I _think_ it's just as secure, and perhaps easier to read / check it hasn't changed.

src/utils.js
```diff
@@ -162,6 +169,17 @@ export function projectKeyToPublicId(projectKey) {
   return keyToPublicId(projectKey)
 }
 
+/**
+ * Generate an invite ID from a project key
+ * @param {Readonly<Buffer>} projectKey
+ * @returns {Buffer}
+ */
+export function projectKeyToProjectInviteId(projectKey) {
```
gmaclennan, line 177: I think this belongs in @mapeo/crypto, so that we can version it separately. We can move it in a follow-up PR.

gmaclennan (approved):
Looks ok to me. A surprising amount that needed changed! I got bored and didn't completely read the changes to the tests, but the rest looks ok with a couple of minor comments.

src/mapeo-manager.js
EvanHahn: Yes, this line above:

https://github.com/digidem/mapeo-core-next/blob/dd18fea272ec2d0d1b09f4e006f8a190f9d067cf/src/mapeo-manager.js#L163

Happy to change it back if you like.

src/utils.js
EvanHahn: Done in bc5714357f9d6e9259ae4d80e8045665ccbc2779.

src/utils.js
EvanHahn: Done in <https://github.com/digidem/mapeo-crypto/pull/26>.

PR #574, opened 2024-04-17
docs: remove Typedoc

EvanHahn:
We haven't kept this up to date, so let's remove it. We can add it later as needed.

`git grep -i typedoc` only returns results in `package-lock.json` for dependencies.


EvanHahn:
goodnight sweet prince

PR #580, opened 2024-04-23
chore: add `createPersistentProject` helper for e2e tests

tomasciccola:
This would help with testing with stuff related to persitent storage.

the function needs a path, and creates a project when run the first time. It will also saved the created `projectId` on a file in the passed path. On subsequent runs it will load the `projectId` from the path

### DIFF
--- a/test-e2e/project-crud.js
+++ b/test-e2e/project-crud.js
@@ -88,7 +88,7 @@ function getUpdateFixture(value) {
 const CREATE_COUNT = 100
 
 test('CRUD operations', async (t) => {
-  const manager = createManager('device0', t)
+  const manager = createManager({ seed: 'device0', t })
 
   for (const value of fixtures) {
     const { schemaName } = value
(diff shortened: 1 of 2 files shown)


gmaclennan:
I'm not sure this is quite the right abstraction for the testing needs. It repeats code in other helpers there's not an easy way to use it for invitees. Also we don't need to persist the projectId to disk for tests, we just need to persist it between instances of Manager. How about modifying the existing `createManager` function to accept `dbFolder` and `coreStorage` options, then you can use it like this in a test (psuedo code):

```js
const createRAM = RAM.reusable()
const dbFolder = createTempDir()
const seed = randomBytes(16)

const manager = createManager({ seed, dbFolder, coreStorage: createRAM })
const projectId = manager.createProject({ name })
const project = manager.getProject(projectId)

// add some data to the project

project.close()
manager.close()

const newManager = createManager({ seed, dbFolder, coreStorage: createRAM })
const project = manager.getProject(projectId)

// check the data added to the project above is in the project (e.g. data is persisted)

```

EvanHahn:
See #581 for an alternate (weird) approach.

tomasciccola:
closing this PR in favor or #581

PR #581, opened 2024-04-23
chore: test re-loading manager with the same data

EvanHahn:
This test-only change creates a project. Then, it makes a new manager with the same files and asserts that the project is still there.

Because the only way to close a manager is to close the Node process, this test adds a bunch of scaffolding to do that.

### DIFF
--- a/test-e2e/utils.js
+++ b/test-e2e/utils.js
@@ -3,6 +3,11 @@ import sodium from 'sodium-universal'
 import RAM from 'random-access-memory'
 import Fastify from 'fastify'
 import { arrayFrom } from 'iterpal'
+import * as path from 'node:path'
+import { fork } from 'node:child_process'
+import { createRequire } from 'node:module'
+import { fileURLToPath } from 'node:url'
+import * as v8 from 'node:v8'
 
 import { MapeoManager } from '../src/index.js'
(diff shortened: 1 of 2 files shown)


test-e2e/utils.js
```diff
@@ -215,6 +220,179 @@ export function createManager(seed, t, deviceType) {
...
+    const argsBuffer = Buffer.from(${s(argsHex)}, 'hex')
+    const args = v8.deserialize(argsBuffer)
+
+    const result = await fn(manager, ...args)
+
+    process.send(result)
+    `
+
+    const result = temporaryFile({ extension: 'mjs' })
+    this.#t.teardown(() => fsPromises.rm(result, { maxRetries: 2 }))
```
tomasciccola, line 353: only nit, why is there two calls to `t.teardown`?

tomasciccola (approved):

test-e2e/utils.js
EvanHahn: `dbFolder` and `coreStorage` are created once per `ManagerCustodian`. These JS files are created once per call to `withManagerInSeparateProcess`.

In my test, for example, `dbFolder` is created once (and therefore needs to be removed once) and the JS file is created twice (and therefore needs to be removed twice).

Does that answer your question?

PR #588, opened 2024-04-25
add typedoc, test doc generation

### LINKED ISSUE #586: Set typedoc to generate documentation from types
## Tasks

- [ ]

tomasciccola:
This should close #586 

In order for this documentation to be useful, it would be nice to know expectations for this (what things do we think are necessary to make this useful?).
For now I:
* set up a plugin to render it as markdown
* commit doc files so they can be viewed directly here


EvanHahn:
What if we could build these to `digidem.github.io/mapeo-core-next` using CI, rather than committing them to the `doc/` folder?

tomasciccola:
> What if we could build these to `digidem.github.io/mapeo-core-next` using CI, rather than committing them to the `doc/` folder?

Yeah, that could be an option. I really don't have a strong opinion on this. 
Do you see a disadvantage on committing docs to the repo? 
Having a dedicated page for typedoc kinda makes sense. But I also like having documentation in markdown directly on the repo.
A middle ground could be:
* commit docs on the repo as markdown
* have that same documentation published as gh-pages (this may make sense to set up using CI, since we need to generate docs as html, not md)

keen on other's opinion @achou11 @gmaclennan

achou11:
I was writing a decently long response to answer the questsions here but we just discussed this over a call, so will document what was agreed instead:

- Stick to committing the generated docs to the repo and deploying the directory using repo-specific github pages feature. we want to avoid touching the org-specific github pages because there's a [preexisting site](http://lab.digital-democracy.org/) there ([repo](https://github.com/digidem/digidem.github.io)).
- Adjust the configuration for the generator so that it only generates documentation for _exported_ APIs. Having all of the internal stuff is not as useful in the short-term for us
- Figure out whether we want docs to be updated per PR/commit or if there's some kind of automation we use to update the docs after PRs are merged in. Would be nice to set up a workflow that can do provide a "preview" url (similar to Netlify, Vercel, etc) for a PR.

tomasciccola:
Following andrews proposal I've basically set up documentation so that only public things gets docgenerated and commited to the repo. On the other hand I've set up a workflow so that we can publish an html version of documentation to ghpages. 
This is now deployed [here](http://lab.digital-democracy.org/mapeo-core-next/). This is a custom domain set up for the repo. I think the idea is not to hosted there, but on the default url (so, smth like mapeo-core-next.githab.pages.io ? ). 
But I can't seem to find a way to ignore a custom domain for a specific page. 
I'm wondering if we can delete the custom domain setup for this repo (I think so since I don't think is being use for anything else in this repo) so it can be deployed to the default url.
Regarding WHEN to generate the docs. For now is set up so that only updates on the main branch deploy the docs (which I think it makes sense?). After merging this PR I need to remove a protection rule so that this branch can also deploy the page.
The thing that I may need some help setting up is the preview of the page. What I'm thinking is that this preview should be accessible on a PR that is going to be merged to main. For that I may need to generate a token (gonna look at this a little bit more but that seems to be the case...)

tomasciccola:
Also, along side `MapeoManager` (which is kinda the the thing we want to generate docs for) there are some other stuff exported mostly related to fastify. We can explicitly ignore those classes (ussing a @hidden tag in the jsdoc) or just leave them there...

EvanHahn:
This is looking good!

> I'm wondering if we can delete the custom domain setup for this repo (I think so since I don't think is being use for anything else in this repo) so it can be deployed to the default url.

Let me know if you want me to do this.

>  Regarding WHEN to generate the docs. For now is set up so that only updates on the main branch deploy the docs (which I think it makes sense?). After merging this PR I need to remove a protection rule so that this branch can also deploy the page.

I think it makes sense to generate every time `main` changes.

> After merging this PR I need to remove a protection rule so that this branch can also deploy the page.

I wonder if we can disable the branch protection _only_ for the GitHub Actions bot. I see this in the settings:

> The thing that I may need some help setting up is the preview of the page.

I think this can be done in a followup PR.

> Also, along side `MapeoManager` (which is kinda the the thing we want to generate docs for) there are some other stuff exported mostly related to fastify. We can explicitly ignore those classes (ussing a `@hidden` tag in the jsdoc) or just leave them there...

I also think we should do this in a followup.

achou11:
not sure if disabling the custom domain for this repo will be trivial/possible based on what's described here:

https://docs.github.com/en/pages/configuring-a-custom-domain-for-your-github-pages-site/about-custom-domains-and-github-pages#using-a-custom-domain-across-multiple-repositories

we have the custom domain set up on our _org_ , which means that all of the projects are subject to it. in order to override it, we need to point to a new custom domain, but I don't think we can "revert" it to the project-specific default of digidem.github.io/mapeo-core-next

achou11:
Gregor mentioned that he's set up https://core.comapeo.app/ to point to GH pages so can use that for this project

tomasciccola:
> I wonder if we can disable the branch protection _only_ for the GitHub Actions bot. I see this in the settings:

doesn't seeems like it, it won't find the Github actions bot...

EvanHahn (changes requested):
1. How does `docs/api/md/` get automatically updated? How do we make sure the docs stay up-to-date?

1. There are a lot of links to source code. For example:

   https://github.com/digidem/mapeo-core-next/blob/518de6dcd6dbaf2011bdcd13651f82d1f74ab5cc/docs/api/md/classes/MapeoManager.md?plain=1#L69

   These links reference a Git commit ID. This means that every commit will update all these links. Maybe we could just have these update when releasing a new version of the module?

1. Is there a way to prevent all the internal files from being included?

1. A bunch of files seem to reference `/home/szgy/src/dd/`. When I run them on my machine, they reference _my_ working directory. Is there a way to normalize this?

tomasciccola:
> 1. How does `docs/api/md/` get automatically updated? How do we make sure the docs stay up-to-date?
> 
>     2. There are a lot of links to source code. For example:
>        https://github.com/digidem/mapeo-core-next/blob/518de6dcd6dbaf2011bdcd13651f82d1f74ab5cc/docs/api/md/classes/MapeoManager.md?plain=1#L69
>        
>        These links reference a Git commit ID. This means that every commit will update all these links. Maybe we could just have these update when releasing a new version of the module?

I think an interesting thing to see is the possibility of, when merging something to main or when publishing a new version, a flow is run that updates the repo markdown documentation.

>     3. Is there a way to prevent all the internal files from being included?

Currently I'm using a plugin `typedoc-plugin-missing-exports` that basically includes files that are referenced by a public file. This makes documentation more useful (i.e., you can see `MapeoProject` documentation now). The plugin doesn't have any options the allow smth like defining how deep in the dep chain it goes, so it kinda includes everything that is referenced...

>     4. A bunch of files seem to reference `/home/szgy/src/dd/`. When I run them on my machine, they reference _my_ working directory. Is there a way to normalize this?

Haven't found a way of fixing this yet. The files that reference an absolute path are kinda random from what I've seen

EvanHahn:
> I think an interesting thing to see is the possibility of, when merging something to main or when publishing a new version, a flow is run that updates the repo markdown documentation.

I think it'd be cool to do this when we publish a new version, probably as part of our existing release script (see `.github/workflows/release.yml`).

> Do you think having links to the source code makes sense? Because we can also disable them...

I think it's probably fine to disable them for now.

> Currently I'm using a plugin typedoc-plugin-missing-exports that basically includes files that are referenced by a public file. This makes documentation more useful (i.e., you can see MapeoProject documentation now). The plugin doesn't have any options the allow smth like defining how deep in the dep chain it goes, so it kinda includes everything that is referenced...

:thumbsup: Would be cool to fix this somehow in the future, not sure how.

tomasciccola:
> I think it'd be cool to do this when we publish a new version, probably as part of our existing release script (see `.github/workflows/release.yml`).
I've added a flow to `release.yml` to basically do this by using git and commiting the files. I've also added it to the docs.yml flow so that we can test without needing to do a release (will probably remove that step before merging though...)

> > Currently I'm using a plugin typedoc-plugin-missing-exports that basically includes files that are referenced by a public file. This makes documentation more useful (i.e., you can see MapeoProject documentation now). The plugin doesn't have any options the allow smth like defining how deep in the dep chain it goes, so it kinda includes everything that is referenced...
> 
> 👍 Would be cool to fix this somehow in the future, not sure how.
Yeah, me neither, same with all the internal files that are named with an absolute path...

tomasciccola:
Things to pay attention to:
* Security implications of giving gh-action permission to write the repo (can this be avoided by setting a different git user - like one of us - in the action itself?)
* improving the flow by using additional parameters ([see](https://github.com/stefanzweifel/git-auto-commit-action))

EvanHahn:
I'll plan to review tomorrow.

.github/workflows/release.yml
```diff
@@ -32,6 +32,15 @@ jobs:
         uses: actions/setup-node@v3
         with:
           node-version: 18
+      - name: Publish docs api to repo
+        run: |
+          npm ci
+          npm run doc
+          git status
+          git add docs/api/md
+          git commit -am "updated api docs"
```
EvanHahn, line 41: Two nitpicks:

1. Can we move this to the conventional commits style?
2. `-a` adds all files, which I think we'd rather not do.

```suggestion
          git commit -m "docs: update API docs"
```

.github/workflows/docs.yml
```diff
@@ -0,0 +1,44 @@
...
+          node-version: ${{ matrix.node-version }}
+      - name: Generate Api docs
+        run: |
+          npm ci
+          npm run doc
+      - name: Push updated markdown docs
+        uses: stefanzweifel/git-auto-commit-action@v5
+        with:
+          commit_message: 'Update api markdown docs'
+          file_pattern: 'docs/api/md/*'
```
EvanHahn, line 34: Do we want to update the Markdown docs after every commit to `main`? I feel like...maybe not? I like that you've added it as part of the release task.

EvanHahn (changes requested):

.github/workflows/release.yml
tomasciccola: Yeah, this became out of sync with the actual flow that uses an external action that actually commits the file (see the docs.yml flow). I'll set the message so that it uses conventional commits

.github/workflows/docs.yml
EvanHahn: Would you mind removing this, then?

.github/workflows/release.yml
```diff
@@ -32,6 +32,18 @@ jobs:
...
+      - name: Set git user
+        uses: fregante/setup-git-user@v1
+      - name: Generate Api docs
+        run: |
+          npm ci
+          npm run doc
+      - name: Push updated markdown docs
+        uses: stefanzweifel/git-auto-commit-action@v5
+        with:
+          commit_message: 'feat: update api markdown docs'
```
EvanHahn, line 45: Nit: I think this should be `docs`, not `feat`.

```suggestion
          commit_message: 'docs: update API docs'
```

package.json
```diff
@@ -47,6 +49,9 @@
       "node": true
     },
     "extends": "eslint:recommended",
+    "ignorePatterns": [
+      "docs/*"
+    ],
```
EvanHahn, line 54: I think we already have this down below, so we can remove this.

```suggestion
```

EvanHahn (approved):
LGTM other than my two small nitpicks.

PR #590, opened 2024-04-25
chore: use `instanceof`, not message, for "not found" error

EvanHahn:
Depends on #589.


tests/data-type.js
```diff
@@ -141,6 +142,11 @@ test('test validity of `createdBy` field from another peer', async (t) => {
   await destroy()
 })
 
+test('getByDocId() throws if no document exists with that ID', async (t) => {
+  const { dataType } = await testenv({ projectKey: randomBytes(32) })
+  await t.exception(() => dataType.getByDocId('foo bar'), NotFoundError)
+})
```
EvanHahn, line 148: This test isn't strictly necessary as part of this work, but I think it's useful to test this functionality in isolation.

tomasciccola (approved):

PR #593, opened 2024-04-30
chore: fix type errors in NamespaceSyncState tests

EvanHahn:
This test-only change fixes a couple of small type errors in our tests for `NamespaceSyncState`.


tests/sync/namespace-sync-state.js
```diff
@@ -190,12 +190,12 @@ test('replicate with updating data', async function (t) {
   for (const key of cm1Keys) {
     if (key.equals(writer1.core.key)) continue
     const core = cm1.getCoreByKey(key)
-    core.download({ live: true, start: 0, end: -1 })
+    core?.download({ start: 0, end: -1 })
```
EvanHahn, line 193: Best I could tell, `live` was not used in Hypercore at all.

PR #594, opened 2024-04-30
fix: update pre-haves as they change

### LINKED ISSUE #582: Adding data after a peer connects doesn't show up in sync state
If you add observations after a peer connects, the sync state will not update to show that there is data to sync. The sync state is static from the moment of connection.

EvanHahn:
In the following scenario:

1. Alice and Bob connect but do not start sync.
2. Alice adds an observation.

Bob should see that there is something new to sync.

Previously, this wouldn't happen. Now, when a writer core is appended to or cleared, we send new pre-have messages.

Closes #582.

### DIFF
--- a/src/core-manager/index.js
+++ b/src/core-manager/index.js
@@ -164,7 +164,7 @@ export class CoreManager extends TypedEmitter {
     })
 
     this.#creatorCore.on('peer-add', (peer) => {
-      this.#sendHaves(peer)
+      this.#sendHaves(peer, this.#coreIndex)
     })
     this.#creatorCore.on('peer-remove', (peer) => {
       // When a peer is removed we clean up any unanswered key requests, so that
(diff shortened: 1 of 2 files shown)


src/core-manager/index.js
```diff
@@ -304,12 +304,29 @@ export class CoreManager extends TypedEmitter {
       peer._maybeWant(0, core.length)
     })
 
-    // A non-writer core will emit 'append' when its length is updated from the
-    // initial sync with a peer, and we will not have sent a "maybe want" for
-    // this range, so we need to do it now. Subsequent appends are propogated
-    // (more efficiently) via range broadcasts, so we only need to listen to the
-    // first append.
```
EvanHahn, line 311: I moved this comment down and re-formatted it so no line was longer than 80 characters; I didn't change it other than fixing a typo.

src/core-manager/index.js
```diff
@@ -304,12 +304,29 @@ export class CoreManager extends TypedEmitter {
...
+      // Tell connected peers, who we aren't necessarily syncing with, about
+      // what we just added or cleared. Hypercore doesn't emit anything when
+      // clearing, so we patch it in.
+      core.on('append', sendHaves)
+      const originalClear = core.clear
+      core.clear = function clear() {
+        const result = originalClear.apply(this, /** @type {any} */ (arguments))
+        result.then(sendHaves)
+        return result
+      }
```
EvanHahn, line 323: I considered patching Hypercore itself, which doesn't feel meaningfully different. I don't think Hypercore offers a way to address this in a better way, even on newer versions.

EvanHahn:
Merging without review. I'll ask @gmaclennan to review soon.

PR #602, opened 2024-05-02
feat: config import translations

### LINKED ISSUE #587: Import translations when importing a configuration
## Tasks

- [ ]

tomasciccola:
May fix #587


tomasciccola:
I've basically followed the pattern with the `*fields` and `*preset`iterators. 
I had to change the preset iterator so that it also yields its name (which is actually the object key) since that is used in translations to reference it (so translations don't use the `name` field in presets but the key itself; i.e. "Boundary Line" is actually "boundary-line")

tomasciccola:
Also, keen to read alternatives to having 4 nested loops. Using a reducer is an alternative but I don't think it'll necessarily be clearer code...

tomasciccola:
there's a type of translation thats always empty called `category`. I don't know the use of it, but I'm skipping it - for now - and only importing 'fields' and 'presets'. We may want to update the default config (and config builder) in the future to remove this..

tests/config-import.js
```diff
@@ -212,6 +212,22 @@ test('config import - presets', async (t) => {
   t.is(config.warnings.length, 0, `no warnings on the file`)
 })
 
+test('config import - translations', async (t) => {
+  const config = await readConfig(defaultConfigPath)
+  for (let { _, value: translationValue } of config.translations()) {
```
EvanHahn, line 217: Two small nitpicks: (1) we don't need `_` (2) we should use `const` instead of `let` because we aren't changing it.

```suggestion
  for (const { value: translationValue } of config.translations()) {
```

tests/config-import.js
```diff
@@ -212,6 +212,22 @@ test('config import - presets', async (t) => {
...
+    t.is(
+      translationValue.schemaName,
+      'translation',
+      `schemaName is 'translation'`
+    )
+    t.ok(
+      translationValue.schemaNameRef === 'presets' || 'fields',
```
EvanHahn, line 224: What I think you meant:

```suggestion
      translationValue.schemaNameRef === 'presets' ||
        translationValue.schemaNameRef === 'fields',
```

As written, this assertion will always succeed. If `translationValue.schemaNameRef === 'presets'`, the assertion will pass. Otherwise, the value of the expression will be `'fields'`, which is always truthy.

src/mapeo-project.js
```diff
@@ -733,24 +735,55 @@ export class MapeoProject extends TypedEmitter {
...
+    await Promise.all(presetPromises)
+
+    const translationPromises = []
+    for (let { name, value } of config.translations()) {
+      let docIdRef
+      if (value.schemaNameRef === 'fields') {
+        docIdRef = fieldNameToId.get(name)
+      } else if (value.schemaNameRef === 'presets') {
+        docIdRef = presetNameToId.get(name)
+      }
```
EvanHahn, line 765: Maybe we could error if the `schemaNameRef` is something unexpected?

```suggestion
      } else {
        throw new Error(
          `${JSON.stringify(value.schemaNameRef)} must be "fields" or "presets"`
        )
```

src/config-import.js
```diff
@@ -215,6 +285,53 @@ async function findPresetsFile(entries) {
...
+  for (let [lang, translationByLang] of Object.entries(translations)) {
+    const langTranslation = { presets: {}, fields: {} }
+    if (!isRecord(translationByLang))
+      throw new Error('invalid translation lang object')
+    for (let [docType, translationByDocType] of Object.entries(
+      translationByLang
+    )) {
+      if (!isRecord(translationByDocType))
+        throw new Error('invalid translation docType object')
+      if (!(docType === 'presets' || docType === 'fields')) continue
```
EvanHahn, line 325: Should we error or warn in this case?

src/config-import.js
```diff
@@ -215,6 +285,53 @@ async function findPresetsFile(entries) {
...
+
+      langTranslation[docType] = translationByDocType
+    }
+    if (Object.keys(langTranslation).length === 0) continue
```
EvanHahn, line 329: `Object.keys(langTranslation).length` will always be nonzero because of line 317 above. What's the intent of this line?

src/config-import.js
```diff
@@ -215,6 +285,53 @@ async function findPresetsFile(entries) {
...
+  )
+
+  if (!translationEntry) return
+
+  /** @type {unknown} */
+  let result
+  try {
+    result = await json(await translationEntry.openReadStream())
+  } catch (err) {
+    throw new Error('Could not parse translation.json')
```
EvanHahn, line 303: Nitpick:

```suggestion
    throw new Error('Could not parse translations.json')
```

src/config-import.js
```diff
@@ -17,6 +18,20 @@ const MAX_ICON_SIZE = 10_000_000
  *   fields:  { [id: string]: unknown }
  * }} PresetsFile
  */
+
+/**
+ * @typedef {{
+ *  [lang: string] : unknown
+ * }} TranslationsFile
```
EvanHahn, line 25: I think the difference between the `TranslationsFile` and `Translations` types is a little confusing. `TranslationsFile` is the raw data that comes from the file and `Translations` the partly-parsed data from the file, which is only fully parsed later.

IMO, `findTranslationsFile` should return its result without passing anything to `parseTranslations`. `parseTranslations` should be deleted, its should go into the `*translations()` generator, and the `Translations` type should be removed.

What do you think?

src/config-import.js
```diff
@@ -183,6 +199,60 @@ export async function readConfig(configPath) {
...
+          if (!isRecord(languageTranslationsForDocType))
+            throw new Error(`invalid translation docType`)
+          for (let [docName, fieldsToTranslate] of Object.entries(
+            languageTranslationsForDocType
+          )) {
+            if (!isRecord(fieldsToTranslate))
+              throw new Error(`invalid translation field`)
+            for (let [fieldRef, message] of Object.entries(fieldsToTranslate)) {
+              // we skip messages that aren't strings, for now
+              if (isRecord(message)) continue
```
EvanHahn, line 229: When would `message` ever be a record?

src/config-import.js
```diff
@@ -183,6 +199,60 @@ export async function readConfig(configPath) {
...
+                throw new Error(`invalid translation message`)
+              let translationValue = {
+                /** @type {'translation'} */
+                schemaName: 'translation',
+                languageCode,
+                regionCode: regionCode || '',
+                schemaNameRef,
+                fieldRef,
+                message,
+                docIdRef: '',
```
EvanHahn, line 240: I was surprised that `docIdRef` was an empty string here. It seems like it's necessary for types/validation but it's impossible to actually know at this point.

I wonder if we should return a slightly different type here, where `docIdRef` is omitted?

EvanHahn (changes requested):
I wonder if you could avoid the nested loops by defining functions that make it clearer what each step does. Something like:

```javascript
for (let [lang, languageTranslations] of Object.entries(
  translationsFile
)) {
  yield* translationsForLanguage(lang, languageTranslations)
}

// ...

/**
 * @param {string} lang
 * @param {Record<ValidDocTypes, unknown>} languageTranslations
 * @returns {Iterable<{name: string, value:import('@mapeo/schema').TranslationValue}>}
 */
function* translationsForLanguage(lang, languageTranslations) {
  const { language: languageCode, region: regionCode } = parseBCP47(lang)
  if (!languageCode) throw new Error(`invalid translation language`)
  for (let [schemaNameRef, languageTranslationsForDocType] of Object.entries(
    languageTranslations
  )) {
    // TODO: Define this function...
    yield* translationsForDocType(
      languageCode,
      regionCode,
      schemaNameRef,
      languageTranslationsForDocType
    )
  }
}
```

Might be clearer? Not sure.

> there's a type of translation thats always empty called `category`. I don't know the use of it, but I'm skipping it - for now - and only importing 'fields' and 'presets'. We may want to update the default config (and config builder) in the future to remove this..

I agree that it'd be useful to remove this from the config.

tests/config-import.js
tomasciccola: ooops!

src/config-import.js
tomasciccola: I think we should warn but maybe after we update the default config so that it doesn't have `categories` (I can leave a TODO comment that addressed this, but for know we just `continue`?)

src/config-import.js
tomasciccola: I agree. `parseTranslations` is kinda redundant here...

src/config-import.js
tomasciccola: Yeah, I'll spread the `translationValue` with `docIdRef: ''` only when doing validation, but return a type that `Omit`s that field (since it will be filled later when actually importing the config)

src/config-import.js
tomasciccola: It shouldn't actually, it was a misconception I had based on badly parsing the file...

tomasciccola: PR updating translation options in fields -> https://github.com/digidem/mapeo-default-config/pull/58

tomasciccola: PR to update default-config -> https://github.com/digidem/mapeo-default-config/pull/58

src/config-import.js
```diff
@@ -17,24 +18,34 @@ const MAX_ICON_SIZE = 10_000_000
...
+ *  [lang: string]: unknown
+ * }} TranslationsFile
+ */
+
 /**
  * @typedef {Parameters<import('./icon-api.js').IconApi['create']>[0]} IconData
  */
 
+/** @type {Error[]} */
+let warnings = []
```
EvanHahn, line 34: I think I see why this got hoisted, but I don't love that this is a global. Is there a way to avoid doing something like this? Maybe by moving `translationsForLanguage` into `*translations()`?

src/config-import.js
```diff
@@ -17,24 +18,34 @@ const MAX_ICON_SIZE = 10_000_000
...
-
+  warnings = []
   const zip = await yauzl.open(configPath)
   if (zip.entryCount > MAX_ENTRIES) {
     // MAX_ENTRIES in MAC can be inacurrate
     throw new Error(`Zip file contains too many entries. Max is ${MAX_ENTRIES}`)
   }
   const entries = await zip.readEntries(MAX_ENTRIES)
   const presetsFile = await findPresetsFile(entries)
+  const translationsFile = await findTranslationsFile(entries)
```
EvanHahn, line 48: Minor: could we do these concurrently?

```suggestion
  const [presetsFile, translationsFile] = await Promise.all([
    findPresetsFile(entries),
    findTranslationsFile(entries),
  ])
```

src/config-import.js
```diff
@@ -133,29 +144,29 @@ export async function readConfig(configPath) {
     },
 
     /**
-     * @returns {Iterable<{ fieldNames: string[], iconName: string | undefined, value: import('@mapeo/schema').PresetValue }>}
+     * @returns {Iterable<{ fieldNames: string[], iconName: string | undefined, value: import('@mapeo/schema').PresetValue, name: String}>}
```
EvanHahn, line 147: Nitpick (here and elsewhere in the file): could we use `string` instead of `String`? I _think_ TypeScript says these are equivalent but I really don't wanna use the [`String` object](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String#string_primitives_and_string_objects) by accident.

src/config-import.js
```diff
@@ -215,6 +247,164 @@ async function findPresetsFile(entries) {
...
+function* translationsForLanguage(lang, languageTranslations) {
+  const { language: languageCode, region: regionCode } = parseBCP47(lang)
+  if (!languageCode) {
+    warnings.push(new Error(`invalid translation language ${lang}`))
+    return
+  }
+  for (const [schemaNameRef, languageTranslationsForDocType] of Object.entries(
+    languageTranslations
+  )) {
+    // TODO: push to warnings when removing categories from default config
```
EvanHahn, line 285: What does this mean? And should we actually add to `warnings` in this case, uncommenting line 287?

src/config-import.js
```diff
@@ -215,6 +247,164 @@ async function findPresetsFile(entries) {
...
+      warnings.push(
+        new Error(`Invalid translation message type ${typeof message}`)
+      )
+      continue
+    }
+  }
+}
+/**
+ * @param {Object} opts
+ * @param {any} opts.value
```
EvanHahn, line 378: I think we know the type of this...could we remove this `any`?

src/config-import.js
```diff
@@ -215,6 +247,164 @@ async function findPresetsFile(entries) {
...
+ * @param {string} opts.docName
+ * @param {Record<string,unknown>} opts.message
+ */
+function* translateMessageObject({ value, message, docName }) {
```
EvanHahn, line 382: Rather than translating arbitrary objects, what do you think about only translating objects of the shapes we expect?

tests/config-import.js
```diff
@@ -213,6 +213,22 @@ test('config import - presets', async () => {
   assert.equal(config.warnings.length, 0, `no warnings on the file`)
 })
 
+test('config import - translations', async () => {
+  const config = await readConfig(defaultConfigPath)
+  for (const { value } of config.translations()) {
+    assert.equal(value.schemaName, 'translation', `schemaName is 'translation'`)
+    assert(
+      value.schemaNameRef === 'presets' || value.schemaNameRef === 'fields',
+      `Config translates only 'fields' or 'presets`
```
EvanHahn, line 222: Nitpick:

```suggestion
      `Config translates only 'fields' or 'presets'`
```

EvanHahn (changes requested):
LGTM other than my small comments.

src/config-import.js
EvanHahn: If two configs are imported at the same time, these warnings will get squashed because `readConfig` is asynchronous. Could we move this "back in" to `readConfig`?

EvanHahn (changes requested):
LGTM other than my small comment. Happy to pair if helpful.

src/config-import.js
```diff
@@ -215,6 +248,188 @@ async function findPresetsFile(entries) {
...
+    const { language: languageCode, region: regionCode } = parseBCP47(lang)
+    if (!languageCode) {
+      warnings.push(new Error(`invalid translation language ${lang}`))
+      return
+    }
+    for (const [
+      schemaNameRef,
+      languageTranslationsForDocType,
+    ] of Object.entries(languageTranslations)) {
+      // TODO: remove categories check when removed from default config
```
EvanHahn, line 291: This is still necessary even with the new default config, right?

EvanHahn (approved):
LGTM! One small non-blocking question.

PR #607, opened 2024-05-06
chore: require `default` case in `switch` statements

EvanHahn:
This enables ESLint's [`default-case`][0] and [`default-case-last`][1] rules to ensure that all `switch` statements end with a `default`.

There was one spot where we weren't doing this, which I fixed.

[0]: https://eslint.org/docs/rules/default-case
[1]: https://eslint.org/docs/rules/default-case-last

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,6 +53,8 @@
     },
     "rules": {
       "eqeqeq": "error",
+      "default-case": "error",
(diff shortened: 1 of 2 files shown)


achou11 (approved):
how does this play with TS? Isn't there a config option that could conflict with this?

PR #611, opened 2024-05-06
feat: add "on backgrounded" function

### LINKED ISSUE #576: Expose API to stop syncing when app goes into background
When CoMapeo goes into the background, we should gracefully shut down sync.

For this task, that means the following:

1. Add a function like `shutdownSync`[^1], which the frontend will call when the app goes into the background.
1. Close the local discovery TCP listener (probably means calling `LocalDiscovery.prototype.#stop`)
1. For each open connection...
   1. Stop replicating (similar to blocked) if they haven't started replicating back.
   1. Wait until sync completes. (Likely requires no waiting because the app usually isn't syncing when the app goes into the background.)
   1. Close the connection.

[^1]: Come up with a better name if you can! Please!!

EvanHahn:
`MapeoManager#onBackgrounded()` should be called when the app goes into the background. It will do its best to gracefully shut down sync.

Closes #576.


src/sync/peer-sync-controller.js
```diff
@@ -184,10 +247,7 @@ export class PeerSyncController {
     }
     this.#log('capability %o', this.#syncCapability)
 
-    // If any namespace has new data, update what is enabled
-    if (Object.values(didUpdate).indexOf(true) > -1) {
```
gmaclennan, line 188: I think removing this might have significant performance implications.

EvanHahn: I'm not too worried about this.

`#updateEnabledNamespaces` eventually calls `#enableNamespace` and `#disableNamespace`. `#enableNamespace` won't double-enable namespaces:

https://github.com/digidem/mapeo-core-next/blob/2cb7a0b0eba0fb14e24edefc2a9a6032ef70494a/src/sync/peer-sync-controller.js#L331-L332

Same thing is true for `#disableNamespace`.

It's possible that there's something slow along the way, but I see no (1) async (2) variable-sized loops[^0] (3) external system calls.

Might be wrong, though!

[^0]: We iterate over constant-sized arrays, like `NAMESPACES`.

gmaclennan: oh yes sorry my mistake, I was muddling this up with the reading sync state (bitfields) which we try to minimize because that _is_ expensive. I don't see anything expensive here, and I think we throttle sync state updates, so it won't be called too often either.

src/sync/peer-sync-controller.js
```diff
@@ -25,7 +30,8 @@ export class PeerSyncController {
   #roles
   /** @type {Record<Namespace, SyncCapability>} */
   #syncCapability = createNamespaceMap('unknown')
-  #isDataSyncEnabled = false
```
EvanHahn, line 28: Previously, sync could be in two possible states: data sync disabled and data sync enabled.

Now, there are three possible states: syncing nothing, only syncing initial namespaces, and syncing everything. I called these three states "namespace groups" and would love a better name.

src/sync/autostopper.js
```diff
@@ -0,0 +1,125 @@
...
+ * - `autostopAfter` can be `0` if you want to stop sync as soon as it's done.
+ *
+ * - `autostopAfter` can be `Infinity` if you never want to auto-stop.
+ *
+ * - Changing `autostopAfter` does not restart the timer. For example, imagine
+ *   `autostopAfter` is `100` and sync completes at time `T`. If you change
+ *   `autostopAfter` to `200` at time `T < 100`, the timer will fire at time
+ *   `T + 200`.
+ */
+export default class Autostopper {
```
EvanHahn, line 33: I found the auto-stop logic fairly complex and decided to put it into its own (unit tested) class.

src/sync/sync-api.js
```diff
@@ -50,11 +71,60 @@ export class SyncApi extends TypedEmitter {
...
+      case 'initial only':
+        return false
+      case 'all':
+        return true
+      default:
+        throw new ExhaustivenessError(enabledNamespaceGroups)
+    }
+  }
+
+  #update() {
```
EvanHahn, line 107: Previously, `start()` and `stop()` modified the `PeerSyncController`s directly. Now, you describe what you want in `#internalSyncState` and apply it with `#update()`. You should be able to call `#update()` whenever you want.

Note that _logging was removed_ because it was hard to apply that here. Let me know if I should re-add that back.

src/sync/peer-sync-controller.js
```diff
@@ -184,18 +181,45 @@ export class PeerSyncController {
...
+        isAnySyncEnabled = true
+        isDataSyncEnabled = false
+        break
+      case 'all':
+        isAnySyncEnabled = isDataSyncEnabled = true
+        break
+      default:
+        throw new ExhaustivenessError(this.#namespaceGroupToReplicate)
+    }
+
```
gmaclennan, line 216: Personal preference comment: I'm not sure if the indirection created here is helpful, or just makes the code below harder to follow vs. replacing `if (!isAnySyncEnabled)` with `if (this.#nameSpaceGroupToReplicat !== 'none')` and replacing `else if (isDataSyncEnabled)` with `else if (this.#nameSpaceGroupToReplicate === 'all'`?

src/sync/peer-sync-controller.js
gmaclennan: Re. naming, I think it's confusing to introduce the term "replicate" when we are using "sync" mainly here.

Also I think we should align the term we use in `PRESYNC_NAMESPACES` with the term `initial only`.

I suggest, although clunky and not the best:

- `type SyncEnabledState = 'none' | 'presync' | 'all'`
- `this.#syncEnabledState: SyncEnabledState = 'none'`
- `setSyncEnabledState(syncEnabledState)`

gmaclennan (commented):
I find it quite hard to follow the logic here, and the overall approach doesn't "feel" quite right (sorry to be vague, maybe talking it through will help me get clear on specifics). I _think_ what autostop is doing is stopping sync after a fixed period of time, rather than unreplicating cores when sync has finished, but I'm not sure because I don't completely understand the logic. I think it would be worth co-working on this a bit more.

src/sync/peer-sync-controller.js
EvanHahn: Done in b8c5d7decd92d009dabdf0537ceb0f0a1ab941a7 and f427f649f112b2cbe56c190ce41f2e6ea530c3c8.

src/sync/sync-api.js
```diff
@@ -125,34 +129,87 @@ export class SyncApi extends TypedEmitter {
    */
   #getState(namespaceSyncState) {
     const state = reduceSyncState(namespaceSyncState)
-    state.data.syncing = this.#isSyncing
+
+    state.data.syncing =
+      this.#wantsToSyncData &&
+      (!this.#isBackgrounded ||
+        !isSynced(namespaceSyncState, 'full', this.#peerSyncControllers))
+
```
gmaclennan, line 137: It probably makes sense to rename `state['data' | 'initial'].syncing` to `state['data' | 'initial'].syncEnabled` to be clear about what this means.

src/sync/sync-api.js
```diff
@@ -50,7 +56,8 @@ export class SyncApi extends TypedEmitter {
   #pscByPeerId = new Map()
   /** @type {Set<string>} */
   #peerIds = new Set()
-  #isSyncing = false
+  #wantsToSyncData = false
+  #isBackgrounded = false
```
gmaclennan, line 60: I think this naming (`isBackgrounded`) is confusing, because it doesn't describe what this does, it describes a state that only really makes sense on mobile, and it's behaviour is determined by platform specifics of how we want sync to behave when in the background. Understanding what it does requires reading the code, or understanding that we want sync to gracefully stop when complete when the app is in the background. I'm thinking ahead to when we integrate this into Desktop and Mapeo Cloud server. This is why in my hacking around yesterday I was moving towards the concept of starting and stopping both sync types (initial & data), and being more explicit about auto-starting initial sync on initialization (e.g. starting initial sync in the constructor).

I'm not great at names, but perhaps we can call this either `wantsToStopAllSync` or the inverse `wantsToSyncInitial` or `wantsToSyncPresync`. Then perhaps the methods can be `kStopAllSyncGracefully` and `kResumeSync`?

src/sync/sync-api.js
```diff
@@ -125,34 +129,87 @@ export class SyncApi extends TypedEmitter {
...
+  #updateState = () => {
+    const namespaceSyncState = this[kSyncState].getState()
+
+    /** @type {SyncEnabledState} */ let syncEnabledState
+    if (this.#isBackgrounded) {
+      let isStopped = true
+      for (const peerSyncController of this.#peerSyncControllers.values()) {
+        isStopped = peerSyncController.syncEnabledState === 'none'
+        break
+      }
```
gmaclennan, line 150: I think it starts to get quite confusing storing this state on the peer sync controllers, since a future maintainer could without understanding set a peer controller sync enabled state somewhere, so that this state is different for some PSCs, which would break this. There is also kind of vague behaviour around when PSC instances are created with syncEnabledState === 'none', before it is set from this class, because this class reads that state as the source of truth while also setting that state. It's kind of circular...

I suggest that the syncEnabled state is hoisted, either to this class or to the SyncState class, which is already passed to the PSC instances, so they could read syncState.syncEnabledState from there. That would also avoid the not-quite-determined behaviour when PSC instances are created, because they will read the shared value from SyncState, rather than having their own state until it starts being controlled by this class.

gmaclennan: What I'm wanting to avoid here is implicit context knowledge in the codebase, e.g. "you must set the syncEnabledState of all PSCs at the same time, but you can't entirely rely on any one PSC's syncEnabledState to be accurate, so we should re-set syncEnabledState each time we update".

src/sync/sync-api.js
gmaclennan: This is essentially calculating the value of `syncEnabledState` based on the same logic that is in `#updateState()`, and we need to remember to keep those two pieces of logic in sync. Hoisting syncEnabledState to somewhere that can be read here would avoid having this duplication.

For completeness we should also set `state.initial.syncing` here, because currently it just defaults to always be `true`, but now it can be false.

gmaclennan (changes requested):
Good work on this, the logic looks correct to me, but there is one place in particular where it seems very fragile around management of the `syncEnabledState` - it should work as-is, but only because of nuances of Map iteration order and how we read it. I think we should fix this before merging.

Summary of other comments:

- We need to define the `state.initial.syncing` now, because currently it defaults to `true` always.
- There are a few naming things that might help future maintainers understand things.
- Finally I think it's not great to have duplicate logic for deriving syncEnabledState in both the `#getState` and `#updateState` methods.

src/sync/sync-api.js
EvanHahn: Renamed in 4a641e77dc74e9b316201f4c667fa5fcad2eb52e.

src/sync/sync-api.js
EvanHahn: Refactored in 8fe34aaec18b07facd878c7c93fe96c67d8295a8.

EvanHahn:
Thanks for all the review here. Hopefully this is the last round!

src/sync/sync-api.js
```diff
@@ -125,34 +131,94 @@ export class SyncApi extends TypedEmitter {
...
+        state.initial.isSyncEnabled = state.data.isSyncEnabled = true
+        break
+      default:
+        throw new ExhaustivenessError(this.#previousSyncEnabledState)
+    }
+
     return state
   }
 
+  #updateState = () => {
```
gmaclennan, line 153: Maybe add namespace sync state as an optional parameter, to save recalculating it on state events. It's expensive to calculate.

EvanHahn: Done in db4ff6c6318bc43897c5bf84ff8c60ece7e6ef67.

transform.ts
gmaclennan, line 1: Is this meant to be in here?

PR #613, opened 2024-05-07
chore: replace `var` with `let` and `const`

EvanHahn:
This enables the [ESLint no-var rule][no-var] and fixes all violations.

Most of the changes were mechanical but a few required a small amount of work.

[no-var]: https://eslint.org/docs/latest/rules/no-var


tests/blob-store/combine-states.js
```diff
@@ -97,31 +97,44 @@ test('arithmetic test', (t) => {
...
- * @returns {IterableIterator<Array<T>>}
+ * @param {ReadonlyArray<T>} arr
+ * @returns {IterableIterator<ReadonlyArray<T>>}
  */
-export function* permute(arr) {
-  var length = arr.length,
-    c = Array(length).fill(0),
-    i = 1,
-    k,
-    p
```
EvanHahn, line 110: I had too much trouble fixing this so I just reimplemented it (using [a solution I wrote last year][0]).

[0]: https://evanhahn.com/nested-list-permutations-in-javascript/#flat-permutations

achou11 (commented):
Will to defer to @gmaclennan about the changes here because I vaguely remember some of the files that have been updated here were vendored from the holepunch source code. otherwise I'm personally fine with the change

EvanHahn:
This [was initially vendored from Holepunch](https://github.com/digidem/mapeo-core-next/blob/cb43d38fa66d796ac3d50236fd66caf0c098edda/src/core-manager/bitfield-rle.js#L1-L3) but it's changed so much that I think it's become a separate thing. I don't feel strongly, but I think dropping `var` is worth it here, especially because we already have in several places in this file.

Alternatively, we could disable the rule for this file. Again, I don't feel strongly.

gmaclennan (approved):
fine by me. FWIW using `var` inside `try...catch` was deliberate, I remember discussing with someone on the team the best pattern for assigning variables in a `try...catch`. I'm not beholden to that pattern though.

achou11:
> fine by me. FWIW using `var` inside `try...catch` was deliberate, I remember discussing with someone on the team the best pattern for assigning variables in a `try...catch`. I'm not beholden to that pattern though.

that someone was me - also fine with abandoning that pattern 😄

PR #624, opened 2024-05-07
chore: fix type errors in bitfield-rle tests

EvanHahn:
This fixes all the type errors in the tests for `bitfield-rle`. Some of the type errors were entirely test-only, but others required changes to the "real" code. Most notably, many `Buffer` types were changed to `Uint8Array`.

### DIFF
--- a/src/core-manager/bitfield-rle.js
+++ b/src/core-manager/bitfield-rle.js
@@ -15,8 +15,8 @@ const n = 4
 class State {
   /**
    *
-   * @param {Uint8Array} input
-   * @param {Uint8Array | undefined} output
+   * @param {Buffer} input
+   * @param {Buffer | undefined} output
    * @param {number} offset
    */
   constructor(input, output, offset) {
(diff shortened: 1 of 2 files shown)


src/core-manager/bitfield-rle.js
```diff
@@ -31,13 +31,10 @@ class State {
 encode.bytes = 0
 
 /**
- * @param {Uint32Array} bitfield
- * @param {Buffer} [buffer]
- * @param {number} [offset]
+ * @param {import('type-fest').TypedArray} bitfield
+ * @returns {Buffer}
  */
-export function encode(bitfield, buffer, offset) {
```
EvanHahn, line 38: These parameters were unused so I removed them.

src/core-manager/bitfield-rle.js
```diff
@@ -31,13 +31,10 @@ class State {
 encode.bytes = 0
 
 /**
- * @param {Uint32Array} bitfield
- * @param {Buffer} [buffer]
- * @param {number} [offset]
+ * @param {import('type-fest').TypedArray} bitfield
```
gmaclennan, line 34: Are you sure this is safe to do? I don't fully remember all of this code, but when writing it was specifically written for uint32Arrays - I think there are issues with endianess when accepting other options, given the modifications I made to the code.

gmaclennan (commented):
I don't feel confident that this works correctly for typedarrays other than uint32Array. Do we need to make this change?

EvanHahn:
I redid this to limit source changes.

`encodingLength` was only used in tests and its tests had tricky type errors. I made `encodingLength` private and removed the tests.

gmaclennan (approved):

PR #626, opened 2024-05-08
chore: throttle pre-have updates to avoid network flooding

EvanHahn:
Followup to 6ad3beaaff8d10d4a23a52380e73d8f76c833928. If we're adding lots of data at once (like during a config import), we shouldn't flood the network.

See [this comment](https://github.com/digidem/mapeo-core-next/commit/6ad3beaaff8d10d4a23a52380e73d8f76c833928#r141813463).

### DIFF
--- a/src/core-manager/index.js
+++ b/src/core-manager/index.js
@@ -1,7 +1,7 @@
 // @ts-check
 import { TypedEmitter } from 'tiny-typed-emitter'
 import Corestore from 'corestore'
-import { throttle } from 'throttle-debounce'
+import { debounce } from 'throttle-debounce'
 import assert from 'node:assert/strict'
 import { sql, eq } from 'drizzle-orm'
 import { discoveryKey } from 'hypercore-crypto'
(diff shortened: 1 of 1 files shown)


achou11 (approved):
changes seem fine although i wouldn't consider myself qualified to review this, especially for the question of what a proper throttle time is 😅

EvanHahn:
I'll merge this next week to give @gmaclennan a chance to chime in.

gmaclennan:
I would lean towards debouncing rather than throttling for this, with a shorter delay.

During imports (e.g. lots of additions to the DB), we only really want to send one pre-have, when it completes. For the user actually recording data, e.g. saving an observation, we want the prehave to be sent soon after they save.

I suggest a debounce of 1000ms. That should result in imports (e.g. config import) just sending a single pre-have, and for user created data, the prehave will be sent within 1 second of them saving, which should feel "snappy" to most users I think.

EvanHahn:
Changed to a 1000ms debounce. I'll land if CI passes.

PR #630, opened 2024-05-09
chore: clean up test createManager utility

EvanHahn:
_This is a test-only change. It is not urgent._

This fixes two small issues with the createManager test utility:

- Removes a `@ts-ignore` directive
- Avoid an unnecessary `t.teardown` call

These changes should have no impact.

(For some reason, I have wanted to tidy this every time I've looked at it...this
is me being nitpicky to the max.)

### DIFF
--- a/test-e2e/utils.js
+++ b/test-e2e/utils.js
@@ -194,21 +194,28 @@ export async function createManagers(count, t, deviceType) {
  * @param {import('../src/generated/rpc.js').DeviceInfo['deviceType']} [deviceType]
  */
 export function createManager(seed, t, deviceType) {
-  const dbFolder = FAST_TESTS ? ':memory:' : temporaryDirectory()
-  const coreStorage = FAST_TESTS ? () => new RAM() : temporaryDirectory()
+  /** @type {string} */ let dbFolder
+  /** @type {string | import('../src/types.js').CoreStorage} */ let coreStorage
 
-  t.teardown(async () => {
-    if (FAST_TESTS) return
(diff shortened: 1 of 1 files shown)


test-e2e/utils.js
```diff
@@ -194,21 +194,28 @@ export async function createManagers(count, t, deviceType) {
...
+
+  if (FAST_TESTS) {
+    dbFolder = ':memory:'
+    coreStorage = () => new RAM()
+  } else {
+    const directories = [temporaryDirectory(), temporaryDirectory()]
+    ;[dbFolder, coreStorage] = directories
+    t.teardown(() =>
+      Promise.all(
+        directories.map((dir) =>
```
achou11, line 208: countering your nitpick with my own nitpick 😄 (i always find the semi-colon trick(?) a little hard to parse)

```suggestion
    dbFolder = temporaryDirectory()
    coreStorage = temporaryDirectory()
    t.teardown(() =>
      Promise.all(
        [dbFolder, coreStorage].map((dir) =>
```

achou11 (approved):

test-e2e/utils.js
EvanHahn: This doesn't work because TypeScript doesn't know that `coreStorage` is a string at this point. We could fix that by defining an intermediate variable?

achou11: ah welp nevermind 😂  - probably not worth going out of your way to change what you had before, especially since this is just test-specific

PR #631, opened 2024-05-09
chore: move `@digidem/types` package to devDependencies

EvanHahn:
*This change is not urgent.*

This is a types-only dependency. We don't need it in production.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,6 @@
       "hasInstallScript": true,
       "license": "MIT",
       "dependencies": {
-        "@digidem/types": "^2.3.0",
         "@electron/asar": "^3.2.8",
         "@fastify/error": "^3.4.1",
         "@fastify/static": "^7.0.3",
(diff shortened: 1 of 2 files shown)


achou11:
would consult @gmaclennan on this one since I'm pretty sure we included it as a normal dependency intentionally - something related to needing to include the types when you publish this as a package 🤔

EvanHahn:
Looks like this was originally added in #125. I didn't see any discussion there, but I'm sure there's a good reason.

achou11:
> Looks like this was originally added in #125. I didn't see any discussion there, but I'm sure there's a good reason.

not necessarily, could easily be misremembering 😅 kind of tedious, but maybe doing an `npm pack` and trying to consume the module in another project will confirm it?

gmaclennan:
Yes I think the typescript docs or definitely typed has info about which type packages need to be included as dependencies (vs devDependencies) when publishing a package. Don't remember where that info is off the top of my head, but I think this needs to be a dep for types to work when consuming as a module. I may be wrong.

EvanHahn:
Not going to investigate this.

Sorry to waste some time!

PR #633, opened 2024-05-13
chore: update dot-prop dependency to 9.0.0

EvanHahn:
This updates dot-prop to its latest version, v9.0.0.

The only breaking change in [the release notes][0] doesn't affect us: Node 18+ is now required.

[0]: https://github.com/sindresorhus/dot-prop/releases/tag/v9.0.0

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -28,7 +28,7 @@
         "compact-encoding": "^2.12.0",
         "corestore": "^6.8.4",
         "debug": "^4.3.4",
-        "dot-prop": "^8.0.2",
+        "dot-prop": "^9.0.0",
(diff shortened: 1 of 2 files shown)


tomasciccola:
It seems that this version requires minimum node 18, I don't think this may hurt us because we use 18 on the backend, but could this have implications on the frontend?. Do `nodejs-mobile` uses node 18 (I think so?) does react native cares about this? (I don't think so?)

achou11 (approved):
requiring node 18 should be fine. this is only run in the context of the backend process in the client apps, which will be running Node >=18

tomasciccola (approved):

PR #643, opened 2024-05-13
test: move FastifyController tests to `node:test`

EvanHahn:
_I recommend [reviewing this with whitespace changes disabled](https://github.com/digidem/mapeo-core-next/pull/643/files/?w=1)._

This test-only change drops Brittle from our FastifyController tests and replaces them with `node:test` and `node:assert`.

### DIFF
--- a/tests/fastify-controller.js
+++ b/tests/fastify-controller.js
@@ -1,10 +1,11 @@
 // @ts-check
-import { test } from 'brittle'
+import test from 'node:test'
+import assert from 'node:assert/strict'
 import Fastify from 'fastify'
 
 import { FastifyController } from '../src/fastify-controller.js'
 
-test('lifecycle', async (t) => {
+test('lifecycle', async () => {
   const fastify = Fastify()
   const fastifyController = new FastifyController({ fastify })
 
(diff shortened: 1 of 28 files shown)


tests/fastify-controller.js
```diff
@@ -15,17 +16,17 @@ test('lifecycle', async (t) => {
     { host: '0.0.0.0' },
   ]
 
-  for (const opts of startOptsFixtures) {
-    await fastifyController.start(opts)
-    await fastifyController.start(opts)
-    await fastifyController.stop()
-    await fastifyController.stop()
+  assert.doesNotReject(async () => {
```
achou11, line 19: think the usage of assert can be skipped entirely in this case

> Using assert.doesNotReject() is actually not useful because there is little benefit in catching a rejection and then rejecting it again. Instead, consider adding a comment next to the specific code path that should not reject and keep error messages as expressive as possible.

https://nodejs.org/docs/latest-v20.x/api/assert.html#assertdoesnotrejectasyncfn-error-message

achou11 (changes requested):

tests/fastify-controller.js
EvanHahn: Good point. Changed in 49397fda0d020974b7fb276de2f117efea74cdd9.

EvanHahn:
Not sure why there are so many commits listed here (I probably messed up a merge), but I'll squash before merging.

achou11 (approved):

PR #647, opened 2024-05-13
test: move schema tests to `node:test`

EvanHahn:
This test-only change drops Brittle from our schema tests and replaces
them with `node:test` and `node:assert`.

### DIFF
--- a/tests/schema.js
+++ b/tests/schema.js
@@ -1,6 +1,6 @@
 // @ts-check
-/* eslint-disable no-unused-vars */
-import { test } from 'brittle'
+import test from 'node:test'
+import assert from 'node:assert/strict'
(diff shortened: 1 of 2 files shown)


tests/schema.js
```diff
@@ -72,7 +66,7 @@ test('Expected table config', (t) => {
  * @typedef {Extract<MapeoDoc, { schemaName: T }>} MapeoType
  */
 
-test('Types match', { skip: true }, (t) => {
+test('Types match', { skip: true }, () => {
```
achou11, line 69: completely unrelated, but wondering if this belongs in `test-types/` 🤔

achou11 (approved):

tests/schema.js
EvanHahn: Ask and ye shall receive: #657

PR #648, opened 2024-05-13
test: move CoreSyncState tests to `node:test`

EvanHahn:
This test-only change drops Brittle from our CoreSyncState tests and
replaces them with `node:test` and `node:assert`.

### DIFF
--- a/tests/sync/core-sync-state.js
+++ b/tests/sync/core-sync-state.js
@@ -376,7 +376,7 @@ test('bitCount32', () => {
 // This takes several hours to run on my M2 Macbook Pro (it's the slowBitCount
 // that takes a long time - bitCount32 takes about 23 seconds), so not running
 // this by default. The test did pass when I ran it though.
-test('bitCount32 (full test)', { skip: true }, () => {
+test.skip('bitCount32 (full test)', () => {
   for (let n = 0; n < 2 ** 32; n++) {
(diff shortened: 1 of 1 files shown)


tests/sync/core-sync-state.js
```diff
@@ -364,19 +365,23 @@ test('bitCount32', (t) => {
...
+      `${n.toString(2)} has ${expected} bit(s) set`
+    )
   }
 })
 
 // This takes several hours to run on my M2 Macbook Pro (it's the slowBitCount
 // that takes a long time - bitCount32 takes about 23 seconds), so not running
 // this by default. The test did pass when I ran it though.
-test.skip('bitCount32 (full test)', (t) => {
+test('bitCount32 (full test)', { skip: true }, () => {
```
achou11, line 379: is using the opt instead of `test.skip` a preference? I have no strong preference, just wanted to make sure I'm not missing some subtle difference

achou11 (approved):

tests/sync/core-sync-state.js
EvanHahn: They're no different [according to the docs][0]. Not sure why I changed it. Reverted it in 0e34b324cd8fd82a7203ce451360be4a5184a6d9.

[0]: https://nodejs.org/docs/latest-v18.x/api/test.html#testskipname-options-fn

PR #652, opened 2024-05-15
chore: add Node 20 to CI

achou11:
Adds Node 20 to the matrix for CI, as that's the current LTS and it would seem advisable to be testing on that version in addition to 18 (which is still needed because of our usage of NodeJS Mobile)


achou11:
ooo interesting - test failure on 20: https://github.com/digidem/mapeo-core-next/actions/runs/9099775156/job/25013115175?pr=652#step:6:3694

EvanHahn (changes requested):
I support this but can we fix the test failure? Happy to help investigate later.

achou11:
haven't been able to reproduce the test failure locally yet 🙃

EvanHahn:
I don't have a solution but I reproduced it on my machine with no issue. I wonder if it's a Linux thing?

achou11:
> I don't have a solution but I reproduced it on my machine with no issue. I wonder if it's a Linux thing?

Failure seems to occur on all machines in CI though?

EvanHahn:
Ah true. Uh...did you run `npm rebuild`? I don't expect that to be the issue.

achou11:
> Ah true. Uh...did you run `npm rebuild`? I don't expect that to be the issue.

yep, no luck though 🤔

EvanHahn:
Strange...let me know if I can help at all.

achou11:
> Strange...let me know if I can help at all.

if you can just...fix everything, that would be a great help 👍

EvanHahn:
I have a simple fix but don't really understand why it works. Investigating further...

EvanHahn:
Pushed eb23f663b9895a07a21746de452b6ddd654b0dfa to fix the problem. It has a wordy explanation about the race condition that happened here.

EvanHahn (approved):

PR #654, opened 2024-05-15
chore: fix lockfile to include platform builds of @node-rs/crc32

achou11:
Restores the inclusion of all platform-specific builds for `@node-rs/crc32`. Without this, I was not able to run the code locally (usually as tests) on macOS, with an error looking like this:

```sh
node:internal/modules/cjs/loader:1140
  const err = new Error(message);
              ^

Error: Cannot find module '@node-rs/crc32-darwin-x64'
Require stack:
- /Users/andrewchou/GitHub/digidem/mapeo-core-next/node_modules/@node-rs/crc32/index.js
- /Users/andrewchou/GitHub/digidem/mapeo-core-next/node_modules/yauzl-promise/lib/zip.js
- /Users/andrewchou/GitHub/digidem/mapeo-core-next/node_modules/yauzl-promise/lib/index.js
    at Module._resolveFilename (node:internal/modules/cjs/loader:1140:15)
    at Module._load (node:internal/modules/cjs/loader:981:27)
    at Module.require (node:internal/modules/cjs/loader:1231:19)
    at require (node:internal/modules/helpers:177:18)
    at Object.<anonymous> (/Users/andrewchou/GitHub/digidem/mapeo-core-next/node_modules/@node-rs/crc32/index.js:105:29)
    at Module._compile (node:internal/modules/cjs/loader:1364:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1422:10)
    at Module.load (node:internal/modules/cjs/loader:1203:32)
    at Module._load (node:internal/modules/cjs/loader:1019:12)
    at Module.require (node:internal/modules/cjs/loader:1231:19) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [
    '/Users/andrewchou/GitHub/digidem/mapeo-core-next/node_modules/@node-rs/crc32/index.js',
    '/Users/andrewchou/GitHub/digidem/mapeo-core-next/node_modules/yauzl-promise/lib/zip.js',
    '/Users/andrewchou/GitHub/digidem/mapeo-core-next/node_modules/yauzl-promise/lib/index.js'
  ]
}
```

Git bisect showed that these deps were removed from the lockfile via 9fa05aa99eef1bd2a9e0dab9cc6dd0a7b8069d03. @tomasciccola I'd be curious if there's some different behavior from npm that occurs when you install/update deps? I've run into this problem before where the lockfile unexpectedly changes this dep after you've made updates 🤔 

Additionally, wondering what the best way of avoiding this issue moving forward is. Otherwise it will probably keep happening.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -248,6 +248,22 @@
         "@bufbuild/buf-win32-x64": "1.26.1"
       }
     },
+    "node_modules/@bufbuild/buf-darwin-arm64": {
+      "version": "1.26.1",
+      "resolved": "https://registry.npmjs.org/@bufbuild/buf-darwin-arm64/-/buf-darwin-arm64-1.26.1.tgz",
+      "integrity": "sha512-nmyWiT/59RFja0ZuXFxjNGoAMDPTApU66CZUUevkFVWbNB9nzeQDjx2vsJyACY64k5fTgZiaelSiyppwObQknw==",
+      "cpu": [
+        "arm64"
+      ],
+      "dev": true,
+      "optional": true,
+      "os": [
+        "darwin"
+      ],
+      "engines": {
+        "node": ">=12"
+      }
+    },
     "node_modules/@digidem/types": {
       "version": "2.3.0",
       "resolved": "https://registry.npmjs.org/@digidem/types/-/types-2.3.0.tgz",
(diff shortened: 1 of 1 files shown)


achou11:
worth keeping in mind that this could very much be some kind of bug around packaging for `@node-rs/crc32` (`yauzl-promise` depends on a version that's about 3 minor releases older)

tomasciccola:
Mhm! interesting. Yeah, I know my lockfiles have messed you in the past...
Do you think installing some dep (like yauzl, i.e.) with `--verbose` may help??

EvanHahn (approved):
I'm surprised `npm ci` doesn't catch this. From its docs:

> If dependencies in the package lock do not match those in package.json, npm ci will exit with an error, instead of updating the package lock.

I would expect that to have caught this...

achou11:
> I'm surprised `npm ci` doesn't catch this. From its docs:
> 
> > If dependencies in the package lock do not match those in package.json, npm ci will exit with an error, instead of updating the package lock.
> 
> I would expect that to have caught this...

If i understand that snippet, I wouldn't expect this to be caught. The binaries are a transitive dependency of `yauzl-promise` via `@node-rs/crc32` and are marked as optional dependencies for `@node-rs/crc32`.

I'm just kind of confused about how to prevent this from happening - seems like it's kind of dependent on the machine that installs/updates the deps, plus some hidden behavior of npm depending on OS (e.g. seems like in Tomas' PR, it kept the binaries needed for Linux and pruned everything else)

package-lock.json
```diff
@@ -248,6 +248,22 @@
...
+      ],
+      "dev": true,
+      "optional": true,
+      "os": [
+        "darwin"
+      ],
+      "engines": {
+        "node": ">=12"
+      }
+    },
```
achou11, line 266: this was unintentional, but it reminds me that i've run into this issue with the `buf` tool as well, which is used for processing schemas

EvanHahn:
I wonder if `npm config ls` shows anything different on Tomás's machine?

PR #658, opened 2024-05-16
test: remove `t.plan()` calls

EvanHahn:
We plan to replace Brittle with `node:test` which doesn't have an equivalent for `t.plan`. Remove it to make it easier to migrate.

Notably, this removes `t.plan` from the type definitions to discourage its use.

### DIFF
--- a/types/brittle.d.ts
+++ b/types/brittle.d.ts
@@ -64,6 +64,7 @@ declare module 'brittle' {
   }
 
   export interface TestInstance extends Assertion {
+    plan(n: number): void
     teardown(
       fn: () => unknown | Promise<unknown>,
       options?: { order?: number }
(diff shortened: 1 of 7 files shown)


types/brittle.d.ts
achou11, line 1: think it's okay to keep it around. useful to have in case these definitions ever go upstream (or for anyone who's using brittle to reference)

achou11 (approved):
changes look fine to me. would consider keeping the type def around in the spirit of completeness, for others who may be using brittle :)

types/brittle.d.ts
EvanHahn: Restored in d94a9f3330ca1b2a397420c11c902367e3455e71.

EvanHahn:
Submitted our Brittle types upstream at https://github.com/DefinitelyTyped/DefinitelyTyped/pull/69647.

PR #665, opened 2024-05-20
chore: remove unused "device type" property from MapeoManager

EvanHahn:
This change should have no impact.

This property was never used, so we can remove it.

### DIFF
--- a/src/mapeo-manager.js
+++ b/src/mapeo-manager.js
@@ -93,8 +93,6 @@ export class MapeoManager extends TypedEmitter {
   #loggerBase
   #l
   #defaultConfigPath
-  /** @readonly */
-  #deviceType
 
   /**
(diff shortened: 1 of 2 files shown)


achou11:
I'm wondering if this is highlighting something that's missing that should be implemented? Just looking through https://github.com/digidem/mapeo-core-next/issues?q=is%3Aissue+deviceType, seems like we definitely want to support this field for sending during discovery 🤔 perhaps there was a follow-up that never came to be?

achou11 (approved):
Poking around and seeing that there's a `MapeoManager.setDeviceInfo()` which can persist that device type info. As long as it's not helpful to keep that in memory via the instance field, then this change seems reasonable to me!

EvanHahn:
Yeah, that was my thinking too.

PR #666, opened 2024-05-21
chore: test serving audio

tomasciccola:
add a test and an audio test file


tests/fixtures/blob-api/audio.mp3
EvanHahn, line 1: incredible choice. 10/10

EvanHahn (approved):
LGTM.

It seems like [GitHub Actions are having issues](https://www.githubstatus.com/)...maybe we can wait to see if those resolve before merging?

tests/fixtures/blob-api/audio.mp3
tomasciccola: heh, its the whilhelm scream, a mytical audio sample that appears in A LOT of movies

tomasciccola:
> It seems like [GitHub Actions are having issues](https://www.githubstatus.com/)...maybe we can wait to see if those resolve before merging?

Mhm, it seems that the fastify test is hanging and not related with the CI issues. Its hanging on the test after the one I've added (`retrieving style.json using stable url`). 
I wonder whats that about...

tomasciccola:
I was forgetting to stop the fastify controller..

EvanHahn (approved):

PR #672, opened 2024-05-23
fix: at startup, clean up left projects that failed to delete

### LINKED ISSUE #583: At startup, clean up left projects that failed to delete
If we start to leave a project and the app crashes, we may fail to remove data from indexers. At startup, we should find left projects and remove the "orphaned" data.

EvanHahn:
If you start to leave a project and the operation fails partway, we now finish the next time you try to load that project.

For example, imagine the following scenario:

1. The user presses the "leave project" button.
2. We mark them as having left the project, but haven't yet deleted any data.
3. Their phone runs out of battery.

Previously, the project state would be messed up in this scenario. Now, we finish the cleanup before returning the project.

Fixes #583.

### DIFF
--- a/src/mapeo-project.js
+++ b/src/mapeo-project.js
@@ -61,6 +61,7 @@ export const kBlobStore = Symbol('blobStore')
 export const kProjectReplicate = Symbol('replicate project')
 export const kDataTypes = Symbol('dataTypes')
 export const kProjectLeave = Symbol('leave project')
+export const kClearDataIfLeft = Symbol('clear data if left project')
 
 const EMPTY_PROJECT_SETTINGS = Object.freeze({})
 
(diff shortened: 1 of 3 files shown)


src/mapeo-project.js
```diff
@@ -667,15 +668,32 @@ export class MapeoProject extends TypedEmitter {
...
         'config',
         'data',
         'blob',
         'blobIndex',
       ])
+    const dataStoresToUnlink = Object.values(this.#dataStores).filter(
+      (dataStore) => dataStore.namespace !== 'auth'
+    )
+
+    await Promise.all(dataStoresToUnlink.map((ds) => ds.close()))
```
EvanHahn, line 696: We need to close earlier in the process now, because of the following race condition:

1. We open a project.
1. We close the writer core (see line 636 below).
1. We start indexing the writer core (via `MultiCoreIndexer`). _This fails because the writer core has been closed._

(This was theoretically a problem before. Someone could load a project and immediately leave it.)

test-e2e/project-leave.js
```diff
@@ -345,4 +347,51 @@ test('leaving a project while disconnected', async (t) => {
...
+        refs: [],
+        metadata: {},
+      })
+      await project.$member.assignRole(
+        manager.getDeviceInfo().deviceId,
+        /**
+         * This test-only hack assigns the "left" role ID without actually
+         * cleaning up the project.
+         * @type {any}
+         */ (LEFT_ROLE_ID)
```
EvanHahn, line 373: I'd love a better way to do this...

test-e2e/project-leave.js
```diff
@@ -345,4 +347,51 @@ test('leaving a project while disconnected', async (t) => {
   )
 })
 
+test('partly-left projects are cleaned up on startup', async (t) => {
```
achou11, line 350: it'd be nice to confirm that data (on disk) is actually cleaned up in this scenario. not sure how complicated that would be to test though

achou11 (approved):

test-e2e/project-leave.js
EvanHahn: Opened #693 to test this. (Doesn't test this _specific_ cleanup case; let me know if you want me to do that.)

PR #673, opened 2024-05-23
chore: replace Brittle with node:test

EvanHahn:
_This is a test-only change that should have no user impact._

We've slowly been removing Brittle in favor of `node:test`. This
finishes that work and uninstalls Brittle.


types/brittle.d.ts
EvanHahn, line 1: This lives on in [`@types/brittle`](https://www.npmjs.com/package/@types/brittle).

achou11 (approved):
awesome stuff!

PR #676, opened 2024-05-23
test: move icon API tests to `node:test`

EvanHahn:
This test-only change drops Brittle from our icon API tests and replaces them with `node:test` and `node:assert`.

### DIFF
--- a/tests/icon-api.js
+++ b/tests/icon-api.js
@@ -1,4 +1,5 @@
-import test from 'brittle'
+import test from 'node:test'
+import assert from 'node:assert/strict'
 import RAM from 'random-access-memory'
 import Database from 'better-sqlite3'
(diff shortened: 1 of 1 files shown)


tests/icon-api.js
```diff
@@ -560,14 +561,18 @@ test('getBestVariant() - params prioritization', (t) => {
...
+    'mime type > size > pixel density'
+  )
 })
 
 // TODO: Currently fails. Not sure if we'd run into this situation often in reality
 test(
   'getBestVariant - multiple exact matches return deterministic result',
-  { todo: true },
-  (t) => {
+  { skip: true },
```
EvanHahn, line 574: Changed `todo` to `skip` so errors didn't show up in the output, but this has no other effect (e.g., test will still be skipped either way).

tomasciccola (approved):

PR #677, opened 2024-05-23
test: move core manager tests to `node:test`

EvanHahn:
_I recommend [reviewing this with whitespace changes disabled](https://github.com/digidem/mapeo-core-next/pull/677/files?w=1)._

This test-only change drops Brittle from our core manager tests and replaces them with `node:test` and `node:assert`.


tests/core-manager.js
```diff
@@ -681,18 +678,17 @@ async function countOpenFileDescriptors(dir) {
...
+function replicateCores(a, b, { delay = 0, ...opts } = {}) {
   const s1 = a.replicate(true, { keepAlive: false, ...opts })
   const s2 = b.replicate(false, { keepAlive: false, ...opts })
   s1.on('error', (err) =>
-    t.comment(`replication stream error (initiator): ${err}`)
+    console.debug(`replication stream error (initiator): ${err}`)
   )
   s2.on('error', (err) =>
-    t.comment(`replication stream error (responder): ${err}`)
+    console.debug(`replication stream error (responder): ${err}`)
```
EvanHahn, line 691: I felt like `console.debug` was the closest analog to `t.comment`...

achou11: can use `console.log()` since `console.debug()` is an [alias for it in node](https://nodejs.org/docs/latest-v20.x/api/console.html#consoledebugdata-args)

also, if tests are run concurrently, is the location of the output where you'd expect it to be? slightly worried that this log would show up as though it's under some other test, visually

EvanHahn: Rather than answer this question, I just removed the logs entirely in d9802b95fbf662836f468b140bea7edfc7eced0d. As far as I can tell, they don't happen in normal conditions.

Also, TIL that `console.debug` is an alias for `console.log` in Node!

PR #689, opened 2024-05-28
test: speed up "create and sync data" test

EvanHahn:
*I recommend [reviewing this with whitespace changes disabled](https://github.com/digidem/mapeo-core-next/pull/689/files?w=1).*

This speeds up our "create and sync data" test. On my machine, it went from ~73 seconds to ~12.

I did this by making a few changes:

- Decrease the number of managers from 10 to 5, dramatically reducing the amount of connections required. I don't think 10 really offers much here, other than perhaps a stress test, and lowering it offers a huge speedup.

- Run assertions concurrently.

- When comparing sets, use `Set` instead of sorted arrays. This speeds things up a bit.

- Use `node:assert` instead of Brittle for a slight performance boost.  We also plan to drop Brittle in the future and this will ease the transition.

Each of these could theoretically be done separately.

I've been running this test a lot which is why I'm fixing this now.

### DIFF
--- a/test-e2e/sync.js
+++ b/test-e2e/sync.js
@@ -3,12 +3,12 @@ import assert from 'node:assert/strict'
 import { pEvent } from 'p-event'
 import { setTimeout as delay } from 'timers/promises'
 import { excludeKeys } from 'filter-obj'
+import { map } from 'iterpal'
 import {
   connectPeers,
   createManagers,
   invite,
   seedDatabases,
-  sortById,
   waitForPeers,
   waitForSync,
 } from './utils.js'
@@ -26,8 +26,11 @@ import { kSyncState } from '../src/sync/sync-api.js'
 
 const SCHEMAS_INITIAL_SYNC = ['preset', 'field']
 
-test('Create and sync data', { timeout: 100_000 }, async function (t) {
-  const COUNT = 10
+test('Create and sync data', { timeout: 60_000 }, async (t) => {
+  // NOTE: Unlike other tests in this file, this test uses `node:assert` instead
+  // of `t` to ease our transition away from Brittle. We can remove this comment
+  // when Brittle is removed.
+  const COUNT = 5
   const managers = await createManagers(COUNT, t)
   const [invitor, ...invitees] = managers
   const disconnect = connectPeers(managers, { discovery: false })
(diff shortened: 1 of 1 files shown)


test-e2e/sync.js
```diff
@@ -26,8 +26,11 @@ import { kSyncState } from '../src/sync/sync-api.js'
 
 const SCHEMAS_INITIAL_SYNC = ['preset', 'field']
 
-test('Create and sync data', { timeout: 100_000 }, async function (t) {
```
EvanHahn, line 29: I lowered the timeout because it should no longer be necessary.

test-e2e/sync.js
```diff
@@ -49,47 +52,59 @@ test('Create and sync data', { timeout: 100_000 }, async function (t) {
...
-      t.alike(
-        sortById(docs),
-        sortById(expected),
-        `All ${schemaName} docs synced to ${deviceId}`
+  await Promise.all(
+    map(generatedSchemaNames, (schemaName) =>
+      Promise.all(
+        map(projects, async (project) => {
+          const deviceId = project.deviceId.slice(0, 7)
+          // @ts-ignore - to complex to narrow `schemaName` to valid values
```
achou11, line 94: nit: may as well fix the grammar 😄 

```suggestion
          // @ts-ignore - too complex to narrow `schemaName` to valid values
```

test-e2e/sync.js
```diff
@@ -26,8 +26,11 @@ import { kSyncState } from '../src/sync/sync-api.js'
 
 const SCHEMAS_INITIAL_SYNC = ['preset', 'field']
 
-test('Create and sync data', { timeout: 100_000 }, async function (t) {
-  const COUNT = 10
+test('Create and sync data', { timeout: 60_000 }, async (t) => {
+  // NOTE: Unlike other tests in this file, this test uses `node:assert` instead
+  // of `t` to ease our transition away from Brittle. We can remove this comment
+  // when Brittle is removed.
+  const COUNT = 5
```
achou11, line 33: having this stay higher to stress test seems generally useful, but maybe there's an alternative approach to achieve that? i know we introduced something like a `FAST_TESTS` env variable but i don't think it's being used 🤔

achou11 (approved):

test-e2e/sync.js
EvanHahn: Nah, I think it's useful too. Reverted to 10 in 340bf820ebf500b4f7a3bdb0964e575ff8053158. I might consider using `FAST_TESTS` in the future, but not necessary now.

test-e2e/sync.js
EvanHahn: Done in da8dba059dac0a9263de73f352ceca748aff2efe.

test-e2e/sync.js
achou11: curious how much faster the test is now for you, since the only thing related change is parallelizing the assertions?

EvanHahn: It sped things up by about 10 seconds on my machine.

Issue #627, opened 2024-05-09
Add auto-stop option to $sync.start()


EvanHahn:
`project.$sync.start()` should take an option, `autostopAfter`, which is a millisecond duration.

Once sync has been finished for that number of milliseconds, data sync should be stopped.

EvanHahn:
#611 addresses this almost completely.

PR #690, opened 2024-05-28
feat: add auto-stop option to `$sync.start()`

EvanHahn:
`myProject.$sync.start({ autostopDataSyncAfter: 3000 })` will auto-stop data sync after it's inactive for 3 seconds.

Addresses #627.

### DIFF
--- a/src/sync/sync-api.js
+++ b/src/sync/sync-api.js
@@ -62,7 +62,8 @@ export class SyncApi extends TypedEmitter {
   #previousSyncEnabledState = 'none'
   /** @type {null | number} */
   #previousDataHave = null
-  #autostopDataSyncAfter = Infinity
+  /** @type {null | number} */
+  #autostopDataSyncAfter = null
   /** @type {null | ReturnType<typeof setTimeout>} */
   #autostopDataSyncTimeout = null
   /** @type {Map<import('protomux'), Set<Buffer>>} */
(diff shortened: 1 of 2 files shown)


test-e2e/sync.js
```diff
@@ -160,6 +162,99 @@ test('start and stop sync', async function (t) {
   await disconnect()
 })
 
+test('auto-stop', async (t) => {
```
achou11, line 165: would also test that listening for events on the sync api provides the expected sequence of payloads e.g `$sync.addListener('sync-state', ...)`

src/sync/sync-api.js
```diff
@@ -60,6 +60,11 @@ export class SyncApi extends TypedEmitter {
   #hasRequestedFullStop = false
   /** @type {SyncEnabledState} */
   #previousSyncEnabledState = 'none'
+  /** @type {null | number} */
+  #previousDataHave = null
+  #autostopDataSyncAfter = Infinity
```
achou11, line 65: nit: use of `Infinity` is a bit misleading since it's never actually used (kind of acts as an unset value) but guessing it's makes you write less type-related code

achou11 (approved):

test-e2e/sync.js
EvanHahn: Done in d4521f30edad827737858929681a69af24e9c05b.

src/sync/sync-api.js
EvanHahn: Changed to `null | number` in 8675cd16f1af407de4eaa72bfcfb9226a8ec114b.

PR #693, opened 2024-05-30
test: leaving a project should delete data from disk

EvanHahn:
This test-only change checks that data is actually deleted from disk after leaving a project.


test-e2e/utils.js
```diff
@@ -511,6 +513,40 @@ async function seedProjectDatabase(
...
+ *
+ * If the path is a directory, return the size of all regular files inside.
+ * Recurses through subdirectories.
+ *
+ * If the path is anything else, such as a symlink, `0` is returned.
+ *
+ * @param {string} filePath
+ * @returns {Promise<number>}
+ */
+export async function getFileSize(filePath) {
```
achou11, line 527: surprised their isn't an easier way to do this

nit: rename to something slightly more generic e.g. `getDiskUsage`

test-e2e/utils.js
```diff
@@ -511,6 +513,40 @@ async function seedProjectDatabase(
   return Promise.all(promises)
 }
 
+/**
+ * If the path is a regular file, return its size.
```
achou11, line 517: nit: clarify size unit here

achou11 (approved):

PR #694, opened 2024-06-05
chore: use `npm ci` for deployment, not `npm install`

EvanHahn:
[`npm ci`][0] is slightly better for deployments than `npm install`.

- It ensures all installed packages match `package-lock.json`.
- It automatically removes `node_modules/` if it exists, ensuring a clean slate.
- It won't update `package.json` or `package-lock.json`.

Not a huge difference, but recommended for deployments.

[0]: https://docs.npmjs.com/cli/v10/commands/npm-ci


tomasciccola (approved):
weird that this slipped us!

PR #700, opened 2024-06-12
feat: add e2e tests for config-import.

### LINKED ISSUE #482: Create e2e tests for loading the config in a project
## Description

This should at least test correct deletion of records after changing a config

tomasciccola:
should close #482 

Additionally this adds code to `MapeoProject`. Specifically:

1. Expose `translation` as a getter (only used in tests, but may be useful in the future)
2. Delete translations that refer to deleted presets and fields

There's a test case where I try to load the default config twice in parallel that results in - what I think - unwanted expectations. Basically we get double the presets, fields and translations since when each config is loaded there's no docs to delete. 
We should think if this is a valid (possible) case and how we can fix the code to avoid this?


src/mapeo-project.js
```diff
@@ -840,6 +849,35 @@ async function deleteAll(dataType) {
...
+    schemaNameRef,
+  } of await dataTypes.translation.getMany()) {
+    try {
+      if (schemaNameRef === 'presets' || schemaNameRef === 'fields') {
+        const toDelete = await dataTypes[schemaNameRef].getByDocId(docIdRef)
+        if (toDelete.deleted)
+          deletions.push(dataTypes.translation.delete(docId))
+      }
+    } catch (e) {
+      console.log(
```
EvanHahn, line 872: Nit: should this use `this.#l` instead of `console.log`?

src/mapeo-project.js
```diff
@@ -840,6 +849,35 @@ async function deleteAll(dataType) {
   return Promise.all(deletions)
 }
 
+/**
+@param {Object} dataTypes
+@param {MapeoProject['translation']} dataTypes.translation
+@param {MapeoProject['preset']} dataTypes.presets
+@param {MapeoProject['field']} dataTypes.fields
```
EvanHahn, line 856: Nitpick for consistency:

```suggestion
 * @param {Object} dataTypes
 * @param {MapeoProject['translation']} dataTypes.translation
 * @param {MapeoProject['preset']} dataTypes.presets
 * @param {MapeoProject['field']} dataTypes.fields
```

src/mapeo-project.js
```diff
@@ -840,6 +849,35 @@ async function deleteAll(dataType) {
...
+async function deleteTranslations(dataTypes) {
+  const deletions = []
+  for (const {
+    docId,
+    docIdRef,
+    schemaNameRef,
+  } of await dataTypes.translation.getMany()) {
+    try {
+      if (schemaNameRef === 'presets' || schemaNameRef === 'fields') {
+        const toDelete = await dataTypes[schemaNameRef].getByDocId(docIdRef)
```
EvanHahn, line 867: Could we avoid `await`ing in a loop here? Sketched out an alternate implementation which avoids it (untested):

```js
async function deleteTranslations(dataTypes) {
  const translations = await dataTypes.translation.getMany()

  await Promise.all(translations.map(async (translation) => {
    const { docId, docIdRef, schemaNameRef } = translation

    if (schemaNameRef === 'presets' || schemaNameRef === 'fields') {
      const toDelete = await dataTypes[schemaNameRef].getByDocId(docIdRef)
      if (toDelete.deleted) {
        await dataTypes.translation.delete(docId)
      }
    } else {
      // TODO: log
      await dataTypes.translation.delete(docId)
    }
  }))
}
```

test-e2e/config-import.js
```diff
@@ -0,0 +1,80 @@
...
+  await Promise.all([
+    project.importConfig({ configPath: defaultConfigPath }),
+    project.importConfig({ configPath: defaultConfigPath }),
+  ])
+  const presets = await project.preset.getMany()
+  const fields = await project.field.getMany()
+  const translations = await project.translation.getMany()
+  console.log('presets', presets.length)
+  console.log('fields', fields.length)
+  console.log('translations', translations.length)
```
EvanHahn, line 79: Can we test these rather than logging the values?

EvanHahn (changes requested):
I'll give this a more thorough review on Monday, but found a few small things for now.

src/mapeo-project.js
tomasciccola: yeah, this makes sense. But the function is outside of the class. I wrapped it in a closure (not ideal but I think it makes sense for this fn to live outside the class

test-e2e/config-import.js
tomasciccola: Yeah, that was left by accident, no there is the (for now failing) test case

src/mapeo-project.js
```diff
@@ -840,6 +849,36 @@ async function deleteAll(dataType) {
...
+ * @param {Logger} logger
+ */
+function deleteTranslations(logger) {
+  /**
+   * @param {Object} dataTypes
+   * @param {MapeoProject['translation']} dataTypes.translation
+   * @param {MapeoProject['preset']} dataTypes.presets
+   * @param {MapeoProject['field']} dataTypes.fields
+   */
+  return async function (dataTypes) {
```
EvanHahn, line 862: This "function that returns a function" feels unnecessary. Could you just pass the `Logger` as an argument? Or is there a reason for this closure?

EvanHahn (changes requested):
Looks like tests are failing. Could you fix those?

src/mapeo-project.js
tomasciccola: yeah, reverted on last commit

tomasciccola:
> Looks like tests are failing. Could you fix those?

So, the issue with the failing test is the following:
That test loads the config twice in parallel (using `Promise.all`) and results in `presets`, `fields` and `translations` being loaded to the database twice. Those doctypes should be deleted before loading a new config, but this doesn't happen when loading in parallel.
So to fix this tests I'm thinking:
1. Delete this test, we shouldn't expect a config being loaded twice in parallel
2. Find a way to force sequential excecution (So, if you call `Promise.all([project.importConfig, project.importConfig])` internally it would force sequetial execution). I don't know if we can actually enforce this

src/mapeo-project.js
```diff
@@ -815,7 +832,7 @@ export class MapeoProject extends TypedEmitter {
         relation: [],
       },
     })
-
+    this.#loadingConfig = false
```
EvanHahn, line 835: If anything above `throw`s, we won't reset this boolean. Could we wrap the above in a `try`/`catch`?

src/mapeo-project.js
```diff
@@ -840,6 +856,32 @@ async function deleteAll(dataType) {
...
+        if (schemaNameRef === 'presets' || schemaNameRef === 'fields') {
+          const toDelete = await opts[schemaNameRef].getByDocId(docIdRef)
+          if (toDelete.deleted) opts.translation.delete(docId)
+        }
+      } catch (e) {
+        opts.logger.log(
+          `referred ${schemaNameRef} is not found, deleting translation`
+        )
+        opts.translation.delete(docId)
+      }
```
EvanHahn, line 880: This code has two small problems:

1. We don't `await` the result of `opts.translation.delete()`.
2. If the first `opts.translation.delete()` rejects, we'll just try to delete it again.

Proposed fix:

```suggestion
      if (schemaNameRef !== 'presets' && schemaNameRef !== 'fields') return

      let shouldDelete = true
      try {
        const toDelete = await opts[schemaNameRef].getByDocId(docIdRef)
        shouldDelete = toDelete.deleted
      } catch (e) {
        opts.logger.log(
          `referred ${schemaNameRef} is not found, deleting translation`
        )
      }

      if (shouldDelete) {
        opts.translation.delete(docId)
      }
```

src/mapeo-project.js
```diff
@@ -492,6 +495,10 @@ export class MapeoProject extends TypedEmitter {
     return this.#dataTypes.field
   }
 
+  get translation() {
```
EvanHahn, line 498: I was a little surprised to see `$translation` and `translation` both be defined here. It _seems_ like we're only using it to get the type in `deleteTranslations` and in the tests. Is there a way to avoid this confusion?

src/mapeo-project.js
```diff
@@ -717,9 +724,19 @@ export class MapeoProject extends TypedEmitter {
    *  @returns {Promise<Error[]>}
    */
   async importConfig({ configPath }) {
+    if (this.#loadingConfig) return /** @type Error[] */ []
```
EvanHahn, line 727: If I call `importConfig({ configPath: 'path 1' })` and then `importConfig({ configPath: 'path 2' })`, the second one will be silently dropped. Is that what we want?

IMO, for simplicity, this should throw an error instead.

```suggestion
    assert(!this.#loadingConfig, 'Cannot run multiple config imports at the same time')
```

EvanHahn (changes requested):
I like the way you've enforced that imports only happen sequentially. Just a few more little things from me.

src/mapeo-project.js
tomasciccola: this makes sense. I think I was trying to catch to cases on this:
1. Delete every `translation` linked to by a `preset` or `field` (since we are also deleting those)
2. If there's a translation that links to an unexisting `preset` or `field` we should also deleted (since its basically a _useless_ translation)

tomasciccola: Additionally, typescript doesn't seem to get that the inverted logic `( !== 'presets' || !== 'fields')` is a typeguard?

src/mapeo-project.js
tomasciccola: deamn. Yeah, `$translation` refers to the api, while `translation` to the `dataType`. Changing would make this the exception (which I'm not against). Some possibilities would be:
1. Make `get $translation()` into `get $translationApi()`
2. Make `get translation()` into `get translationDataType()`
3. Expose `dataType` in `translationApi` so that one can do `this.$translation.dataType`

EvanHahn: I think I like the last option best.

EvanHahn (approved):
LGTM once we move to `this.$translation.dataType` (happy to discuss alternatives) and fix the failing test.

tomasciccola:
> LGTM once we move to `this.$translation.dataType` (happy to discuss alternatives) and fix the failing test.

I've changed the failing tests so that it now expects an error being throwed. Despite that, the test still fails...
any pointers??

test-e2e/config-import.js
```diff
@@ -0,0 +1,78 @@
...
+    translations.length,
+    870,
+    're-loading the same config leads to the same number of translations (since they are deleted)'
+  )
+})
+
+test('failing on loading multiple configs in parallel', async (t) => {
+  const manager = createManager('device0', t)
+  const project = await manager.getProject(await manager.createProject())
+  assert.throws(
```
tomasciccola, line 70: This tests still fails, but it shouldn't...

tomasciccola: Its related to node test runner. Since the test suite runner finishes before the test itself, it is marked as failed. I remembered how to solve this, but forgot...

EvanHahn: I think you want to change this to `await assert.rejects`.

EvanHahn:
```suggestion
  await assert.rejects(
```

EvanHahn (changes requested):

test-e2e/config-import.js
tomasciccola: this still fails. I think its because since the test is async, the runner completes before the task. With `brittle` we could do `t.plan(numberOfTests)` so that the runner waited. I don't think the node test suite can do this...

test-e2e/config-import.js
```diff
@@ -0,0 +1,76 @@
...
+
+test('failing on loading multiple configs in parallel', async (t) => {
+  const manager = createManager('device0', t)
+  const project = await manager.getProject(await manager.createProject())
+  await assert.rejects(async () => {
+    await Promise.all([
+      project.importConfig({ configPath: defaultConfigPath }),
+      project.importConfig({ configPath: defaultConfigPath }),
+    ])
+  }, 'loading configs in parallell should throw')
```
EvanHahn, line 75: I think this will fix it:

```suggestion
  const results = await Promise.allSettled([
    project.importConfig({ configPath: defaultConfigPath }),
    project.importConfig({ configPath: defaultConfigPath }),
  ])
  assert.equal(results[0]?.status, 'fulfilled', 'first import should work')
  assert.equal(results[1]?.status, 'rejected', 'second import should fail')
```

The important change was switching `Promise.all` to `Promise.allSettled`. With `Promise.all`, the following would happen:

1. `Promise.all` is called.
1. The first import starts.
1. The second import starts, and fails immediately.
1. The result of `Promise.all` (from step 1) rejects, causing the test to end, and the database to be closed.
1. The first import continues and tries to write to the database, which fails because it's closed.

Changing to `Promise.allSettled` lets both finish before shutting everything down.

EvanHahn (approved):
LGTM once we fix tests.

test-e2e/config-import.js
tomasciccola: owww, nice. That did it. Thanks for maintaining this loong back and forth! I'm gonna merge this

PR #704, opened 2024-06-19
feat: expose creator, coordinator, and member role IDs

### LINKED ISSUE #532: Expose RoleId constants to front end
The [roles are constants used as unique identifiers](https://github.com/digidem/mapeo-core-next/blob/52689b6dc5ed049210c87455c8a6074c5ab0f0e8/src/roles.js#L7).

The front end needs access to these constants in order to match the `roleIds` with each role. 

Currently the front end is[ copying and pasting these constants](https://github.com/digidem/CoMapeo-mobile/pull/205/files#diff-ca3239535e66a9a29ff9649d1ba99896a78b92acfeda726078c45b03acdcc9ba) into the front end from the `@mapeo/core`

EvanHahn:
You can now import `CREATOR_ROLE_ID`, `COORDINATOR_ROLE_ID`, and `MEMBER_ROLE_ID`.

```javascript
import { roles } from '@mapeo/core'

myProject.$member.invite(deviceId, { roleId: roles.MEMBER_ROLE_ID })
```

CoMapeo Mobile [currently hard-codes these values][0].

Note that "weirder" IDs, like `LEFT_ROLE_ID`, are deliberately *not exported right now* because they're easy to add later if needed.

Closes #532.

[0]: https://github.com/digidem/comapeo-mobile/blob/2cfa789360a2faa066a85fde8aad73d5e80a5dfd/src/frontend/sharedTypes/index.ts#L31-L37

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,5 +1,16 @@
+import {
+  CREATOR_ROLE_ID,
+  COORDINATOR_ROLE_ID,
+  MEMBER_ROLE_ID,
+} from './roles.js'
 export { plugin as MapeoStaticMapsFastifyPlugin } from './fastify-plugins/maps/static-maps.js'
 export { plugin as MapeoOfflineFallbackMapFastifyPlugin } from './fastify-plugins/maps/offline-fallback-map.js'
 export { plugin as MapeoMapsFastifyPlugin } from './fastify-plugins/maps/index.js'
 export { FastifyController } from './fastify-controller.js'
 export { MapeoManager } from './mapeo-manager.js'
+
+export const roles = /** @type {const} */ ({
+  CREATOR_ROLE_ID,
+  COORDINATOR_ROLE_ID,
+  MEMBER_ROLE_ID,
+})
--- a/test-e2e/manager-invite.js
+++ b/test-e2e/manager-invite.js
@@ -3,7 +3,9 @@ import assert from 'node:assert/strict'
 import { InviteResponse_Decision } from '../src/generated/rpc.js'
 import { once } from 'node:events'
 import { connectPeers, createManagers, waitForPeers } from './utils.js'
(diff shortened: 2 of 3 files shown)


src/index.js
```diff
@@ -1,5 +1,16 @@
...
+  COORDINATOR_ROLE_ID,
+  MEMBER_ROLE_ID,
+} from './roles.js'
 export { plugin as MapeoStaticMapsFastifyPlugin } from './fastify-plugins/maps/static-maps.js'
 export { plugin as MapeoOfflineFallbackMapFastifyPlugin } from './fastify-plugins/maps/offline-fallback-map.js'
 export { plugin as MapeoMapsFastifyPlugin } from './fastify-plugins/maps/index.js'
 export { FastifyController } from './fastify-controller.js'
 export { MapeoManager } from './mapeo-manager.js'
+
+export const roles = /** @type {const} */ ({
```
EvanHahn, line 12: Curious about feedback on this API.

test-e2e/manager-invite.js
EvanHahn, line 1: Updated a couple of tests to use these new values to ensure things were working.

src/index.js
achou11: not really an improvement, but alternative idea is to use `ROLE_ID` as a namespace and then have more semantic enum values that point to the internal constants e.g.

```ts
export const ROLE_ID = /** @type {const} */ ({
  CREATOR: CREATOR_ROLE_ID,
  COORDINATOR: ...,
  MEMBER: ...,
})
```

achou11 (approved):

src/index.js
EvanHahn: I think I'd like the exports to stay as `CREATOR_ROLE_ID`, `COORDINATOR_ROLE_ID`, and `MEMBER_ROLE_ID`. I think it's a little clearer if you destructure them, especially because we might expose things like `CREATOR` in the future.

If you disagree I'm happy to make a followup change.

PR #712, opened 2024-06-26
feat: detect closed connections with TCP keepalive

EvanHahn:
If two peers connect and one disconnects, the other won't necessarily know.

1. Alex and Charlie connect. Alex sees that they are connected to 1 device.
2. Charlie disconnects from wifi. Alex still sees a connection to 1 device, even though the connection should be broken.

We can fix this in several ways. I think the best is to enable [TCP keepalive], which effectively sends a packet every second asking, "is this connection still on?".

I did not add unit tests for this (I could not think of a good way) but did test manually in CoMapeo with two Android devices.

[TCP keepalive]: https://tldp.org/HOWTO/TCP-Keepalive-HOWTO/overview.html


gmaclennan (approved):
Great, glad to see a solution to this!

PR #715, opened 2024-07-10
feat!: update `@mapeo/schema` to latest version

### LINKED ISSUE #371: Change `createdBy` to `originalVersionId`
## Description

The discussion in #153 about createdBy went back-and-forth between various ideas, and currently `createdBy` is the discoveryId of the core where the document was first created. However, per #136 we want to derive `createdBy` and `updatedBy` records for the client to be device IDs, so that the client can use these to quickly and easily establish which records were created by themselves.

I think it makes more sense for the field that is actually stored in the database to be `originalVersionId` per https://github.com/digidem/mapeo-core-next/issues/153#issuecomment-1717070827. This avoids confusion with the derived field `createdBy` which will be appended at read time and contain the deviceId.

It could be tricky to store `originalVersionId` for the first document when it is created, because generally we do not know the versionId until read time (because it contains the sequence number of the document in the core). I think it's ok to read the length of a core before writing to know what the sequence number will be, but not sure if this is subject to race conditions?

Proposed plan:

1. Don't store `originalVersionId` in the first created record.
2. Store `originalVersionId` in all subsequent edits of the record
3. In `@mapeo/schema`, append `originalVersionId` to the first created record at decode time (like we do with `versionId` - it will be the same)

This introduces a slight complexity for types, but manageable: internally for `create`, `originalVersionId` will not be required. For client types however it is never passed as a write value, and on read it will always be present.

The follow-up to this is #136 to append `createdBy` and `updatedBy` fields which are device IDs. We can look these up at read-time from the core ownership records. The client can then use these fields to establish editing permissions.

EvanHahn:
This patch:

- Removes `createdBy` from documents. Removes `MapeoProject.prototype.$createdByToDeviceId` as a side-effect.

- Returns `originalVersionId` on all documents.

- Updates `@mapeo/schema` to the latest version, which requires a bunch of other changes.

Closes #371. Depends on <https://github.com/digidem/mapeo-schema/pull/191> and <https://github.com/digidem/mapeo-mock-data/pull/7>.

Co-Authored-By: Tomás Ciccola <tciccola@digital-democracy.com>


package.json
```diff
@@ -107,7 +107,7 @@
   "devDependencies": {
     "@bufbuild/buf": "^1.26.1",
     "@mapeo/default-config": "4.0.0-alpha.5",
-    "@mapeo/mock-data": "^1.0.2",
+    "@mapeo/mock-data": "file:mapeo-mock-data-1.0.2.tgz",
```
EvanHahn, line 110: See <https://github.com/digidem/mapeo-mock-data/pull/7>. I'll deploy a new version of the dependency before merging this PR.

package.json
```diff
@@ -157,7 +157,7 @@
     "@fastify/type-provider-typebox": "^4.0.0",
     "@hyperswarm/secret-stream": "^6.1.2",
     "@mapeo/crypto": "1.0.0-alpha.10",
-    "@mapeo/schema": "3.0.0-next.16",
+    "@mapeo/schema": "file:mapeo-schema-3.0.0-next.16.tgz",
```
EvanHahn, line 160: See <https://github.com/digidem/mapeo-schema/pull/191>; same as above.

src/mapeo-project.js
```diff
@@ -558,20 +558,6 @@ export class MapeoProject extends TypedEmitter {
...
-   */
-  async $createdByToDeviceId(createdBy) {
-    const discoveryKey = Buffer.from(createdBy, 'hex')
-    const coreId = this.#coreManager
-      .getCoreByDiscoveryKey(discoveryKey)
-      ?.key.toString('hex')
-    if (!coreId) throw new Error('NotFound')
-    return this.#coreOwnership.getOwner(coreId)
-  }
-
```
gmaclennan, line 574: What's the replacement for this? Don't we need a `$originalVersionIdToDeviceId` for now? Ideally we just append `createdBy` to the read value of observations, but in the meantime I think the frontend relies on this function?

gmaclennan (changes requested):
This all looks good, but see question about the createBy API, and comments on the schema PR

src/mapeo-project.js
EvanHahn: Replaced in ee39a1fccbdfd6ea608cfac15f418d0659117719.

EvanHahn:
I think this is finally ready for another review.

A lot of (unrelated) stuff was required to update `@mapeo/schema` to the latest version.

EvanHahn:
I'm going to merge this because:

- Gregor has reviewed most of this,
- Tomás and I paired on the rest,
- and it's blocking some other work.

PR #720, opened 2024-07-22
feat: you can now leave a project if you're the only member

### LINKED ISSUE #717: User cannot leave project as the only device
Currently if a user tries to leave a project using `mapeoApi.leaveProject(projectId)`, and they are the only user on that project, they are unable to leave a project. The Error `Cannot leave a project as the only device` is thrown.

This is undesired behaviour for a few reasons:
1. We do not warn the user that they are unable to leave a project when they create a project.
2. Difficult for QA testing. Qa tester's usually only have 2 devices. It they want to test the different states of being invited to a project while on a project, they will be blocked.
3. This discourages the user from testing out the creation of a project as they will be trapped in that project unless they invite someone else.

EvanHahn:
Closes #717, which has more background on the motivation for this change.

### DIFF
--- a/src/mapeo-project.js
+++ b/src/mapeo-project.js
@@ -630,11 +630,12 @@ export class MapeoProject extends TypedEmitter {
     return this.#iconApi
   }
 
-  async [kProjectLeave]() {
-    // 1. Check that the device can leave the project
+  /**
+   * @returns {Promise<void>}
+   */
+  async #throwIfCannotLeaveProject() {
     const roleDocs = await this.#dataTypes.role.getMany()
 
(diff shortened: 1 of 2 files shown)


achou11 (approved):
Think the test named `Creator cannot leave project if no other coordinators exist` should be renamed so that it's clear that it's referring to when there are multiple project members. Currently sounds a little conflicting with the feature you're adding here (when it's actually not)

EvanHahn:
Changed the title in 06488d4f8aa88d8545b187ea4015ec53bbd8f9d0. I'll merge if CI passes.

PR #721, opened 2024-07-23
feat: import config metadata and import to `projectSettings`

tomasciccola:
tied to https://github.com/digidem/comapeo-mobile/issues/527

it uses an unpublished version of mapeo-schema with `configName` and `configVersion` in `projectSettings`

I think this PR should be the last to be merged, after updating `@mapeo/schema`, `@mapeo/mock-data`, `@mapeo/default-config`


src/config-import.js
```diff
@@ -270,6 +281,27 @@ async function findTranslationsFile(entries) {
...
+/**
+ * @param {ReadonlyArray<Entry>} entries
+ * @returns {Promise<MetadataFile | undefined>}
+ */
+async function findMetadataFile(entries) {
+  const metadataEntry = entries.find(
+    (entry) => entry.filename === 'metadata.json'
+  )
+  if (!metadataEntry) return
+  //assert(metadataEntry, 'Zip file does not contain metadata.json')
```
tomasciccola, line 293: since metadata, translations and presets files are loaded in parallel (see line 52), failing where it can't find a `metadata.json` file messes assertions in tests.

src/config-import.js
```diff
@@ -479,6 +511,24 @@ function parseIcon(filename, buf) {
   }
 }
 
+/**
+ * @param {Record<string,unknown>} obj
+ * @returns {obj is MetadataFile}
+ */
+function isValidMetadataFile(obj) {
```
tomasciccola, line 518: should we allow extra fields in `metadata.json` I would say since it would mean not needing to update the config builder, and we can ignore values we don't want

EvanHahn:
Let me know when I should take a look at this (it's currently a draft). Happy to review whenever!

tomasciccola:
> Let me know when I should take a look at this (it's currently a draft). Happy to review whenever!

I think there's some doubts I should clear out before moving forward (see stuff asked here -> https://github.com/digidem/comapeo-mobile/issues/527). 
Most of the work is done, just need to fix tests and type errors...

src/config-import.js
EvanHahn: Yes, I think we should allow extra fields. This seems good to me.

tests/fixtures/config/completeConfig.zip
EvanHahn, line 1: _Not_ for this PR, but it'd be nice to be able to see the contents of these in source without having to unzip them manually.

tests/config-import.js
```diff
@@ -83,9 +123,8 @@ test('config import - icons', async () => {
   let config = await readConfig(
     './tests/fixtures/config/invalidIconFilename.zip'
   )
-  /* eslint-disable-next-line */
-  for await (const icon of config.icons()) {
-  }
+
+  await arrayFrom(config.icons())
```
EvanHahn, line 127: Minor: I'd move these to a separate PR, as it's not strictly related. But I don't feel strongly.

EvanHahn (commented):
Draft looks good so far!

tests/fixtures/config/completeConfig.zip
tomasciccola: Yes, that IS a drag. Don't even mention about updating each zip file where data changes. Probably gonna address it in the future (or now in a separate PR). I think having the unzipped files and _maybe_ zip them as a sort of `pre-test` step?

EvanHahn:
Lemme know when I should take a look at this.

tests/fixtures/config/completeConfig.zip
EvanHahn: That's exactly what I'd do. Have them unzipped in the source code and zip them when it's time to run the test. But again, not for this PR.

src/config-import.js
```diff
@@ -36,23 +39,31 @@ const MAX_ICON_SIZE = 10_000_000
...
+    `invalid or missing config file version ${metadataFile.fileVersion}. We support version ${SUPPORTED_CONFIG_VERSION}}`
+  )
 
   return {
     get warnings() {
       return warnings
     },
 
+    get metadata() {
+      return { ...metadataFile, importDate: new Date().toISOString() }
```
gmaclennan, line 64: this will return a different value every time this property is accessed. The date should be set when import happens.

src/config-import.js
```diff
@@ -36,23 +39,31 @@ const MAX_ICON_SIZE = 10_000_000
...
     throw new Error(`Zip file contains too many entries. Max is ${MAX_ENTRIES}`)
   }
   const entries = await zip.readEntries(MAX_ENTRIES)
-  const [presetsFile, translationsFile] = await Promise.all([
-    findPresetsFile(entries),
-    findTranslationsFile(entries),
-  ])
+  const presetsFile = await findPresetsFile(entries)
+  const translationsFile = await findTranslationsFile(entries)
+  const metadataFile = await findMetadataFile(entries)
```
gmaclennan, line 52: why switch to serial vs. parallel?

src/config-import.js
gmaclennan: How about using ajv and JSONSchema for this? We already have them as deps via mapeo-schema. Could be more work than necessary for now.

tomasciccola: (heh, was already doing that: MetadataFile is grabbed from @mapeo/schema)

EvanHahn:
I'm catching up after some time off. I'll plan to review this tomorrow or Wednesday.

tests/fixtures/config/completeConfig.zip
EvanHahn, line 1: Probably makes sense to merge #739, then rebase and update these ZIPs.

EvanHahn (approved):
LGTM but we probably want to merge #739 first.

gmaclennan:
Just a note for myself more than anything to say I did a quick review and all looks good.

PR #722, opened 2024-07-24
fix: auto-download blob namespace

EvanHahn:
Media isn't syncing. This seems to be because we hard-coded the `blob` namespace not to auto-download. This removes that and adds an automated test to check it.

I don't know why we did this, but it seems to have originated in [a 2023 commit][0] (later updated in [another commit][1]).

Should address [comapeo-mobile#514].

[0]: https://github.com/digidem/mapeo-core-next/pull/362/commits/5a47b569b6ad7b417fb866f2c518626fca9b39ad
[1]: https://github.com/digidem/mapeo-core-next/commit/ea442a35ef7113dd98dcb3bd984739f333e314f3
[comapeo-mobile#514]: https://github.com/digidem/comapeo-mobile/issues/514

### DIFF
--- a/src/core-manager/index.js
+++ b/src/core-manager/index.js
@@ -290,7 +290,7 @@ export class CoreManager extends TypedEmitter {
       keyPair,
       encryptionKey: this.#encryptionKeys[namespace],
     })
-    if (namespace !== 'blob' && this.#autoDownload) {
+    if (this.#autoDownload) {
       core.download({ start: 0, end: -1 })
     }
     // Every peer adds a listener, so could have many peers
--- a/test-e2e/sync.js
+++ b/test-e2e/sync.js
@@ -1,9 +1,12 @@
 import test from 'node:test'
 import assert from 'node:assert/strict'
+import * as fs from 'node:fs/promises'
 import { pEvent } from 'p-event'
 import { setTimeout as delay } from 'timers/promises'
+import { request } from 'undici'
 import { excludeKeys } from 'filter-obj'
 import FakeTimers from '@sinonjs/fake-timers'
+import Fastify from 'fastify'
 import { map } from 'iterpal'
 import {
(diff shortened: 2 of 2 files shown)


EvanHahn:
@achou11 Would you mind testing this in CoMapeo when you get a chance? I want to make sure this actually solves the real problem.

achou11:
Yeah seems to fix the specific issue of syncing blobs! Was able to get blobs synced under the following conditions:

1. Device A creates project and adds Device B to project.
2. Device A creates observation (with photo attachment)
3. Device B creates observation (with or without photo).
4. Both devices go to sync screen
5. Both devices detect syncable changes, both enable sync

There's a separate sync issue that I've discovered but I think it's unrelated to this change, so I would lean towards saying that it does fix the issue in question.

achou11 (approved):
infuriatingly small fix

gmaclennan:
> I don't know why we did this, but it seems to have originated in [a 2023 commit](https://github.com/digidem/mapeo-core-next/pull/362/commits/5a47b569b6ad7b417fb866f2c518626fca9b39ad) (later updated in [another commit](https://github.com/digidem/mapeo-core-next/commit/ea442a35ef7113dd98dcb3bd984739f333e314f3)).

@EvanHahn @achou11 this was done this way before we decided to drop selective sync of blobs from MVP. The intention was to control blob download via [`blobStore.download()`](https://github.com/digidem/mapeo-core-next/blob/main/src/blob-store/index.js#L117-L122). However, since we are downloading all blobs for MVP, calling `core.download({ start: 0, end: -1 })` does the same thing as `blobStore.download()` without a filter, e.g. this is an suitable fix for now. When we do add the media management feature to download only previews, then we can revert this change and add functionality to turn on blob downloading according to the user setting.

PR #731, opened 2024-07-30
feat: add unspecified to rpc.proto enums

### LINKED ISSUE #716: Include an Unspecified Value in Enums
For each enum in our protobuf definitions, add `UNSPECIFIED` as the `0` enum, and increment the values for other enum values by `1`.

e.g.

```proto
message InviteResponse {
  enum Decision {
    UNSPECIFIED = 0;
    REJECT = 1;
    ACCEPT = 2;
    ALREADY = 3;
  }
  bytes inviteId = 1;
  Decision decision = 2;
}

```

We either need to map this to a default enum in the backend, or we need to include this in the API type defs, and then add handling to the frontend.

See https://github.com/digidem/mapeo-schema/issues/171 for more background.

tomasciccola:
should close #716 

I _think_ that only `InviteResponse.Decision` enum is used and on `local-peers.js`. From what I could grok, when receiving an `invite-response` we just echo it? What if this decision is `unspecified`? How should we handle that?

On the other hand, this PR is also handling updates to enums in '@mapeo/schema' (see changes on `icon-api`). I can move this to another PR, but I think it makes sense to work on it here so we can centralize the discussion around how to handle `unspecified` fields in enums in different cases


gmaclennan:
> What if this decision is `unspecified`? How should we handle that?

I think rejecting is ok for now, since that avoids any changes to the API / changes required in the front-end. In the future we could add specific error states for an unrecognized or unspecified response, but I think treating it as the invite being rejected is fine for now. There should actually be no case when this is unspecified.

src/icon-api.js
```diff
@@ -123,6 +123,8 @@ export class IconApi {
  * @type {Record<IconVariant['size'], number>}
  */
 const SIZE_AS_NUMERIC = {
+  // We're defaulting for medium size when size is unspecified, is this the correct approach??
+  size_unspecified: 2,
```
tomasciccola, line 127: is this the correct place to make this decision? does defaulting to medium makes sense?

src/member-api.js
```diff
@@ -149,6 +149,8 @@ export class MemberApi extends TypedEmitter {
           return inviteResponse.decision
         case InviteResponse_Decision.UNRECOGNIZED:
           return InviteResponse_Decision.REJECT
+        case InviteResponse_Decision.DECISION_UNSPECIFIED:
+          return InviteResponse_Decision.REJECT
```
EvanHahn, line 153: Nitpick to save one line:

```suggestion
        case InviteResponse_Decision.UNRECOGNIZED:
        case InviteResponse_Decision.DECISION_UNSPECIFIED:
           return InviteResponse_Decision.REJECT
```

src/icon-api.js
EvanHahn: I think this is fine! Probably worth updating/removing the comment to remove the question.

EvanHahn (approved):

gmaclennan (changes requested):
Looks good to me, however I suggest moving the mapeo-schema update for other unspecified handling to a different PR. There are other unspecified enums that need handling, and it would be good to finish and merge this PR just focussed on RPC.

Please add a note to the commit message that this changes the public API to add a new device type `device_type_unspecified`.

tomasciccola:
> I suggest moving the mapeo-schema update for other unspecified handling to a different PR.

Yeah that makes sense

> Please add a note to the commit message that this changes the public API to add a new device type `device_type_unspecified`.

Should I mark this as breaking change? also, should we start marking breaking changes on mapeo-schemas? I would say yes, since it may make easier to integrate on the frontend (wondering if there are standarize ways of describing them, don't have much experience on this...)

tomasciccola:
We have a type error on tests since there's a discrepancy on `DeviceInfo.deviceType`. Since we removed updated (locally installed) version of mapeo/schema with the `device_type_unspecified` now the `rpc` version of `deviceType` is different from the mapeo/schema one. I can probably do some type mangling to add `device_type_unspecified` with a TODO to remove when `mapeo/schema` is updated...

EvanHahn (approved):
> Should I mark this as breaking change? also, should we start marking breaking changes on mapeo-schemas? I would say yes, since it may make easier to integrate on the frontend (wondering if there are standarize ways of describing them, don't have much experience on this...)

Yes, I think so. Merging this with `feat!:` in the commit message should do it.

PR #732, opened 2024-07-30
fix: stale caches for sync state when peer is added

### LINKED ISSUE #724: incorrect sync state when data collected for a project happens before new device is added to project
Reproduction steps:

1. Device A creates project
2. Device A creates observation
3. Device A adds Device B to project
4. Device B detects syncable changes and enables sync. Device A **does not detect syncable changes**

Looking at the sync state for Device A in step (4), this is the value of `data` in the sync state:

```json
{"dataToSync": false, "have": 1, "isSyncEnabled": false, "missing": 0, "want": 0, "wanted": 0}
```

I would've expected `wanted` to be `1` and thus `dataToSync` to be `true`

If Device A or B create another piece of data (e.g. observation) after step (4), then that bypasses the issue and both devices are able to detect syncable changes. Did not confirm if the sync state `data` payload is actually accurate in that case on Device A though.

EvanHahn:
This fixes the following bug:

1. Device A creates a project.
2. Device A creates an observation.
3. Device A reads the sync state.
4. Device A invites Device B.
5. Device A reads the sync state again. It's wrong! It's the same state as in Step 3.

This happened because of two stale state caches, both of which I fixed.

- `NamespaceSyncState`'s cache needed to be invalidated when a peer gets added. I did this by adding an update in `CoreSyncState#addPeer`, which `NamespaceSyncState` "listens to".

- `SyncState`'s cache also needed to be invalidated. I investigated doing this before realizing that its cache only saves a single method call to `NamespaceSyncState.prototype.getState`, which is itself cached. Rather than properly fix the cache (which would add some additional complexity), I simply removed it.

Fixes #724.

### DIFF
--- a/src/sync/core-sync-state.js
+++ b/src/sync/core-sync-state.js
@@ -172,6 +172,7 @@ export class CoreSyncState {
   addPeer(peerId) {
     if (this.#remoteStates.has(peerId)) return
     this.#remoteStates.set(peerId, new PeerState())
+    this.#update()
   }
 
   /**
--- a/src/sync/sync-state.js
+++ b/src/sync/sync-state.js
@@ -3,7 +3,6 @@ import { NAMESPACES } from '../constants.js'
 import { NamespaceSyncState } from './namespace-sync-state.js'
 import { throttle } from 'throttle-debounce'
 import mapObject from 'map-obj'
-import { createMap } from '../utils.js'
 
 /**
  * @typedef {Record<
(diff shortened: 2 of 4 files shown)


gmaclennan (approved):
Thanks for all the research into this, and for finding both a fix and tests for the failing case. Apologies that this code is a bit complicated - I found it hard to simplify! The solution makes sense, and it does seem like we were "over-caching" and I think it's fine to remove that unnecessary cache. At some point I would like to gather some metrics via Sentry about how long the state calculations are taking on low-end devices.

PR #735, opened 2024-07-30
chore: improve types for Hypercore peers

EvanHahn:
*This is a types-only change.*

Before this change, we used `any` for a Hypercore peer's type.

After this change, we use a new type, `HypercorePeer`, designed to match [Hypercore's `Peer` class][0].

Ideally, we would put this in `@digidem/types` or DefinitelyTyped. For now, I think this is an improvement over the `any`s we currently use.

I did this because I'm working with peer-related code and found the `any`s difficult to work with.

[0]: https://github.com/holepunchto/hypercore/blob/de993dc051301f03570df0eb7929c06286745563/lib/replicator.js#L331

### DIFF
--- a/src/core-manager/index.js
+++ b/src/core-manager/index.js
@@ -14,6 +14,8 @@ import { coresTable } from '../schema/project.js'
 import * as rle from './bitfield-rle.js'
 import { CoreIndex } from './core-index.js'
 
+/** @typedef {import('../types.js').HypercorePeer} HypercorePeer */
+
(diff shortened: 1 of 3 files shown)


achou11:
Just something to consider: seems like that class has changed a bit between the version of hypercore we use and the commit you link to (which is latest commit on their main branch at time of writing).

EvanHahn:
Good point. These types are for the version of Hypercore we use, not the version linked.

tomasciccola (approved):

PR #736, opened 2024-07-30
docs: remove useless `@returns` from several functions

EvanHahn:
*This is a minor, JSDoc-only change.*

`@returns` by itself doesn't do much in JSDoc. This either removes those directives or adds their return type.


tomasciccola (approved):
LGTM!

PR #737, opened 2024-07-30
[DO NOT MERGE] Sketch of fix for #268

EvanHahn:
_I recommend reviewing this [with whitespace changes disabled](https://github.com/digidem/mapeo-core-next/pull/737/files?w=1)._

I'm unsure of the right approach for #268. This *incomplete* PR sketches out what I think it should look like.

### DIFF
--- a/src/roles.js
+++ b/src/roles.js
@@ -246,6 +246,16 @@ export class Roles {
   }
 
   /**
+   * @returns {void}
+   */
+  start() {
+    // TODO: Add a capabilities.on('update', updateCapabilities) event. Internally it needs to listen for both new role records, but also new coreOwnership records, because we only know the role of the project creator via their core ownership record (because they own the auth core with the key that matches the project key).
+    this.#dataType.on('updated-docs', (roles) => {
+      console.log('roles updated', roles)
+    })
+  }
+
+  /**
    * Get the role for device `deviceId`.
    *
    * @param {string} deviceId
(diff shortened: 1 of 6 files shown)


src/roles.js
```diff
@@ -249,29 +250,35 @@ export class Roles {
...
-      // The project creator will have the creator role
-      const authCoreId = await this.#coreOwnership.getCoreId(deviceId, 'auth')
-      if (authCoreId === this.#projectCreatorAuthCoreId) {
-        return CREATOR_ROLE
-      } else {
-        // When no role assignment exists, e.g. a newly added device which has
-        // not yet synced role records.
-        return NO_ROLE
-      }
+  async getRole(deviceId, options = {}) {
```
EvanHahn, line 257: I chose to add new behavior to `Roles`, as I felt it was a clean place to put it. Let me know if that seems wrong.

gmaclennan:
The original issue is about keys that are being sent to peers, where-as this sketch seems to be focussed on what to do with keys that are received. E.g. we don't want to send the hypercore keys for anything other than the auth cores unless we actually validate that a peer is part of a project.

Currently as long as a peer has the project key, we send keys for all cores. We also send "haves".

We want to change behaviour so that only auth core keys are sent to peers that have the project key, and then only send keys for cores in other namespaces once we evaluate whether that peer is authorized to sync that namespace. Similarly with sending "haves". This means that we need to check the capabilities and roles before sending this data, and we also need to listen for updates to capabilities and roles, because as auth cores sync, this data can update/change.

Did you see the draft PR for this https://github.com/digidem/mapeo-core-next/pull/390/files? I think that is roughly a reasonable approach, but there are probably a few errors in it and I don't think tests were actually passing.

EvanHahn:
Closing because we don't intend to incorporate this.

PR #739, opened 2024-07-31
chore: use unzipped config fixtures

tomasciccola:
For ease of debugging I created a script that zips every folder in `tests/fixtures/config`, so the idea is to commit uncompressed folders instead of zips.
I'm wondering if we should have some exceptions for this. e.g. `tooBigOfAZip` -> we can keep the zipped file since its smth we're not going to touch further. It makes the script less straight away thought, but wondering how github will deal with a folder with so many entries (I don't want to make the repo slow to navigate...)

Still haven't added those folders (want to double check before). 
Also I used `yazl` instead of the `zip` cli since I don't want to worry about platform compatibility (despite using a bash script would be muuuch simpler)


scripts/build-config-fixtures.js
```diff
@@ -0,0 +1,52 @@
...
+    const filePath = `${path}/${content}`
+    for await (const p of walk(filePath)) {
+      const zipPath = p.replace(`${path}/`, '')
+      zip.addFile(p, zipPath)
+    }
+  }
+  zip.end()
+}
+
+/**
```
EvanHahn, line 40:
```suggestion
/**
 * @param {string} path
 * @returns {AsyncGenerator<string>}
 */
```

scripts/build-config-fixtures.js
```diff
@@ -0,0 +1,52 @@
...
+ */
+async function* walk(path) {
+  const stat = await fs.stat(path)
+  if (stat.isDirectory()) {
+    const dir = await fs.readdir(path)
+    for (const newPath of dir) {
+      yield* await walk(`${path}/${newPath}`)
```
EvanHahn, line 47: Two nitpicks:

1. `await` has no effect here and can be removed.
2. `/` might not work on Windows. Could we use `path.join` instead?

```suggestion
      yield* walk(join(path, newPath))
```

scripts/build-config-fixtures.js
```diff
@@ -0,0 +1,52 @@
...
+import { createWriteStream } from 'node:fs'
+
+const CONFIG_FIXTURES_PATH = new URL(
+  '../tests/fixtures/config',
+  import.meta.url
+).pathname
+const dir = await fs.readdir(CONFIG_FIXTURES_PATH)
+console.log('zipping config fixtures')
+for (const fileOrFolder of dir) {
+  const p = `${CONFIG_FIXTURES_PATH}/${fileOrFolder}`
```
EvanHahn, line 12: This might not work on Windows. Can we use `path.join` instead?

```js
import { join } from 'node:path';
// ...
```

```suggestion
  const p = join(CONFIG_FIXTURES_PATH, fileOrFolder)
```

.gitignore
```diff
@@ -8,3 +8,4 @@ proto/build
 !/drizzle/**/*.sql
 .eslintcache
 docs/api/html/*
+test/fixtures/config/*.zip
```
EvanHahn, line 11: What happens to all the existing fixtures when we do this?

scripts/build-config-fixtures.js
```diff
@@ -0,0 +1,52 @@
+import yazl from 'yazl'
+import fs from 'node:fs/promises'
```
EvanHahn, line 2: Nitpick: this will work but I think we should import `fs` as a module, not its default export.

```suggestion
import * as fs from 'node:fs/promises'
```

scripts/build-config-fixtures.js
```diff
@@ -0,0 +1,52 @@
...
+import { createWriteStream } from 'node:fs'
+
+const CONFIG_FIXTURES_PATH = new URL(
+  '../tests/fixtures/config',
+  import.meta.url
+).pathname
+const dir = await fs.readdir(CONFIG_FIXTURES_PATH)
```
EvanHahn, line 9: You could instead pass the `withFileTypes` option, which will give back an array of `fs.Dirent` objects instead of strings. Then you don't need to call `fs.stat` below, because you can check if it's a directory right there.

Take it or leave it!

EvanHahn (changes requested):

.gitignore
tomasciccola: Before merging I'll `git rm` all the zips, so with this they should be untracked, right?

scripts/build-config-fixtures.js
tomasciccola: done!

EvanHahn:
I probably won't be able to review this until I get back on August 12. Just FYI!

.gitignore
EvanHahn: Yeah. Could you do that before we merge?

EvanHahn (changes requested):
LGTM but we should unzip all the ZIPs before merging.

PR #745, opened 2024-08-13
feat: `MemberInfo` - expose when a project member joined the project

### LINKED ISSUE #744: Surface when a project member joined a project
## Description

There are cases where we want to surface when a project member was added to a project. Our `MemberApi` currently does not surface that information in the return types of the relevant methods (`getById()` and `getMany()`). See https://github.com/digidem/comapeo-mobile/issues/588 as an example of where this info will be used

## Tasks

- [ ] Update the `MemberInfo` type to include a field that provides information about when a member joined a project. Something along the lines of `joinedAt`, `joinedDate`, `addedAt`, `addedDate`, etc. Think this can remain as a `string` and just point to the `createdAt` field of the relevant document.

- [ ]  Update `MemberApi` so that `getById()` and `getMany()` adhere to the updated `MemberInfo` interface

tomasciccola:
should close #744

### DIFF
--- a/src/member-api.js
+++ b/src/member-api.js
@@ -30,6 +30,7 @@ import { ROLES, isRoleIdForNewInvite } from './roles.js'
  * @prop {import('./roles.js').Role} role
  * @prop {import('@mapeo/schema').DeviceInfo['name']} [name]
  * @prop {import('@mapeo/schema').DeviceInfo['deviceType']} [deviceType]
+ * @prop {import('@mapeo/schema').DeviceInfo['createdAt']} [joinedAt]
  */
 
 export class MemberApi extends TypedEmitter {
@@ -255,6 +256,7 @@ export class MemberApi extends TypedEmitter {
 
       result.name = deviceInfo.name
       result.deviceType = deviceInfo.deviceType
+      result.joinedAt = deviceInfo.createdAt
     } catch (err) {
       // Attempting to get someone else may throw because sync hasn't occurred or completed
       // Only throw if attempting to get themself since the relevant information should be available
@@ -290,6 +292,7 @@ export class MemberApi extends TypedEmitter {
 
           memberInfo.name = deviceInfo?.name
           memberInfo.deviceType = deviceInfo?.deviceType
+          memberInfo.joinedAt = deviceInfo?.createdAt
         } catch (err) {
           // Attempting to get someone else may throw because sync hasn't occurred or completed
           // Only throw if attempting to get themself since the relevant information should be available
--- a/test-e2e/members.js
+++ b/test-e2e/members.js
@@ -25,8 +25,7 @@ test('getting yourself after creating project', async (t) => {
   const deviceInfo = manager.getDeviceInfo()
   const project = await manager.getProject(await manager.createProject())
 
-  const me = await project.$member.getById(project.deviceId)
-
+  const { joinedAt: _, ...me } = await project.$member.getById(project.deviceId)
   assert.deepEqual(
     me,
     {


test-e2e/members.js
```diff
@@ -25,8 +25,7 @@ test('getting yourself after creating project', async (t) => {
   const deviceInfo = manager.getDeviceInfo()
   const project = await manager.getProject(await manager.createProject())
 
-  const me = await project.$member.getById(project.deviceId)
-
+  const { joinedAt: _, ...me } = await project.$member.getById(project.deviceId)
```
EvanHahn, line 28: Can we test the value of `joinedAt` somehow?

src/member-api.js
```diff
@@ -30,6 +30,7 @@ import { ROLES, isRoleIdForNewInvite } from './roles.js'
  * @prop {import('./roles.js').Role} role
  * @prop {import('@mapeo/schema').DeviceInfo['name']} [name]
  * @prop {import('@mapeo/schema').DeviceInfo['deviceType']} [deviceType]
+ * @prop {import('@mapeo/schema').DeviceInfo['createdAt']} [joinedAt]
```
EvanHahn, line 33: Is there any way we can make this required? It seems like it should always be defined, at least theoretically. Might require a larger change.

EvanHahn (changes requested):

test-e2e/members.js
tomasciccola, line 28: all tests fail 'cause they don't expect a `joinedAt` wondering options:
1. remove `joinedAt` from the comparison and leave it at that
2. remove `joinedAt` from the comparison but assert that the field should be there (despite not caring about matching values)
3. another way of getting the expected value for `joinedAt` to add to the expected object

tomasciccola: yeah, was thinking about this, see -> https://github.com/digidem/mapeo-core-next/pull/745#discussion_r1715519655

src/member-api.js
tomasciccola: yeah, was going through that. The easiest way I'm thinking of is turning the result ([here](https://github.com/digidem/mapeo-core-next/blob/8784111e312eb46de46adbd7a7725220c6133b1d/src/member-api.js#L245)) into a `SetOptional<MemberInfo,'joinedAt'>` when declared. Is there a better type wizardry for this?

tomasciccola: Mhm, I think is basically the same making it an optional field. I always struggle with ts when declaring an object of a type where some members are added some lines later. Is there a way to signal this to the type system?? (basically, 'I know `joinedAt` is missing, just look some lines later!)

test-e2e/members.js
EvanHahn: I like the 2nd option. Maybe we could also check that the timestamp is near the current timestamp?

EvanHahn:
Looks like there's a test failure, too.

src/member-api.js
tomasciccola: In this case I think the issue is that calling `this.#coreOwnership.getCoreId` and `this.#dataTypes.deviceInfo.getByDocId` can throw. So if that happens then `joinedAt` would be `undefined` (as with the other optionals of `MemberInfo`)

EvanHahn: Ah, I guess if everything else can be undefined then this is okay too.

test-e2e/members.js
```diff
@@ -25,8 +25,16 @@ test('getting yourself after creating project', async (t) => {
...
-  const me = await project.$member.getById(project.deviceId)
+  const { joinedAt, ...me } = await project.$member.getById(project.deviceId)
+  assert.ok(joinedAt, 'has a `joinedAt` field')
 
+  const joinedAtUnix = new Date(joinedAt).getTime()
+  const now = new Date().getTime()
+
+  assert.ok(
+    Math.abs(joinedAtUnix - now) < 1000,
+    'time of joined proyect is close to now'
```
EvanHahn, line 36: Nitpick:

```suggestion
    'time of joined project is close to now'
```

EvanHahn (approved):

PR #749, opened 2024-08-14
chore: use `Object.hasOwn()`

EvanHahn:

`Object.hasOwn()` replaces `Object.prototype.hasOwnProperty()` in modern JavaScript. Its advantages can be seen in the following code snippet:

```javascript
// Still works even if you overwrite `hasOwnProperty`:
const overwritten = { hasOwnProperty: () => true }
Object.hasOwn(overwritten, "foo")
// => false
overwritten.hasOwnProperty("foo")
// => true (incorrect)

// Still works even if the object has a null prototype:
const nullPrototype = Object.create(null)
nullPrototype.foo = "bar"
Object.hasOwn(nullPrototype, "foo")
// => true
nullPrototype.hasOwnProperty("foo")
// TypeError: nullPrototype.hasOwnProperty is not a function
```

This replaces all uses of `Object.prototype.hasOwnProperty()` with `Object.hasOwn()`.

I think this is a useful change on its own, but will also help in a future change.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
     "target": "es2019",
-    "lib": ["es2020"],
+    "lib": ["es2022"],
     "noImplicitAny": true,
     "noImplicitThis": true,
     "strictNullChecks": true,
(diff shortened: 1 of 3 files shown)


tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
     "target": "es2019",
-    "lib": ["es2020"],
+    "lib": ["es2022"],
```
EvanHahn, line 4: `Object.hasOwn` works in our version of Node, so I believe this is a safe change to make.

tomasciccola (approved):

PR #753, opened 2024-08-16
feat!: expose device specific sync state

### LINKED ISSUE #743: Expose sync state of syncing peers
## Description

Currently sync state (`have`, `want`, `wanted`, `missing`, `dataToSync`) is calculated across all connected peers.

We need to expose sync state for only peers that are syncing, in addition to the sync state of all connected peers.

This is necessary so that the front-end can show progress of sync with syncing peers separately from progress of sync across all peers. Example user story:

1. Devices A, B and C are together on the same wifi network and are all connected to each other.
2. Devices A and B sync, intentionally device C does not sync.
3. Devices A and B would expect to see progress of sync between them both, and an indication that their sync with each other is complete.

Instead, currently, because sync state is based on all connected peers, it appears to devices A & B that sync has never completed - because "sync completion" according to our current sync state, only happens when all connected peers have synced.

We want to keep the current sync state (of all connected peers) but add the additional state of sync state with the syncing peers.

## Tasks

- [ ] Add additional sync states (`have`, `want`, `wanted`, `missing`, `dataToSync`) for syncing peers only, for the `initial` and `data` sync types.

Related: #742

tomasciccola:
should close #743

### DIFF
--- a/TODO.txt
+++ b/TODO.txt
@@ -1,0 +1,7 @@
+- can we remove `have`?
+- can we remove `missing`?
+- can we remove `dataToSync`?
+- docs
+- maybe clean up `reduceSyncState`
+- change names?
+- all remaining TODOs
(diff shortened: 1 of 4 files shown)


EvanHahn (commented):
Partial review.

EvanHahn:
Cleaned this up and opened #757.

EvanHahn:
Closing because we merged #757.

PR #764, opened 2024-08-25
fix: sync should only happen when both sides are enabled

### LINKED ISSUE #762: Sync can unexpectedly occur with devices that have not enabled it
## Description

I'm running into a situation where I can get data sync to occur between two devices even though only one of them has data sync enabled. Using myself and Evan as an example, here are the reproduction steps:

1. Andrew invites Evan to a project as a participant.
2. Andrew and Evan collect some data independently.
3. Andrew and Evan go to the sync screen and both press the button to enable data sync.
4. Andrew and Evan leave the sync screen, which disables data sync in this case. EDIT: in this case, it's important that Andrew leaves first (see https://github.com/digidem/mapeo-core-next/issues/762#issuecomment-2305271367). And that's how I've been doing this in practice
5. Andrew **or** Evan (or both) collect more data independently.
6. **Andrew** goes to the sync screen and presses the button to enable sync. Notice that syncing starts between Andrew and Evan, even though Evan hasn't pressed the button to enable sync again.

The expectation here is that in 6, Andrew should be in some pending state, waiting for any other connected devices (in this case, Evan) to also enable sync. Instead, sync immediately happens without Evan opting in.

Important notes:

- In step 1, not sure if it matters what role is chosen for Evan.
  - ~I've only tried this when Evan is a participant, but might be worth trying to reproduce the issue for when Evan is added as a coordinator.~ EDIT: being a coordinator did not make a difference.

- Steps 2,3, and 4 are important. If we do this without those steps, the issue does not seem to occur. In technical terms, it seems like a prior data sync has to have happened before encountering this issue.

- In step 6, it's vital that Andrew enables sync again. If it were Evan that did that instead, this issue does not occur. 
  - ~This suggests that something about being the project creator somehow bypasses the expected sync "etiquette".~ EDIT: This is probably wrong, see https://github.com/digidem/mapeo-core-next/issues/762#issuecomment-2305271367 for a more likely explanation
  - ~Going back to the first note, wondering if the issue could be reproduced if Evan were a coordinator and re-enabled sync instead of Andrew.~ EDIT: being a coordinator did not make a difference

## Tasks

- [ ]

EvanHahn:
Closes #762.


gmaclennan:
I added a commit which _I think_ fixes this. I don't particularly like this, and I feel like this code is fragile, given that we are doing something not supported by hypercore, but I have written some more robust unit tests for this now.

I extracted `unreplicate` into its own function, and added more comprehensive unit tests for different ways that cores could be unreplicated, e.g. in different orders, and with delays. I found that a delay between unreplicating (e.g. if one peer replicates after the other) caused errors, due to the peer being removed from `core.peers` when the other side closes the channel, which results in not finding the channel that needs to be closed.

The solution I found for this is that rather than iterating through peers attached to a core to get the channel reference (to close it), I iterate through the channels of a protomux. I'm not sure what the consequences are if there are multiple channels with the same id on a protomux, but I don't think we do that so I think it's fine. With the unit tests for `unreplicate` and the fuzzy sync tests (thanks @EvanHahn !) I think this is ok, even though I don't like how I'm implementing this!

gmaclennan:
Correction: not yet fixed! Fuzz tests are showing errors still with re-sync not working as expected. I am unable to reproduce locally, I wonder if there is a way we can make fuzz tests reproducible?

gmaclennan:
The issue _could_ be because of edge cases (race conditions) around starting and stopping sync in the same tick. Opening a protomux channel (which happens when replicating a core) is asynchronous, and I'm not sure what happens when we close a protomux channel in the same tick as when it is opening. Our sync api `start()` should probably be async, and implement [`start-stop-state-machine`](https://github.com/digidem/start-stop-state-machine) so that stop waits for start before attempting to stop.

EvanHahn:
> The issue _could_ be because of edge cases (race conditions) around starting and stopping sync in the same tick. Opening a protomux channel (which happens when replicating a core) is asynchronous, and I'm not sure what happens when we close a protomux channel in the same tick as when it is opening. Our sync api `start()` should probably be async, and implement [`start-stop-state-machine`](https://github.com/digidem/start-stop-state-machine) so that stop waits for start before attempting to stop.

Filed #786.

EvanHahn:
Gregor and I paired on this so I will merge without additional review.

PR #765, opened 2024-08-26
feat: delete old config docs only after successful new config import

### LINKED ISSUE #761: don't delete presets, fields and translations if the config-import fails
Before importing a config, we delete all presets, fields and observations. But if the import fails, then you end up with nothing.
We should actually check that the config was successful  before deleting the older docs.

tomasciccola:
should close #761 
The approach I took is basically set a bunch of promises that only get resolved after a successfull `configImport`. I _think_ this approach is right


src/mapeo-project.js
```diff
@@ -727,10 +727,10 @@ export class MapeoProject extends TypedEmitter {
 
     try {
       // check for already present fields and presets and delete them if exist
-      await deleteAll(this.preset)
-      await deleteAll(this.field)
+      const deletePresetsPromise = deleteAll(this.preset)
+      const deleteFieldsPromise = deleteAll(this.field)
       // delete only translations that refer to deleted fields and presets
-      await deleteTranslations({
+      const deleteTranslationsPromise = deleteTranslations({
```
EvanHahn, line 733: I don't think this is right. This will kick off the deletions, and it will only _wait_ for them if config import completes...but it will still delete them.

I think you need to only delete things at the very end.

EvanHahn (changes requested):

src/mapeo-project.js
```diff
@@ -876,25 +897,26 @@ async function deleteAll(dataType) {
...
-        try {
-          const toDelete = await opts[docRefType].getByDocId(docRef.docId)
-          shouldDelete = toDelete.deleted
-        } catch (e) {
-          opts.logger.log(`referred ${docRef.docId} is not found`)
-        }
-        if (shouldDelete) {
-          await opts.translation.delete(docId)
-        }
+  translations.forEach(async ({ docRefType, docRef, docId }) => {
```
EvanHahn, line 906: I think we need to `await` this.

src/mapeo-project.js
```diff
@@ -842,6 +842,24 @@ export class MapeoProject extends TypedEmitter {
...
+        fieldsToDelete.map(async (docId) => {
+          await this.field.delete(docId)
+        })
+      )
+
+      await Promise.all(
+        translationsToDelete.map(async (docId) => {
+          await this.$translation.dataType.delete(docId)
+        })
+      )
```
EvanHahn, line 861: Could we do all of these in parallel?

EvanHahn (changes requested):
Could we test that (1) old data is properly deleted (2) nothing is deleted when config import fails?

We might already have these tests somewhere.

src/mapeo-project.js
tomasciccola: doesn't seem to change anything, and the linter complains that 'await has no effect on this type of expression' ...

tomasciccola:
> Could we test that (1) old data is properly deleted (2) nothing is deleted when config import fails?
> 
> We might already have these tests somewhere.

missed this comments. Gonna add this checks first thing tomorrow and re-ask for review!

test-e2e/config-import.js
```diff
@@ -64,6 +64,65 @@ test('config import - load and re-load config manually', async (t) => {
...
+      configPath: defaultConfigPath,
+    })
+    nPresets = (await project.preset.getMany()).length
+    await project.importConfig({
+      configPath: './tests/fixtures/config/validConfig.zip',
+    })
+    await project.importConfig({
+      configPath: defaultConfigPath,
+    })
+    assert.equal(
```
tomasciccola: This is a failing test that I'm not being able to solve. I basically:
* load default config and take note of the number of presets
* load another config
* load the default config again
* Expect the number of presets to be the same as whe the first default config was loaded
The actual number of presets is `nPresetsDefaultConfig + nPresetsValidConfig`. 
If I run it in a loop, I get the same number every time (so presets are being deleted...)

tomasciccola: This was just me misunderstanding Promises. Before I did:
`Promise.all([[presetsToDelete], [fieldsToDelete], [translationsToDelete]].flat())`
(basically flatting the list of stuff to delete and passing it to `Promise.all`)
Now I do:
`Promise.all([Promise.all(presetsToDelete), Promise.all(fieldsToDelete), Promise.all(translationsToDelete)])`

I think I'm not doing deletions of `presets`, `fields`, `translations` in parallel now (only within themselves, so `preset` deletion is in parallel, `field` deletion too, etc)

@EvanHahn any thought?

src/mapeo-project.js
```diff
@@ -831,7 +831,7 @@ export class MapeoProject extends TypedEmitter {
           )
         } else {
           throw new Error(
-            `docRef for preset or field with name ${name} not found`
+            `docRef for ${value.docRefType} with name ${name} not found`
```
EvanHahn, line 834: 👍🏻

tests/fixtures/config/validConfig/presets.json
EvanHahn, line 1: Out of curiosity, why did this file change?

test-e2e/config-import.js
```diff
@@ -64,6 +64,83 @@ test('config import - load and re-load config manually', async (t) => {
...
+  const projectId = await manager.createProject()
+  const project = await manager.getProject(projectId)
+  // load default config
+  await project.importConfig({ configPath: defaultConfigPath })
+
+  const nPresets = (await project.preset.getMany()).length
+  const nFields = (await project.field.getMany()).length
+  const nTranslations = (await project.$translation.dataType.getMany()).length
+
+  // load wrong config
```
EvanHahn, line 117: Nit: this isn't just a _wrong_ config, it's a bogus path.

```suggestion
  // load a non-existent config
```

src/mapeo-project.js
```diff
@@ -886,25 +913,26 @@ async function deleteAll(dataType) {
...
+      } catch (e) {
+        opts.logger.log(`referred ${docRef.versionId} is not found`)
       }
-    })
-  )
+      if (doc) {
+        toDelete.push(docId)
+      }
+    }
+  })
```
EvanHahn, line 934: Sorry for being unclear earlier. We need to await all of these.

```suggestion
  await Promise.all(
    translations.map(async ({ docRefType, docRef, docId }) => {
      if (docRefType === 'field' || docRefType === 'preset') {
        let doc
        try {
          doc = await opts[docRefType].getByVersionId(docRef.versionId)
        } catch (e) {
          opts.logger.log(`referred ${docRef.versionId} is not found`)
        }
        if (doc) {
          toDelete.push(docId)
        }
      }
    })
  )
```

Also, should we make `toDelete` a `Set`? It's possible for multiple translations to refer to the same field/preset, right?

EvanHahn (approved):
LGTM other than my small comments.

tests/fixtures/config/validConfig/presets.json
tomasciccola: two updates to that file:
1. `key` -> `tagKey`
2. added color
Since I'm now counting loaded presets on that config, I wanted to have it correctly loaded

src/mapeo-project.js
tomasciccola: yep, right. Changed on the last commit

PR #766, opened 2024-08-26
chore: use standard Namespace type

EvanHahn:
This is a types-only change.

The `Namespace` type is pretty central, but was stuffed away in `core-manager/index.js`. Its type was also inferred in many cases.

Now it's in a more central spot: `types.ts`.

I think this is useful on its own, but it should make a future change easier.

### DIFF
--- a/src/core-ownership.js
+++ b/src/core-ownership.js
@@ -9,9 +9,13 @@ import mapObject from 'map-obj'
 import { discoveryKey } from 'hypercore-crypto'
 import pDefer from 'p-defer'
 import { NAMESPACES } from './constants.js'
-
 /**
- * @typedef {import('./types.js').CoreOwnershipWithSignatures} CoreOwnershipWithSignatures
+ * @import {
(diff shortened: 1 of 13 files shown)


src/core-ownership.js
```diff
@@ -9,9 +9,13 @@ import mapObject from 'map-obj'
...
 import { NAMESPACES } from './constants.js'
-
 /**
- * @typedef {import('./types.js').CoreOwnershipWithSignatures} CoreOwnershipWithSignatures
+ * @import {
+ *   CoreOwnershipWithSignatures,
+ *   CoreOwnershipWithSignaturesValue,
+ *   KeyPair,
+ *   Namespace
+ * } from './types.js'
```
EvanHahn, line 18: I was adding an import from `types.js` so I figured I'd fix _all_ the imports in this file. I did that in several other places too.

achou11 (approved):

PR #769, opened 2024-08-26
fix: respect all data namespaces for sync auto-stop

### LINKED ISSUE #768: Sync shouldn't auto-stop while blobs are syncing
We reset the auto-stop timer if we receive any new data since our last sync update:

https://github.com/digidem/mapeo-core-next/blob/e8e1d7a70d8910910044c16f264a8bc91a92e89a/src/sync/sync-api.js#L191-L196

However, this doesn't respect the `blob` namespace (or any other data namespaces we add in the future). We should fix this so that sync doesn't auto-stop when blobs are syncing.

EvanHahn:
We reset the auto-stop timer if we receive any new data since our last sync update. However, we were only respecting the `data` namespace and forgot about the `blob` namespace.

This fixes that.

Closes #768.


src/constants.js
```diff
@@ -13,6 +13,11 @@ export const NAMESPACES = /** @type {const} */ ([
 /** @type {ReadonlyArray<Namespace>} */
 export const PRESYNC_NAMESPACES = ['auth', 'config', 'blobIndex']
 
+/** @type {ReadonlyArray<Namespace>} */
+export const DATA_NAMESPACES = NAMESPACES.filter(
+  (namespace) => !PRESYNC_NAMESPACES.includes(namespace)
+)
```
EvanHahn, line 19: This was removed recently in 2b36edd45f82729d5df841937bc915e8d5d3adc7, and I'm adding it back same as it was.

test-e2e/sync.js
```diff
@@ -333,13 +346,25 @@ test('auto-stop', async (t) => {
 
   await clock.tickAsync(9000)
 
-  const observation2 = await invitorProject.observation.create(
```
EvanHahn, line 336: Pseudocode for this diff:

```diff
-createObservation()
+createBlob()

 checkThatAutostopTimeoutIsUpdated()
```

gmaclennan (approved):

PR #774, opened 2024-08-26
chore: uninstall unused light-my-request dependency

EvanHahn:
This dependency seems unused, so we can remove it.

It looks like it was added in 8e64abdf736b31d1a13a1d9791fd9f677118d8d2 but never used.


achou11 (approved):
Used transitively by fastify but fastify specifies it as a prod dep anyways, so unless we need it pinned to a specific version, this changes seems reasonable

PR #776, opened 2024-08-26
fix: update `streamx` and move it to production dependencies

EvanHahn:
This commit:

- Moves `streamx`, which is used outside development, into production dependencies (from `devDependencies` to `dependencies`).
- Updates `streamx` and `@types/streamx` to the latest versions.


achou11 (commented):
where is this being used outside of the development context?

achou11:
ah nvm i didn't read my search results correctly

https://github.com/digidem/mapeo-core-next/blob/9b4cdf67c381dcfaea45b4102515e08825f7e45d/src/blob-api.js#L3

achou11 (approved):

PR #778, opened 2024-08-26
chore: remove unused patch-package dependency

EvanHahn:
Removing this has two benefits:

1. This simplifies the install for dependents. For example, the frontend has needed to patch itself `@mapeo/core` in the past.

2. The usual benefits of removing a dependency: decreased file size, smaller attack surface, etc.

We can always add it back if we need it again.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -45,7 +45,6 @@
         "p-defer": "^4.0.0",
         "p-event": "^6.0.1",
         "p-timeout": "^6.1.2",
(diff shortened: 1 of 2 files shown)


achou11 (approved):
Seems like this should've been removed some time ago 😄

EvanHahn:
Indeed, I forgot to remove it back in 9a523e559aa48c0844ad3ca7d5474d37a7173681!

PR #779, opened 2024-08-26
chore: remove unused random-bytes-readable-stream dependency

EvanHahn:
This dependency is unused, so we can remove it.

It looks like it was added in aa30f7504029cd975cb44cda397227aff906a8af but never used.


tomasciccola:
can you resolve the conflicts before merging?

EvanHahn:
Done!

PR #782, opened 2024-08-27
fix: check for icon file referenced by preset.icon

### LINKED ISSUE #773: Importing a config with a missing icon should fail
It's possible to import a preset that references a missing icon, which doesn't yield any errors or warnings.

This happens in our code today. For example, `tests/fixtures/config/validPreset/presets.json` references two icons that don't exist. This import should fail, or at least include a warning.

- [ ] Decide whether this should fail completely (@EvanHahn's preference) or just include a warning
- [ ] Write tests for this situation
- [ ] Handle this situation

tomasciccola:
should close #773 
If you import a config where there's a preset with a field `icon: 'someIconName'`, we now check that there's a matching file in `icons/someIconName-*`, if not, we throw (but we can instead add a warning)


src/config-import.js
```diff
@@ -193,6 +187,26 @@ export async function readConfig(configPath) {
...
+          iconEntries.map((icon) => {
+            const matches = path
+              .basename(icon.filename)
+              .match(/([a-zA-Z--0-]+)-/)
+            if (matches) {
+              const [_, name] = matches
+              return name
+            }
+          })
+        )
```
EvanHahn, line 201: Could we move this higher up so we only compute it once?

src/config-import.js
```diff
@@ -193,6 +187,26 @@ export async function readConfig(configPath) {
             presetValue[key] = preset[key]
           }
         }
+        // check that preset references existing icon
+        const iconFilenames = new Set(
+          iconEntries.map((icon) => {
+            const matches = path
+              .basename(icon.filename)
+              .match(/([a-zA-Z--0-]+)-/)
```
EvanHahn, line 195: This regex has several problems. For example, it doesn't match `ok1-x.png`, and matches `ok./-x.png`. Could we repurpose the filename parsing from `parseIcon` instead?

EvanHahn (changes requested):

src/config-import.js
```diff
@@ -193,6 +199,14 @@ export async function readConfig(configPath) {
             presetValue[key] = preset[key]
           }
         }
+        if ('icon' in preset && typeof preset.icon === 'string') {
+          if (!iconFilenames.has(preset.icon)) {
+            throw new Error(
+              `preset references icon with name ${preset.icon} but file doesn't exist`
+            )
+          }
+        }
```
EvanHahn, line 208: This has two problems:

1. It throws an error, rather than adding a warning.
2. `preset.icon` is required, but this doesn't check that.

Fixed this in #791.

src/config-import.js
```diff
@@ -472,16 +486,23 @@ function translateMessageObject(warnings) {
   }
 }
 
+/**
+ * @param {ReadonlyArray<Entry>} entries
+ */
+function getIconEntries(entries) {
+  return entries
+    .filter((entry) => entry.filename.match(/^icons\/([^/]+)$/))
+    .sort((icon, nextIcon) => icon.filename.localeCompare(nextIcon.filename))
```
EvanHahn, line 495: Nit: I don't think we need to sort these.

```suggestion
  return entries.filter((entry) => entry.filename.match(/^icons\/([^/]+)$/))
```

(This diff removes the `.sort` and then re-formats it with Prettier. The `.filter` call is unchanged.)

src/config-import.js
```diff
@@ -177,6 +172,17 @@ export async function readConfig(configPath) {
...
+      // check that preset references existing icon
+      const iconFilenames = new Set(
+        iconEntries.map((icon) => {
+          const matches = path.basename(icon.filename).match(ICON_NAME_REGEX)
+          if (matches) {
+            const [_, name] = matches
+            return name
+          }
+        })
+      )
```
EvanHahn, line 184: Because the `map` function can return `undefined`, this this variable is `Set<undefined | string>` instead of what you probably intended, which is `Set<string>`. I opened #790 to fix that.

src/config-import.js
tomasciccola: mhm, I think we do need to be sorted, because we're yielding. 
And that happens when the icon name has changed. If they are unsorted, then we risk getting a new variant for an icon that has already been yielded, I _think_

EvanHahn (approved):
Thanks for all the back-and-forth!

PR #783, opened 2024-08-27
feat: only send/receive core keys with permission

EvanHahn:
Partly addresses #268.


tests/helpers/core-manager.js
```diff
@@ -50,6 +50,20 @@ const destroyStream = (stream) =>
     stream.destroy()
   })
 
+/**
+ * @param {CoreManager} cmToAdd
+ * @param {CoreManager} cmToReceive
+ * @returns {Promise<void>}
+ */
+async function addWriterCores(cmToAdd, cmToReceive) {
```
gmaclennan, line 58: nit @EvanHahn the naming is slightly confusing, because this is actually adding the cores "owned" by one instance to another instance, but as "read" cores, not writer cores. But I'm not sure what a better naming would be.

EvanHahn: Renamed in #795.

PR #785, opened 2024-08-27
fix: only mark sync state enabled when entire group is enabled

EvanHahn:
Imagine the following scenario:

- *auth* namespace sync is enabled.
- *config* namespace sync is disabled.
- *blobIndex* namespace sync is enabled.

Previously, this would report the initial namespace sync state as *enabled*. Now, this will report *disabled*, because every namespace needs to be on.


gmaclennan (approved):
This has got quite hard to follow, but the code is correct and looks like it will work to me.

PR #792, opened 2024-08-28
test: speed up core manager unreplicate tests

EvanHahn:
_This change is not urgent._

On my machine, this cuts the time from ~12 seconds to ~1 second.


EvanHahn:
Seems like this is flaky. Abandoning!

PR #803, opened 2024-09-02
chore: update mapeo/schema

tomasciccola:
Additionally, I've added `deviceType.selfHostedServer` on proto/rpc.proto since it was introduced on last schema definition


ErikSin (approved):
Thanks

PR #805, opened 2024-09-02
feat: add `remoteDetectionAlert` dataType

### LINKED ISSUE #806: Add alerts data type

tomasciccola:
this installs a local version of mapeo-schema with `remoteDetectionAlert` schema.

should close #806


EvanHahn (commented):
LGTM once we (1) release a new `@mapeo/schema` (2) fix the type errors.

Could also be good to add a test that runs a `remoteDetectionAlert` through the whole system to make sure it works. Maybe just create one and make sure we can read it.

tomasciccola:
I'm getting a type error on the `test-e2e/project-crud.js`, where I added a fixture for `remoteDetectionAlert`. I don't know why it expects its field `metadata` to have a `manualLocation`. Its basically confusing it with an observation. I can add that field to please the type checker since `metadata` in `remoteDetectionAlert` allows for any value for any string key, but it feels finicky.

On the other hand, I can't `npm i` now, because `mapeo/schema` is a peerDependency in `mock-data` so the local version installed in this PR doesn't match the version on mock-data...

EvanHahn (changes requested):

tomasciccola:
@EvanHahn I'm having a couple of type errors on `test-e2e/project-crud.js` which I'm not sure how to interpret...

tomasciccola:
I think something related to ts's structural typing is tripping us here. Since 'observation' and 'remoteDetectionAlert' both have a `metadata` field I think it expects `remoteDetectionAlert.metadata` to be the same as `observation.metadata`? 
e.g. adding `manualLocation` to `remoteDetectionAlert` fixes all type errors... @EvanHahn any clue?

tomasciccola:
So, I after [this PR](https://github.com/digidem/comapeo-core/pull/873) was merged I updated the branch but I still get what is basically the same type error: we have to schemas with a field with the same name (`{observation,remoteDetectionAlert}.metadata`). It expects `remoteDetectionAlert.metadata` to have `manualLocation` which actuall belongs to `observation`.
I _think_ I integrated the changes right (basically adding `remoteDetectionAlert` as a `case` in `create`, `createWithMockData`, `update` and `getUpdatedFixture`) but the issue still persists :'(

EvanHahn (changes requested):
This still needs a new migration (rather than deleting the old one), but I think is ready to go otherwise.

EvanHahn (approved):

PR #807, opened 2024-09-02
fix: remove `selfHostedServer` from `deviceInfo.deviceType` rpc

tomasciccola:
this is a regression from previous release of core. 
We originally added `selfHostedServer` to `DeviceInfo.deviceType` to match schema definition. But we don't want to be able to open rpc connections to a remote server; so instead we just mangle the types to omit that enum member were appropiate

### DIFF
--- a/proto/rpc.proto
+++ b/proto/rpc.proto
@@ -38,7 +38,6 @@ message DeviceInfo {
     mobile = 1;
     tablet = 2;
     desktop = 3;
-    selfHostedServer = 4;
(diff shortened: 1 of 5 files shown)


gmaclennan (approved):
This is fine for now. We may want to allow `setDeviceInfo` to set as `selfHostedServer`, but not advertise to peers, when we finish the server, but @EvanHahn can deal with that later - I couldn't get the types to work.

PR #822, opened 2024-09-11
chore: prepare for first stable release

### LINKED ISSUE #752: Rename to `@comapeo/core`, bump to v1

tomasciccola:
should close #752 
Additionally we are:
* changing the package name to `@comapeo/core`

We could also:
* Edit readme to reflect current state of the project (I wouldn't lose much time on this now, just maybe removing the `Early development` disclaimer and correct `npm` command)
* Edit authors? We can point different authors (probably can automate this so it reflects commits authors) or just put "Awana Digital" (simpler but hides individual authors)

This PR is also made so that we can test OPTICS for releasing a stable version. 
We should have an NPM_TOKEN on the project before merging the OPTICS_RELEASE so we can actually publish the package. 

On the other hand, I'm wondering if PRs merged after this, would be included in the release (I think yes?), but if not, we should make sure we're not missing anything (I _think_ not)


EvanHahn (approved):
LGTM.

- Could you add an issue to update the readme? (Or just update it if that's faster.)
- I'll clean up Authors right now.
- Could you make sure `CHANGELOG.md` is ready to go? I _think_ we probably just want to delete it entirely.

tomasciccola:
This needs first the publishing of mapeo-schema v1 before merging

PR #823, opened 2024-09-11
chore: semver on release flow

tomasciccola:
Since the idea is to follow semver, I've went ahead and only allowed (and uncommented) `patch`, `minor` and `version` on semver options. We may disagree with this and still have a prerelease (but I think we may not want this). 
Also I commented some traces of `prerelease` related to naming the release on github and npm. If approve, I'll remove those commented lines

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -6,14 +6,14 @@ on:
       semver:
         description: 'The semver to use'
         required: true
-        default: 'prerelease'
+        default: 'patch'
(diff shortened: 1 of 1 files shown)


EvanHahn (approved):
LGTM.

I would personally _keep_ prerelease stuff, but we can add it back as soon as we need it.

tomasciccola:
cool, then I'm leaving the commented lines there just in case we want to re-introduce some of this in the future

PR #829, opened 2024-09-16
chore!: update to latest schema version

EvanHahn:
This updates `@mapeo/schema` to 3.0.0-next.27. I expect this to be the final version before `@comapeo/schema@1.0.0`.

This update required several changes:

- Updating dev dependencies of `@mapeo/schema`: `@mapeo/mock-data` and `@mapeo/default-config`.

- Removing unused `isInitialProject` from project settings.

- Allowing translation `regionCode`s to be undefined (but not `null`).

- Requiring `deviceType` on device info.


drizzle/client/meta/_journal.json
```diff
@@ -5,8 +5,8 @@
     {
       "idx": 0,
       "version": "5",
-      "when": 1725381616476,
-      "tag": "0000_chemical_star_brand",
+      "when": 1726514276007,
+      "tag": "0000_bumpy_carnage",
```
achou11, line 9: apt name 😄

src/config-import.js
```diff
@@ -475,7 +478,7 @@ function translateMessageObject(warnings) {
             if (
               !validate('translation', {
                 ...value,
-                docRef: { docId: '', versionId: '' },
+                docRef: { docId: 'placeholder', versionId: 'placeholder' },
```
achou11, line 481: don't have enough context for this, but just making sure this is intentional?

src/mapeo-manager.js
```diff
@@ -741,15 +738,24 @@ export class MapeoManager extends TypedEmitter {
...
     const row = this.#db
       .select()
       .from(localDeviceInfoTable)
       .where(eq(localDeviceInfoTable.deviceId, this.#deviceId))
       .get()
-    return { deviceId: this.#deviceId, ...row?.deviceInfo }
+    return {
+      deviceId: this.#deviceId,
+      deviceType: 'device_type_unspecified',
+      ...row?.deviceInfo,
```
achou11, line 757: just checking - this is relying on the object spread being done at the end? i.e. the desired outcome is something like:

```js
deviceType: row?.deviceInfo.deviceType || 'device_type_unspecified',
```

src/config-import.js
```diff
@@ -330,11 +330,14 @@ function translationsForLanguage(warnings) {
...
   return function* (lang, languageTranslations) {
-    const { language: languageCode, region: regionCode } = parseBCP47(lang)
+    const parsed = parseBCP47(lang)
+    const { language: languageCode } = parsed
     if (!languageCode) {
       warnings.push(new Error(`invalid translation language ${lang}`))
       return
     }
+    let { region: regionCode } = parsed
+    regionCode ||= undefined
```
gmaclennan, line 340: nit: I assume this is to catch regionCode being `null`. This syntax is new/obscure enough that I think this warrants a comment, or changing to code to something more explicit like `if (regionCode === null) regionCode = undefined`

gmaclennan (approved):

achou11 (approved):
Couple of small comments but mostly due to lacking some context

src/config-import.js
EvanHahn: Yes. Similar to something we do elsewhere in the file:

https://github.com/digidem/comapeo-core/blob/745604ac4ffb33857e7fd5dc623f05057f23729e/src/config-import.js#L193

src/mapeo-manager.js
EvanHahn: Yup, exactly.

src/config-import.js
EvanHahn: It's to catch `regionCode` being falsy. `''`, `null`, and `undefined` should all be coerced to `null`.

This is because [`regionCode` has to be between 2 and 3 characters when defined](https://github.com/digidem/comapeo-schema/blob/74a3d0da01282e664653924533bab836f933961f/schema/translation/v1.json#L54-L59); the empty string is not valid.

PR #839, opened 2024-09-18
fix: more lenient initial sync

EvanHahn:
Fixes the weird invite bug we've been seeing (<https://github.com/digidem/comapeo-mobile/issues/671>):

https://github.com/user-attachments/assets/e584116f-1f51-44b4-a522-252543f98fc2

When you join a project, you wait for some data to sync, such as the project config. That effectively happens in two steps:

1. If sync is already done, return.
2. Otherwise, listen for sync state updates, and return when sync is done.

Before this change, these two steps answered "is sync done?" in slightly different ways. After this change, they answer that in the same way.

Here's a diff in pseudocode:

```diff
 function waitForInitialSync() {
   if (!wantsDataFromOtherPeers()) return 'done'

   onSyncState(() => {
-    if (isSyncCompletelyFinished()) {
+    if (!wantsDataFromOtherPeers()) {
       return 'done'
     }
   })
 }
```

More specifically, this makes both parts check "is sync done?" by looking at the `auth` and `config` namespaces. If they want nothing but have something, we consider them done with initial sync.

### DIFF
--- a/src/mapeo-manager.js
+++ b/src/mapeo-manager.js
@@ -49,10 +49,11 @@ import {
   kRescindFullStopRequest,
 } from './sync/sync-api.js'
 /** @import { ProjectSettingsValue as ProjectValue } from '@comapeo/schema' */
-/** @import { SetNonNullable } from 'type-fest' */
+/** @import { ReadonlyDeep, SetNonNullable } from 'type-fest' */
 /** @import { CoreStorage, Namespace } from './types.js' */
 /** @import { DeviceInfoParam } from './schema/client.js' */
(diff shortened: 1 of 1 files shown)


gmaclennan (approved):
Code is correct. Not sure if the naming and the nested function is the clearest. 

Maybe something like `hasRrceivedAllData(syncSyate.auth) && hasRrceivedAllData(syncState.config)`

But much of a muchness really.

gmaclennan (changes requested):
maybe hold on this given #640, and we can change this in the other direction -> less lenient initial sync to actually check everything has synced

EvanHahn:
Closing because we solved this a different way.

PR #840, opened 2024-09-19
fix: pin `corestore` dep to exact 6.8.4

gmaclennan:
This is the cause of bugs in the v1.0.0 release candidate. The `corestore` dependency should have been pinned to the exact version that we are testing here (v6.8.4), but it was not.

The mobile app had already been using `corestore@6.15.9`, which had changed the hypercore dep to `hypercore@10.29.0`. This appeared to still work, to some extent. Commit https://github.com/digidem/comapeo-mobile/commit/e06a7e4554f1ac8f47f5bb407d328bf37f8919cb included a `src/backend/package-lock.json` change that changed the corestore dep to `corestore@6.18.4` and its subdep to `hypercore@10.37.22`.

Unfortunately we rely on hypercore internals, and in addition hypercore is not completely stable across minor versions, so changes to the hypercore version break things. We have thorough tests on comapeo-core to check that all the internals of hypercore that we depend on are tested and working, however the omission of pinning corestore resulted in a different version of hypercore being used in the mobile app.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -25,7 +25,7 @@
         "big-sparse-array": "^1.0.3",
         "bogon": "^1.1.0",
         "compact-encoding": "^2.12.0",
-        "corestore": "^6.8.4",
+        "corestore": "6.8.4",
         "debug": "^4.3.4",
         "dot-prop": "^9.0.0",
         "drizzle-orm": "^0.30.8",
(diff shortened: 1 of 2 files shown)


gmaclennan:
I have tested this with the latest release candidate https://github.com/digidem/comapeo-mobile/tree/0b8dd9f06aa7b0d28b4143ea3ad0b49c5c6af382 and things are working, e.g. the invite bug is fixed.

EvanHahn (approved):
Wowie.

PR #846, opened 2024-09-19
chore: integrate npm-lockfile-changes action

achou11:
Uses a fork (https://github.com/rvanvelzen/npm-lockfile-changes) of the original action (https://github.com/codepunkt/npm-lockfile-changes) because the original does not seem to work with v3 lockfiles.

Uses the action's default values for inputs for now (https://github.com/marketplace/actions/npm-lockfile-changes#-inputs)

Also omits the `permissions` configuration since we don't use dependabot (https://github.com/marketplace/actions/npm-lockfile-changes#the-action-fails-on-the-dependabot-pull-requests). My understanding is that we don't need to specify as a result, but could be wrong.

Out of scope, but also updated this workflow to only run when the pull request changes the `package-lock.json` file, as opposed to anytime a pull request is opened. Happy to revert if not desired.

### DIFF
--- a/.github/workflows/lockfile.yml
+++ b/.github/workflows/lockfile.yml
@@ -21,8 +21,14 @@ jobs:
     steps:
       - name: Check out a copy of the repo
         uses: actions/checkout@v4
+      - name: Create GitHub App Token
+        uses: actions/create-github-app-token@v1
+        id: app-token
+        with:
+          app-id: ${{ vars.LOCKFILE_BOT_APP_ID }}
+          private-key: ${{ secrets.LOCKFILE_BOT_PRIVATE_KEY }}
       - name: NPM Lockfile Changes
         # The original doesn't support v3 lockfiles so we use a fork that adds support for them
         uses: rvanvelzen/npm-lockfile-changes@6fded38b5a054f5ab49efd6850668e796f780604
         with:
-          token: ${{ secrets.GITHUB_TOKEN }}
+          token: ${{ steps.app-token.outputs.token }}
--- a/.github/workflows/lockfile.yml
+++ b/.github/workflows/lockfile.yml
@@ -1,5 +1,9 @@
 name: Check Lockfile
-on: pull_request
+on:
+  pull_request:
+    paths:
+      - 'package-lock.json'
+
 jobs:
   lockfile_version:
     name: Lockfile version check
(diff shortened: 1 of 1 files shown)


.github/workflows/lockfile.yml
```diff
@@ -1,8 +1,12 @@
 name: Check Lockfile
-on: pull_request
+on:
+  pull_request:
+    paths:
+      - 'package-lock.json'
```
achou11, line 5: highlighting this since it's kind of out of scope for the PR and could be reverted if not desired. Figured it would be worth doing to save a few CPU cycles though 😄

.github/workflows/lockfile.yml
```diff
@@ -11,3 +15,14 @@ jobs:
...
           version: 3
+  lockfile_changes:
+    name: Lockfile changes check
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out a copy of the repo
+        uses: actions/checkout@v4
+      - name: NPM Lockfile Changes
+        # The original doesn't support v3 lockfiles so we use a fork that adds support for them
+        uses: rvanvelzen/npm-lockfile-changes@main
```
gmaclennan, line 26: For the sake of security, maybe pin to a specific commit?

.github/workflows/lockfile.yml
gmaclennan: I'm fine with including this.

gmaclennan (approved):
Looks good, however as a post-merge-task I think we should setup a Github app for this and generate a token, see https://github.com/digidem/release-automations/blob/7eddab75ddec8324c17d532ae3b96bbd60182c60/.github/workflows/build-rc.yml#L62-L67 - we can create a new Github App for this workflow. This gives more refined control over permissions, and controls the name and icon of the actor creating the comments on the PR.

.github/workflows/lockfile.yml
achou11: addressed via 6a775843d49aa466f5f13e919f63b03771a334db

achou11:
> Looks good, however as a post-merge-task I think we should setup a Github app for this and generate a token, see https://github.com/digidem/release-automations/blob/7eddab75ddec8324c17d532ae3b96bbd60182c60/.github/workflows/build-rc.yml#L62-L67 - we can create a new Github App for this workflow. This gives more refined control over permissions, and controls the name and icon of the actor creating the comments on the PR.

Addressed with Gregor's help and code changes in db56e7d1bf293f1a60728c4726653cbe0212e8d8

gmaclennan:
Seems like the action is failing, not clear why

achou11:
> Seems like the action is failing, not clear why

Where are you seeing this? Wouldn't expect this to run on this PR due to the change related to only running based on the lockfile changing

achou11:
@gmaclennan  see the example of the integration working in https://github.com/digidem/comapeo-core/pull/853

PR #848, opened 2024-09-19
chore: remove unused property from CoreManager

EvanHahn:
CoreManager was storing a `#projectKey` property which was never used. This removes it.


gmaclennan (changes requested):
Not sure if this is the right fix for this, this is kind of a reminder that we should actually be using these properties to stop use of this class when it is closing or closed, and to stop errors when trying to close a second time while it is closing.

EvanHahn:
I updated this PR to keep one of the properties but remove one that was definitely still unused.

gmaclennan (approved):

PR #854, opened 2024-09-19
chore: only attempt to run release workflow when merging PR

achou11:
Does a couple of things:

- Makes sure the pull request is targeting the main branch. Not sure if there are cases where we want this thing to run when that isn't the case

- Makes sure that the pull request is actually merged. Before, the workflow was being run when closing a pull request (and not merging), which adds more noise than necessary

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -19,9 +19,11 @@ on:
           # - premajor
   pull_request:
     types: [closed]
+    branches: [main]
 
 jobs:
   release:
+    if: github.event.pull_request.merged == true
     runs-on: ubuntu-latest
     permissions:
       contents: write


.github/workflows/release.yml
```diff
@@ -19,9 +19,12 @@ on:
           # - premajor
   pull_request:
     types: [closed]
+    branches:
+      - main
 
 jobs:
   release:
+    if: github.event.pull_request.merged == true
```
achou11, line 27: Docs for this: https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#running-your-pull_request-workflow-when-a-pull-request-merges

EvanHahn (approved):

PR #855, opened 2024-09-22
build: move doc building into same step as release

EvanHahn:
I *think* this will fix the errors we're seeing on `release` CI.

See [the optic-release docs][0].

[0]: https://github.com/nearform-actions/optic-release-automation-action/blob/2bdd6c56fa196bd0c8f4f5a4374fc39b3ff78c8a/action.yml#L13-L15

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -34,19 +34,6 @@ jobs:
         uses: actions/setup-node@v3
         with:
           node-version: 18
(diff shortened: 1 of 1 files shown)


achou11 (approved):
Seems like it would do the trick! At the very least, looks to be an improvement from what's currently there.

https://optic.nearform.com/recipes#release-a-new-github-action-version-with-a-custom-build-command

PR #856, opened 2024-09-23
test: always connect peers more realistically

EvanHahn:
_I recommend [reviewing this with whitespace changes disabled](https://github.com/digidem/comapeo-core/pull/856/files?w=1)._

Our tests currently connect `MapeoManager`s in two ways:

1. By starting peer discovery servers and connecting. This is similar to what the real app does.
2. By manually creating streams and connecting them in tests. This is less realistic.

This commit removes the second way because:

- it is less realistic
- it lets us remove some test-only code in the `src/` directory
- it will make an upcoming change easier
- it *used* to be significantly faster, but that's no longer true

I also tried to fix a possible (test-only) race condition, which *could* have been a reason for the less realistic option:

1. Start connecting peers by calling `connectPeers()`. This begins the process of starting peer discovery servers.
2. Disconnect them by calling the callback returned by `connectPeers()`.
3. The peer discovery servers start, and begin connecting. *This is bad* because we already wanted to disconnect!

### DIFF
--- a/src/mapeo-manager.js
+++ b/src/mapeo-manager.js
@@ -45,7 +45,6 @@ import { LocalPeers } from './local-peers.js'
 import { InviteApi } from './invite-api.js'
 import { LocalDiscovery } from './discovery/local-discovery.js'
 import { Roles } from './roles.js'
-import NoiseSecretStream from '@hyperswarm/secret-stream'
 import { Logger } from './logger.js'
 import {
   kSyncState,
@@ -53,6 +52,7 @@ import {
   kRescindFullStopRequest,
 } from './sync/sync-api.js'
 /** @import { ProjectSettingsValue as ProjectValue } from '@comapeo/schema' */
+/** @import NoiseSecretStream from '@hyperswarm/secret-stream' */
 /** @import { SetNonNullable } from 'type-fest' */
 /** @import { CoreStorage, Namespace } from './types.js' */
 /** @import { DeviceInfoParam } from './schema/client.js' */
@@ -81,7 +81,6 @@ export const DEFAULT_ONLINE_STYLE_URL =
   'https://demotiles.maplibre.org/style.json'
 
 export const kRPC = Symbol('rpc')
-export const kManagerReplicate = Symbol('replicate manager')
 
 /**
  * @typedef {Omit<import('./local-peers.js').PeerInfo, 'protomux'>} PublicPeerInfo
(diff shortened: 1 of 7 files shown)


test-e2e/utils.js
```diff
@@ -27,50 +27,28 @@ const clientMigrationsFolder = new URL('../drizzle/client', import.meta.url)
 
 /**
  * @param {readonly MapeoManager[]} managers
- * @returns {() => void}
```
EvanHahn, line 30: This return type was was just incorrect before, I believe.

test-e2e/utils.js
```diff
@@ -27,50 +27,28 @@ const clientMigrationsFolder = new URL('../drizzle/client', import.meta.url)
...
-              stream.destroy()
-            })
-          )
-        )
+export function connectPeers(managers) {
+  let requestedDisconnect = false
+
+  for (const manager of managers) {
+    manager.startLocalPeerDiscoveryServer().then(({ name, port }) => {
+      if (requestedDisconnect) return
```
EvanHahn, line 37: This check (hopefully) fixes a small possible race condition.

src/mapeo-project.js
```diff
@@ -575,29 +574,6 @@ export class MapeoProject extends TypedEmitter {
...
-        const protomux =
-          /** @type {import('protomux')<import('@hyperswarm/secret-stream')>} */ (
-            replicationStream.noiseStream.userData
-          )
-        this.#syncApi[kHandleDiscoveryKey](discoveryKey, protomux)
-      },
-    })
-    return replicationStream
-  }
-
```
EvanHahn, line 600: No longer needed either.

src/mapeo-manager.js
```diff
@@ -208,22 +207,6 @@ export class MapeoManager extends TypedEmitter {
...
-   *
-   * @param {boolean} isInitiator
-   */
-  [kManagerReplicate](isInitiator) {
-    const noiseStream = new NoiseSecretStream(isInitiator, undefined, {
-      keyPair: this.#keyManager.getIdentityKeypair(),
-    })
-    return this.#replicate(noiseStream)
-  }
-
```
EvanHahn, line 226: No longer needed.

EvanHahn:
This is in draft because I'm not sure we want to do this. I worry there's some historical context I'm missing.

gmaclennan (approved):
looks good, good to simplify things.

PR #858, opened 2024-09-23
chore: add another branch filter for release workflow

achou11:
Adds a branch filter for any release-related branch so that the workflow runs more precisely.

https://optic.nearform.com/docs/github-action/examples#using-the-branches-filter

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -19,7 +19,7 @@ on:
           # - premajor
   pull_request:
     types: [closed]
-    branches: [main]
+    branches: [main, 'release/**']
 
 jobs:
   release:


gmaclennan:
This filter is for the [base branch of the PR](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#running-your-pull_request-workflow-based-on-the-head-or-base-branch-of-a-pull-request) (e.g. what you are merging into). Adding this filter would make this workflow also run if we made pull requests into a branch that matched `release/**`. We don't really do that, so not a huge issue, but nitpick: the change makes the workflow run "less precisely" rather than "more precisely" (it could be triggered in more circumstances)

achou11:
ah whoops - didn't realize it pointed to the branch that's being merged into - will close this then

PR #859, opened 2024-09-24
chore: index with discoveryId not discoveryKey

gmaclennan:
This reduces the crypto operations required during
indexing, which should increase indexing
performance.

Depends on:

- https://github.com/digidem/multi-core-indexer/pull/25
- https://github.com/digidem/comapeo-schema/pull/267


RangerMauve:
Is this still relevant?

gmaclennan:
> Is this still relevant?

Yes, I believe it is. Currently the indexing pipeline calculates the discovery key (a [Blake2b hash](https://github.com/holepunchto/hypercore-crypto/blob/8f1709da52a08823c371fffcc6d51d6932473992/index.js#L148) of the core key) twice for each document:

1. When [decoding blocks in the index writer](https://github.com/digidem/comapeo-core/blob/cf1043f0bb1fe1bc7ef469e649eaae6b73ff5f77/src/index-writer/index.js#L77)
2. When [deriving the versionId](https://github.com/digidem/comapeo-core/blob/cf1043f0bb1fe1bc7ef469e649eaae6b73ff5f77/src/datastore/index.js#L109) to check for pending writes.

It is also calculated for each [coreOwnership](https://github.com/digidem/comapeo-core/blob/cf1043f0bb1fe1bc7ef469e649eaae6b73ff5f77/src/core-ownership.js#L160) and [deviceInfo](https://github.com/digidem/comapeo-core/blob/cf1043f0bb1fe1bc7ef469e649eaae6b73ff5f77/src/mapeo-project.js#L1598) record, but this is probably less of an impact because there are fewer of these.

Hashing, although fast, is not insignificant when done hundreds or thousands of times e.g. during indexing after an initial sync or project creation, and my hypothesis is that hash operations are significantly slower on older mobile devices without hardware crypto optimizations, although I don't have evidence for that yet!

Because during indexing in multi-core-indexer, the discovery key is already calculated for each core, and doesn't need to be calculated for each block/document, passing this through the indexing pipeline removes the need for these crypto opts.

PR #870, opened 2024-09-26
chore: fix release workflow to run when using workflow dispatch

achou11:
Fixes an issue introduced by #854 so that the workflow runs when using the manual workflow dispatch approach.


gmaclennan:
Maybe check in the other direction - the merged check is only relevant if the event is PR closed.

achou11:
> Maybe check in the other direction - the merged check is only relevant if the event is PR closed.

mind clarifying? not entirely sure what the suggestion is

.github/workflows/release.yml
```diff
@@ -23,7 +23,7 @@ on:
 
 jobs:
   release:
-    if: github.event.pull_request.merged == true
+    if: ${{github.event_name == 'workflow_dispatch' || github.event.pull_request.merged}}
```
gmaclennan, line 26:
```suggestion
    if: ${{github.event_name != 'pull_request' || github.event.pull_request.merged}}
```
Just because `event.pull_request.merged` only exists for pull_request events. Doesn't change the functionality in this case, but might do if we want to trigger a new release via a different trigger in the future.

EvanHahn (approved):

PR #881, opened 2024-10-01
feat: support serving Style Map Package maps

### LINKED ISSUE #827: Add support for serving Styled Map Package maps
### Description

We want to be able to serve offline maps that follow the format described in https://github.com/digidem/styled-map-package/. 

Ideally, I think this eventually replaces the "legacy" Mapeo approach that we currently support. For now, can probably build this functionality side-by-side.

### Tasks

- [ ] Expose functionality for importing and retrieving information about an SMP-based map.
  - importing = persisting the file path of the SMP file on the device to core.
  - based on app requirements, we'll need the following information about an SMP map:
    - date added
    - storage used
    - name

- ~Implement a Fastify plugin for serving the SMP file based on the location from ^~

achou11:
Closes #827 

Still blocked on the following:

- ~Need to add a couple of more tests I think.~ DONE

- ~I didn't upload the fixture that I was using for the tests since I wasn't sure if we had a particular approach for it just yet. I've been using an SMP generated from the command listed [here](https://github.com/digidem/styled-map-package?tab=readme-ov-file#usage)~ Added a fixture as described

- Types are not published with `styled-map-package`, as that's blocked on https://github.com/digidem/styled-map-package/pull/21

Implementation notes:

- I made the decision to avoid using the fastify plugin implementation that comes with `styled-map-package` due to the following reasons:

    - it doesn't  account for registering it using a [prefix](https://fastify.dev/docs/v4.28.x/Reference/Routes/#route-prefixing). Part of me wonders if i'm holding Fastify wrong and maybe my previous setup for the other plugins is somewhat flawed, but at least when trying to use the plugin in the same way and do similar tests, things started to misbehave when using prefixing.
    - it required me to create an extra `Reader` instance because it doesn't expose the internal one that it creates at all. not the worst thing but figured that given the other issue, I might as well address this issue by re-working the plugin.

- open to input about the name of the decorator and the plugin that this PR introduces. I currently call it `comapeoSmp` and `comapeo-styled-map-package`, respectfully.

- not entirely too happy about the plugin implementation i have right now, which is kind of related to my thinking that maybe i'm holding fastify wrong a bit. open to feedback on cleaner approaches.


src/fastify-plugins/maps/styled-map-package.js
```diff
@@ -0,0 +1,113 @@
+import fp from 'fastify-plugin'
+// @ts-expect-error Need to publish types for module
+import { Reader } from 'styled-map-package'
+
+/** @import {FastifyPluginAsync} from 'fastify' */
+// @ts-expect-error Need to publish types for module
+/** @import {Resource} from 'styled-map-package/reader' */
+// @ts-expect-error Need to publish types for module
+/** @import {SMPStyle} from 'styled-map-package/types' */
```
achou11, line 9: Just noting that these ignores are allowing tests to run without immediately failing

src/fastify-plugins/maps/index.js
```diff
@@ -76,7 +81,19 @@ async function routes(fastify, opts) {
...
+      {
+        const style = await fastify.comapeoSmp.getStyle().catch(() => {
+          fastify.log.warn('Cannot read styled map package archive')
+          return null
+        })
+
+        if (style) {
+          return style
+        }
+      }
```
achou11, line 94: I made the assumption that we want to give highest precedence to an SMP if possible

achou11:
One thing that this PR doesn't fully address is the criteria listed in the issue:

> based on app requirements, we'll need the following information about an SMP map:
>
>    - date added
>    - storage used
>    - name

Not really sure what the best way to expose this information will be. 

- storage used = easy to do client side
- date added = technically can do client side
- name = not really sure what the best thing to do here is. Is this something that should be derived from the style.json within the SMP, or can it just be based on the name of the SMP file? My guess is that it shouldn't be the latter, which makes the implementation a little trickier to figure out.

gmaclennan (commented):
I had a read through this today. This is the first time I am reading the code from the maps plugin, so some comments are not directly related to this PR, but we may want to consider them when implementing this. I'll try my best to separate the two, but some decisions in the current implementation mean that we need to do a lot more work here, so it may be worth while fixing those now rather than doing extra work.

**PR specific comments**

I do recommend using the `styled-map-package` server plugin. With the latest fixes I think it should do everything needed in a self-contained way, and if it doesn't we should fix that upstream.

For the client requirements of date added, storage used, and name, I think we should add a route outside the plugin. We don't need access to the `Reader` instance for it, we can do something like:

```js
fastify.get('/smp/info', async (req, reply) => {
  const stats = await fs.stat(smpFilepath)
  const style = await fetch('/smp/style.json').json()
  return {
    created: stats.ctime,
    size: stats.size
    name: style.name || path.parse(smpFilepath).name
  }
})
fastify.register(smpPlugin)
```

For selecting a style to server based on availability we have a few options:

1. Either use decorator methods or fetch upstream to load style.json into memory, then return that.
2. No decorator methods, and proxy everything (local or remote) either with our own `fetch` or with [`@fastify/reply-from`](https://github.com/fastify/fastify-reply-from).
3. `HEAD` requests then `302` redirects.

My preference is (3) because I think it's the simplest (pseudo-code):

```js
fastify.get('style.json', async (req, reply) => {
  const res = await fetch({ method: 'HEAD', url: styleUrl }).catch(noop)
  if (res && res.status === 200) reply.status(302).setHeader('location', styleUrl)
})
```

**Architecture opinions**

- I think we should just drop `static-maps` right now. We haven't released anything that supports this, and we haven't communicated any support for it, and keeping it just adds maintenance burden and the need to handle backwards compatibility.
- I think we're making things hard for ourselves by exposing each of the "sub-plugins" outside of this module - we're leaking our implementation details. I think the maps plugin should be a single plugin. Otherwise we don't control the order the sub-plugins are registered, and we need to reach into internals by decorating everything.
- Why not register the maps plugin in Mapeo Manager rather than export it as a separate thing?
- Rather than trying to decorate the plugin with the `getStyleJsonUrl()` method, maybe move that into `MapeoManager` like with the blob and icon servers (easy to do if registering internally)
- I think it's worth using styled map package for the fallback map too, to simplify things, see https://github.com/digidem/comapeo-fallback-smp

There may well be very valid reasons for not choosing to do things this way. I'm happy to jump on a quick call about it, rather than you needing to write down a justification for everything here.

achou11:
@gmaclennan thanks for the feedback! i think it all makes sense and I honestly don't have any strong arguments against the suggestions regarding architecture. At the time I initially worked on the maps plugin, I wasn't aware of the options for style selection that you've outlined, which is how i ended up with this weird setup 😄

fully agree that I would just re-use the smp fastify plugin from upstream, i just didn't want to be blocked on the issues I originally encountered (since you were focusing on other things). Those have been addressed at this point (once published) so I'll need to update the PR with that.

As for updating the general implementation of the maps plugin, would you suggest I do that in a separate PR that this PR stacks on top of? feels like trying to do that work after updating and merging this one would be annoying vs the other way around

achou11:
Closing this PR in favor of a completely new one that will significantly simplify how the maps plugin works, based on the suggestions provided above

PR #883, opened 2024-10-02
chore: change lockfile bot to update comments

gmaclennan:
This proposed change is to the lockfile bot which comments on pull requests with changes to the lockfile. I propose changing the bot config to update the original comment when the PR is updated, rather than creating new comments, in order to reduce "noise" in the PR comment thread. Open to arguments for keeping it as-is with new comments for each commit!


achou11 (approved):
based on the docs, this should be configured to update the comment by default which is why i didn't specify it 🤦‍♂️

https://github.com/marketplace/actions/npm-lockfile-changes#-inputs

another lesson in preferring to be explicit when configuring stuff like this...

gmaclennan:
yeah looks like a bug in their `action.yml` file, they specify the default as `default: 'true'`, when, if github's own actions are anything to go by, should be `default: true`.

PR #887, opened 2024-10-02
fix: sync state with disconnected peers

gmaclennan:
When a peer has disconnected, and another peer connects, sync state never "completes" because it thinks the disconnected peer is waiting for data.


gmaclennan:
Addressed in #891

PR #889, opened 2024-10-03
chore: debug logging improvements

gmaclennan:
Several improvements to debug logging, to improve our ability to parse the logs and gather useful information:

- Fix a memory issue in the `PeerSyncController` logger, that was creating a new `Debugger` instance (by calling `Logger.create()`) each time anything was logged.
- Fixed a memory issue in `PeerSyncController` which was adding a `peer-remove` event listener every time a core is replicated, and never removing the listener. This was solely for debugging. Now only adds the listener if debugging is enabled.
- Log the namespace of cores that are replicated or unreplicated.
- Change the function signatures in `PeerSyncController` and expose the `CoreRecord` of the creator core, to enable this namespace logging.
- Removed unnecesary logging when handling discovery keys in `PeerSyncController`.
- `SyncApi`: only log change of sync enabled state, rather than logging it on every state update.
- `SyncApi`: Remove superflous logging for when cores are added (adding cores is logged by core manager).
- Explicitly include the `supports-color` dev dependency, which `debug` uses if available to support more colors (it was using this anyway, as a transitive dependency included by an unrelated module, but this makes the dependency more explicit).
- Remove unhelpful "No project settings" log in `MapeoProject`
- Add logging to `MapeoManager.#waitForInitialSync` to show why sync is pending.
- Configure the `debug` module to color debug lines according to the deviceId, rather than a different color for each namespace. Simplifies identifying which debug lines come from each peer in tests.
- `Logger.log` is now an instance of `debug.Debugger`, with its properties such as `Debugger.enabled`. Before it was just a function that wrapped `debug.Debugger`.

### DIFF
--- a/src/sync/peer-sync-controller.js
+++ b/src/sync/peer-sync-controller.js
@@ -32,6 +32,7 @@ export class PeerSyncController {
   #downloadingRanges = new Map()
   /** @type {SyncStatus} */
   #prevSyncStatus = createNamespaceMap('unknown')
+  /** @type {import('debug').Debugger} */
   #log
 
   /**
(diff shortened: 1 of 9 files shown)


src/sync/peer-sync-controller.js
```diff
@@ -43,25 +44,27 @@ export class PeerSyncController {
...
-        `[%h] ${formatter}`,
-        protomux.stream.remotePublicKey,
-        ...args,
-      ])
-    }
+    const log = Logger.create('peer', logger).log
+    this.#log = Object.assign(
+      /**
+       * @param {any} formatter
+       * @param  {...any} args
```
EvanHahn, line 51: Could we use the stronger types from before?

EvanHahn (approved):

PR #890, opened 2024-10-03
chore: add pre-have logging to CoreSyncState

gmaclennan:
It's useful to know when pre-have messages are being received. This requires logging on `CoreSyncState`, which unfortunately involves passing the Logger instance down through various parent classes.

This PR also moves the code that was in `PeerSyncController` to add a prefix to each log message in to the `Logger` class, and re-uses this code in `CoreSyncState`.

It's useful to have a prefix for each log message, rather than use a different debug namespace, to keep logs more readable. The prefix is used to reference the peerId being synced with, or the core discovery key being handled. The debug namespace contains the deviceId as the postfix, so adding additional ids to the namespace would be confusing.

### DIFF
--- a/src/logger.js
+++ b/src/logger.js
@@ -82,13 +82,14 @@ export class Logger {
   /**
    * @param {string} ns
    * @param {Logger} [logger]
+   * @param {{ prefix?: string }} [opts]
    */
-  static create(ns, logger) {
-    if (logger) return logger.extend(ns)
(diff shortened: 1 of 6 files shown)


EvanHahn (approved):
LGTM other than the type error in `src/sync/core-sync-state.js`. Looks like `Logger` needs to be imported.

PR #891, opened 2024-10-03
fix: sync state with disconnected peers (take 2)

gmaclennan:
There are two scenarios where a disconnected peer results in sync never showing as complete.

**Scenario A**:
1. 3 or more peers are connected.
2. A peer disconnects (e.g. wifi off)
3. Remaining peers add data (including inviting another device)
4. Any peer that was previously connected to the peer that disconnected thinks that the disconnected peer still "wants" data, so sync never completes.

**Scenario B**:
1. 3 or more peers are connected
2. A peer adds some data that is not synced (e.g. adds an observation while sync is not enabled)
3. That peer disconnects
4. Remaining peers attempt to sync.
5. Sync never completes because peers are waiting for data from disconnected peer.

Note that these bugs affect `project.$sync.waitForSync()`, which I don't think the front-end is not currently using. The mechanism used for determining if sync is completed in the front-end _may_ still work.

### Fix

This PR fixes what appears to be an oversight for what happens when a peer disconnects (e.g. turns wifi off or force-closes app). We were removing the `PeerSyncController`, but we were not removing the remote state from `CoreSyncState`. I think I had some confusion about whether we should keep remote states for disconnected peers. For now, this PR "fixes" this to just completely remove remote states for disconnected peers. If we think keeping the remote state for disconnected peers in the future is useful, we can add it at a later stage. Right now, the remote state for a disconnected peer is just a cause of bugs, so not useful.

### DIFF
--- a/src/sync/core-sync-state.js
+++ b/src/sync/core-sync-state.js
@@ -155,7 +155,7 @@ export class CoreSyncState {
    * @param {Uint32Array} bitfield
    */
   insertPreHaves(peerId, start, bitfield) {
-    const peerState = this.#getPeerState(peerId)
+    const peerState = this.#getOrCreatePeerState(peerId)
     peerState.insertPreHaves(start, bitfield)
     const previousLength = Math.max(
       this.#preHavesLength,
@@ -185,7 +185,7 @@ export class CoreSyncState {
    * @param {Array<{ start: number, length: number }>} ranges
    */
   setPeerWants(peerId, ranges) {
-    const peerState = this.#getPeerState(peerId)
+    const peerState = this.#getOrCreatePeerState(peerId)
     for (const { start, length } of ranges) {
       peerState.setWantRange({ start, length })
     }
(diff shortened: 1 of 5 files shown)


src/sync/core-sync-state.js
```diff
@@ -204,7 +204,17 @@ export class CoreSyncState {
...
+    const wasRemoved = this.#remoteStates.delete(peerId)
+    if (wasRemoved) {
+      this.#update()
+    }
+  }
+
+  /**
+   * @param {PeerId} peerId
+   */
+  #getOrCreatePeerState(peerId) {
```
gmaclennan, line 217: The name of this method was misleading, and its use in #onPeerRemove was causing a bug, because it was recreating a removed state.

src/sync/sync-api.js
```diff
@@ -406,7 +406,11 @@ export class SyncApi extends TypedEmitter {
    *
    * @param {{ protomux: import('protomux')<import('@hyperswarm/secret-stream')>, remotePublicKey: Buffer }} peer
    */
-  #handlePeerRemove = (peer) => {
+  #handlePeerDisconnect = (peer) => {
+    this.#l.log(
+      '@@@@ handlePeerRemove %h',
```
EvanHahn, line 411: Nit: should we include this `@@@@`?

src/sync/sync-state.js
```diff
@@ -49,6 +49,15 @@ export class SyncState extends TypedEmitter {
     }
   }
 
+  /**
+   * @param {string} peerId
+   */
+  disconnectPeer(peerId) {
```
EvanHahn, line 55: Nit: should this be called `removePeer()` to match `addPeer()`? Same comment applies elsewhere.

EvanHahn (approved):

src/sync/sync-state.js
gmaclennan: I renamed it because there is a `removePeer` in CoreSyncState which means something different.

PR #893, opened 2024-10-07
feat: use req.hostname instead of manually configured server host.

gmaclennan:
This removes the server option `serverPublicBaseUrl` and instead uses the hostname from the `Host` header (or `X-Forwarded-Host` if [`trustProxy`](https://fastify.dev/docs/v4.28.x/Reference/Server/#factory-trust-proxy) is enabled).

The reason for this change is to simplify server configuration and avoiding errors stemming from the server value not matching the URL the server is accessible on.

The `Host` header is required by the HTTP 1.1 spec, so will always be present and can be trusted. We use the value that the client sends when creating a project, so the value used will be the value the client enters when they add the server.

### DIFF
--- a/src/server/routes.js
+++ b/src/server/routes.js
@@ -114,11 +114,7 @@ export default async function routes(
         403,
         'Server is already linked to a project'
       )
-
-      const baseUrl = new URL(
-        fastify.prefix,
-        `${req.protocol}://${req.hostname}`
-      ).toString()
+      const baseUrl = req.baseUrl.toString()
 
       const existingDeviceInfo = this.comapeo.getDeviceInfo()
       // We don't set device info until this point. We trust that `req.hostname`
@@ -182,10 +178,6 @@ export default async function routes(
     async function (req, reply) {
       const { projectPublicId } = req.params
       const project = await this.comapeo.getProject(projectPublicId)
-      const baseUrl = new URL(
-        fastify.prefix,
-        `${req.protocol}://${req.hostname}`
-      )
 
       reply.send({
         data: (await project.observation.getMany()).map((obs) => ({
@@ -198,7 +190,7 @@ export default async function routes(
             url: new URL(
               // TODO: Support other variants
               `projects/${projectPublicId}/attachments/${attachment.driveDiscoveryId}/${attachment.type}/${attachment.name}`,
-              baseUrl
+              req.baseUrl
             ),
           })),
         })),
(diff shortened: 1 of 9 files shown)


src/server/routes.js
```diff
@@ -48,9 +48,8 @@ export default async function routes(
     },
     async function (_req, reply) {
       const { deviceId, name } = this.comapeo.getDeviceInfo()
-      this.assert(name, 500, 'Server is missing a name')
       reply.send({
-        data: { deviceId, name },
+        data: { deviceId, name: name || serverName },
```
gmaclennan, line 52: name might not be configured in database at this point

test-e2e/server-integration.js
```diff
@@ -201,14 +201,16 @@ test('observations endpoint', async (t) => {
 
       await project.$sync.waitForSync('full')
 
-      const response = await server.inject({
-        method: 'GET',
-        url: `/projects/${projectId}/observations`,
-        headers: { Authorization: 'Bearer ' + BEARER_TOKEN },
-      })
+      const response = await fetch(
```
gmaclennan, line 204: Mixing the server actually listening on a port and using `fastify.inject()` is confusing because the server behaves in different ways - should really change everything in this test to use fetch rather than inject, but that can be a follow-up.

test-e2e/server.js
```diff
@@ -47,7 +47,7 @@ test('adding a server peer', async (t) => {
   )
   assert.equal(
     serverPeer.selfHostedServerDetails?.baseUrl,
-    'http://localhost:9876',
+    'http://localhost:9876/',
```
gmaclennan, line 50: `baseUrl` now always has a trailing slash.

test-e2e/server.js
```diff
@@ -142,12 +142,11 @@ async function createTestServer(t) {
...
     ...getManagerOptions('test server'),
     serverName: 'test server',
-    serverPublicBaseUrl: 'http://localhost:' + port,
     serverBearerToken: 'ignored',
   })
-  const serverAddress = await server.listen({ port })
+  await server.listen({ port })
   t.after(() => server.close())
-  return serverAddress
+  return `http://localhost:${port}`
```
gmaclennan, line 149: `server.listen()` returns _one_ of the addresses that fastify is listening on (on my Mac, the ipv6 address). We want to test access to the server based on `localhost` rather than the listening address.

PR #894, opened 2024-10-08
Add Dockerfile and fly.io deployment instructions and config

gmaclennan:
Depends on #893 

This adds a Dockerfile plus deployment / hosting instructions for fly.io

I also added an `allowedProjects` configuration option, because I think allowing multiple projects is easier than tearing down and setting up the server each time.

The fly.io config creates a single shared-cpu-1x vm with 256Mb of memory which will stop when idle, and start automatically when sync is started. The cost is $0.0027/hour, or $1.94/month if running permanently.

Data is stored on a fly.io volume that is initially 1Gb. Cost is $0.15/Gb/month. The volume size can be increased via `fly volumes extend -s 2gb`

The volume is snapshotted once per day and snapshots are retained for 14 days. It is possible that a volume can be corrupted and will need to be manually restored from a snapshot.

I also removed the ROOT_KEY option for the server, because tearing down a server and re-creating it with the same root key could cause fork errors with projects already syncing with the server. Better to just not allow it for now.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,16 @@
+# flyctl launch added from .gitignore
+**/.DS_Store
+**/node_modules
+**/coverage
+**/.tmp
+**/tmp
+**/proto/build
+dist
+!drizzle/**/*.sql
+**/.eslintcache
+**/docs/api/html/*
+**/test/fixtures/config/*.zip
+
+# flyctl launch added from .husky/_/.gitignore
+.husky/_/**/*
+fly.toml
--- a/fly.toml
+++ b/fly.toml
@@ -7,7 +7,6 @@ app = 'comapeo-cloud'
 primary_region = 'ord'
 
 [env]
-  SERVER_PUBLIC_BASE_URL = 'https://comapeo-cloud.fly.dev'
   STORAGE_DIR = '/data'
 
 [build]
--- a/fly.toml
+++ b/fly.toml
@@ -1,0 +1,37 @@
+# fly.toml app configuration file generated for comapeo-cloud on 2024-10-07T20:59:21+01:00
+#
+# See https://fly.io/docs/reference/configuration/ for information about how to use this file.
+#
+
+app = 'comapeo-cloud'
+primary_region = 'ord'
+
+[env]
+  SERVER_PUBLIC_BASE_URL = 'https://comapeo-cloud.fly.dev'
+  STORAGE_DIR = '/data'
+
+[build]
+
+[http_service]
+  internal_port = 8080
+  force_https = true
+  auto_stop_machines = 'stop'
+  auto_start_machines = true
+  min_machines_running = 0
+  max_machines_running = 1
+  processes = ['app']
+
+[[http_service.checks]]
+  grace_period = "10s"
+  interval = "30s"
+  method = "GET"
+  timeout = "5s"
+  path = "/healthcheck"
+
+[[vm]]
+  size = 'shared-cpu-1x'
+
+[mounts]
+  source = "myapp_data"
+  destination = "/data"
+  snapshot_retention = 14
--- a/package.json
+++ b/package.json
@@ -134,6 +134,7 @@
     "cpy-cli": "^5.0.0",
     "drizzle-kit": "^0.20.14",
     "eslint": "^8.57.0",
+    "execa": "^9.4.0",
     "husky": "^8.0.0",
     "iterpal": "^0.4.0",
     "lint-staged": "^14.0.1",
--- a/package.json
+++ b/package.json
@@ -158,7 +158,6 @@
     "@comapeo/schema": "file:comapeo-schema-server.tgz",
     "@digidem/types": "^2.3.0",
     "@electron/asar": "^3.2.8",
-    "@fastify/bearer-auth": "^10.0.1",
     "@fastify/error": "^3.4.1",
     "@fastify/sensible": "^5.6.0",
     "@fastify/static": "^7.0.3",
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "db:generate:project": "drizzle-kit generate:sqlite --schema src/schema/project.js --out drizzle/project",
     "db:generate:client": "drizzle-kit generate:sqlite --schema src/schema/client.js --out drizzle/client",
     "prepack": "npm run build:types",
-    "prepare": "husky install"
+    "prepare": "husky install || true"
   },
   "files": [
     "src",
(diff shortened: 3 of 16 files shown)


src/server/README.md
```diff
@@ -0,0 +1,41 @@
+## Deploying CoMapeo Cloud
+
+CoMapeo cloud comes with a [`Dockerfile`](../../Dockerfile) that can be used to build a Docker image. This image can be used to deploy CoMapeo cloud on a server.
```
EvanHahn, line 3: Nit:

```suggestion
CoMapeo Cloud comes with a [`Dockerfile`](../../Dockerfile) that can be used to build a Docker image. This image can be used to deploy CoMapeo Cloud on a server.
```

src/server/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+Server configuration is done using environment variables. The following environment variables are available:
+
+CoMapeo cloud comes with a [`Dockerfile`](../../Dockerfile) that can be used to build a Docker image. This image can be used to deploy CoMapeo cloud on a server.
+
+Server configuration is done using environment variables. The following environment variables are available:
+
```
EvanHahn, line 10: Nit: looks like these lines are duplicated.
```suggestion
```

src/server/server.js
```diff
@@ -27,10 +27,10 @@ const schema = Type.Object({
     description: 'path to directory where data is stored',
     default: DEFAULT_STORAGE,
   }),
-  ROOT_KEY: Type.Optional(
-    Type.String({
-      description:
-        'hex-encoded 16-byte random secret key, used for server keypairs',
+  ALLOWED_PROJECTS: Type.Optional(
```
EvanHahn, line 30: Nit: I think a name like `MAXIMUM_ALLOWED_PROJECTS` is clearer. I initially thought this was an array of project IDs or something, based on the name. (Probably also means updating the `allowedProjects` identifier elsewhere in the code.)

src/server/server.js
```diff
@@ -91,8 +87,10 @@ const fastify = createServer({
   logger: true,
 })
 
+fastify.get('/healthcheck', async () => {})
```
EvanHahn, line 90: Two small nits:

- could we add a basic test for this endpoint?
- does this handler need to be async?

fly.toml
```diff
@@ -0,0 +1,36 @@
+# fly.toml app configuration file generated for comapeo-cloud on 2024-10-07T20:59:21+01:00
+#
+# See https://fly.io/docs/reference/configuration/ for information about how to use this file.
+#
+
+app = 'comapeo-cloud'
+primary_region = 'ord'
```
EvanHahn, line 7: First of all, thank you for thinking of Chicago.

Second, given that we plan to demo this in Colombia, should we set this to `bog` (Bogotá) instead?

package.json
```diff
@@ -23,7 +23,7 @@
     "db:generate:project": "drizzle-kit generate:sqlite --schema src/schema/project.js --out drizzle/project",
     "db:generate:client": "drizzle-kit generate:sqlite --schema src/schema/client.js --out drizzle/client",
     "prepack": "npm run build:types",
-    "prepare": "husky install"
+    "prepare": "husky install || true"
```
EvanHahn, line 26: Is this necessary because we aren't installing Husky in prod?

src/server/README.md
```diff
@@ -0,0 +1,41 @@
...
+| --------------------- | -------- | -------------------------------------------------------------------- | ---------------- |
+| `SERVER_BEARER_TOKEN` | Yes      | Token for authenticating API requests. Should be large random string |                  |
+| `PORT`                | No       | Port on which the server runs                                        | `8080`           |
+| `SERVER_NAME`         | No       | Friendly server name, seen by users when adding server               | `CoMapeo Server` |
+| `ALLOWED_PROJECTS`    | No       | Number of projects allowed to register with the server               | `1`              |
+| `STORAGE_DIR`         | No       | Path for storing app & project data                                  | `$CWD/data`      |
+
+### Deploying with fly.io
+
+CoMapeo cloud can be deployed on [fly.io](https://fly.io) using the following steps:
```
EvanHahn, line 21: Nit:

```suggestion
CoMapeo Cloud can be deployed on [fly.io](https://fly.io) using the following steps:
```

src/server/README.md
```diff
@@ -0,0 +1,41 @@
...
+   ```
+4. Deploy the app by running (optionally setting the `ALLOWED_PROJECTS` environment variable):
+   ```sh
+   flyctl deploy --app <your-app-name> -e ALLOWED_PROJECTS=10
+   ```
+5. The app should now be running on fly.io. You can access it at `https://<your-app-name>.fly.dev`.
+
+To destroy the app (delete all data and project invites), run:
+
+> **Warning:** This action is irreversible and will permanently delete all data associated with the app, and projects that have already added the server will no longer be able to sync with it.
```
EvanHahn, line 37: Nit: we can use GitHub's fancy callouts feature:

```suggestion
> [!WARNING]
> This action is irreversible and will permanently delete all data associated with the app, and projects that have already added the server will no longer be able to sync with it.
```

It will render like this:

> [!WARNING]
> This action is irreversible and will permanently delete all data associated with the app ...

src/server/routes.js
```diff
@@ -14,14 +15,19 @@ const BASE32_STRING_32_BYTES = Type.String({ pattern: BASE32_REGEX_32_BYTES })
...
 /** @type {FastifyPluginAsync<RouteOptions, RawServerDefault, TypeBoxTypeProvider>} */
 export default async function routes(
   fastify,
-  { serverBearerToken, serverPublicBaseUrl }
+  { serverBearerToken, serverName, allowedProjects = 1 }
 ) {
+  /** @type {Set<string> | number} */
+  const allowedProjectsSetOrNumber = Array.isArray(allowedProjects)
+    ? new Set(allowedProjects)
+    : allowedProjects
```
EvanHahn, line 30: Setting this to an array is undocumented. Could we just support the number?

.dockerignore
EvanHahn, line 1: I think this list is fine for now, but I think it could be refined. Is there anything we can do to simplify this?

EvanHahn (approved):
LGTM other than a bunch of nitpicks.

src/server/server.js
gmaclennan: Yes, I guess, but.... adding the test infrastructure for this CLI is a lot of work, and when we actually test the fly deployment, this is implicitly tested, because Fly does not consider the app deployed until this returns 200.

In terms of handler being async, if it's not then it needs to be `(req, reply) => reply.send()`. Writing async is shorthand to avoid that (it replies with the return value of the function)

.dockerignore
gmaclennan: This was created by the fly cli init function, so just left it as fly recommends it.

fly.toml
gmaclennan: changed to iad as default, we can set to bog when deploying for a specific use

EvanHahn (approved):

PR #896, opened 2024-10-08
feat!: simplify fastify maps plugin and add support for styled map packages

achou11:
Towards #827 

Paired on much of this with @gmaclennan 

Leaving as a draft because tests will fail, as the changes rely on publishing a new version of `styled-map-package`.

Important notes:

- removes the `StaticMapPlugin` in favor of using the StyledMapPackage plugin to handle serving a local custom map
- removes the `OfflineFallbackMapPlugin` in favor of using the StyledMapPackage plugin to handle serving the SMP provided by `@comapeo/fallback-smp`
- simplifies implementation of the `MapeoMapsPlugin` due to the removal of the the other plugins.
  - one consequence of this is handling the registration of the plugin within the MapeoManager instance, as opposed to doing it externally. due to this change, a few options are added to the `MapeoManager` constructor, all of which are optional:
    - `customMapPath`: file path to a local SMP file used for the local custom map
    - `fallbackMapPath`: file path to a local SMP file used for the offline fallback map
    - `defaultOnlineStyleUrl`: url to online-hosted StyleJSON


gmaclennan (approved):
Looks good to me, let's merge once tests are passing.

I do agree with the original intention of keeping the plugin outside of MapeoManager, for the sake of separation of concerns, however in practice that would mean more of a tangle to expose `getMapStyleJsonUrl()` to the client. We're going to swap this out anyway for MapManager in the future, so I think it's more practical to take the easy route (e.g. code in this PR) for now, and deal with separation of concerns once we have the map manager.

PR #898, opened 2024-10-09
feat: add endpoint for maps plugin to get info about custom map

### LINKED ISSUE #827: Add support for serving Styled Map Package maps

achou11:
Closes #827 

Stacked on #896 

- adds a GET endpoint for the maps plugin located at `<prefix>/custom/info` to provide the name, size, and creation time of the custom map file if it exists

### DIFF
--- a/src/fastify-plugins/maps.js
+++ b/src/fastify-plugins/maps.js
@@ -4,7 +4,7 @@ import { fetch } from 'undici'
 import { Server as SMPServerPlugin } from 'styled-map-package'
 
 import { noop } from '../utils.js'
-import { NotFoundError } from './utils.js'
+import { NotFoundError, ENOENTError } from './utils.js'
 
 /** @import { FastifyPluginAsync } from 'fastify' */
 /** @import { Stats } from 'node:fs' */
@@ -53,7 +53,7 @@ export async function plugin(fastify, opts) {
         stats = await fs.stat(customMapPath)
       } catch (err) {
         if (err instanceof Error && 'code' in err && err.code === 'ENOENT') {
-          throw new NotFoundError(customMapPath)
+          throw new ENOENTError(customMapPath)
         }
 
         throw err
--- a/src/fastify-plugins/maps.js
+++ b/src/fastify-plugins/maps.js
@@ -45,7 +45,7 @@ export async function plugin(fastify, opts) {
           : undefined
 
       return {
-        created: stats.birthtime,
+        created: stats.ctime,
         size: stats.size,
         name: styleJsonName || path.parse(customMapPath).name,
       }
--- a/src/fastify-plugins/maps.js
+++ b/src/fastify-plugins/maps.js
@@ -45,7 +45,7 @@ export async function plugin(fastify, opts) {
           : undefined
 
       return {
-        created: stats.ctime,
+        created: stats.birthtime,
         size: stats.size,
         name: styleJsonName || path.parse(customMapPath).name,
       }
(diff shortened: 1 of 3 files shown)


src/fastify-plugins/maps.js
```diff
@@ -19,9 +21,39 @@ export const FALLBACK_MAP_PREFIX = 'fallback'
...
+    fastify.get(`/${CUSTOM_MAP_PREFIX}/info`, async () => {
+      const baseUrl = new URL(fastify.prefix, fastify.listeningOrigin)
+
+      if (!baseUrl.href.endsWith('/')) {
+        baseUrl.href += '/'
+      }
+
+      const style = await (
+        await fetch(new URL(`${CUSTOM_MAP_PREFIX}/style.json`, baseUrl))
+      ).json()
```
gmaclennan, line 35: maybe check the response first and pass-through a 404 error.

src/fastify-plugins/maps.js
```diff
@@ -19,9 +21,39 @@ export const FALLBACK_MAP_PREFIX = 'fallback'
...
+
+      const stats = await fs.stat(customMapPath)
```
gmaclennan, line 37: should catch an error and if EOENT then 404 maybe? Shouldn't happen because the route above would 404 before here, but for completeness...

src/fastify-plugins/maps.js
```diff
@@ -19,9 +21,39 @@ export const FALLBACK_MAP_PREFIX = 'fallback'
...
+
+      const styleJsonName =
+        typeof style === 'object' &&
+        style &&
+        'name' in style &&
```
gmaclennan, line 42: I have this kind of check in so many places in my code and I blame typescript, I can't see any way that it's actually necessary (not saying you shouldn't have it here, just venting!)

test/fastify-plugins/maps.js
```diff
@@ -223,6 +225,83 @@ test('/style.json resolves style.json of fallback map when custom and online are
...
+  const address = await server.listen()
+
+  mockAgent.enableNetConnect(new URL(address).host)
+
+  const response = await server.inject({
+    method: 'GET',
+    url: `/${CUSTOM_MAP_PREFIX}/info`,
+  })
+
+  assert.equal(response.statusCode, 500, 'returns server error status code')
```
gmaclennan, line 269: helpful maybe to distinguish "custom map file does not exist" (404) from "custom map file is invalid" (500), since at some point we will probably want different UX for each type, and before then we might want to report actual 500 errors to Sentry, but not 404.

gmaclennan (approved):
Looks good. Couple of minor niggles, but good to merge even without addressing those.

test/fastify-plugins/maps.js
```diff
@@ -223,6 +224,79 @@ test('/style.json resolves style.json of fallback map when custom and online are
...
+
+  assert.equal(response.statusCode, 200)
+
+  const info = response.json()
+
+  assert.deepEqual(info, {
+    created: '2024-10-09T17:41:16.103Z',
+    size: 2815018,
+    name: 'MapLibre',
+  })
```
achou11, line 297: any thoughts on the best approach for this assertion? seems like `created` field is different each time github actions run so it always seems to fail (maybe that's expected??)

gmaclennan: Stat the file in the test that you are passing to the server, and check the value from the server is the same as what you get from the file.

gmaclennan (approved):

PR #900, opened 2024-10-09
chore: improve types for `MapeoProject`'s `kProjectReplicate`

EvanHahn:
*This change should have no user impact.*

*I recommend [reviewing this with whitespace changes disabled](https://github.com/digidem/comapeo-core/pull/900/files?w=1).*

`MapeoProject`'s `kProjectReplicate` member will give type errors if called with a boolean. We aren't doing this today, but [we will in our upcoming server work][0].

This happened because TypeScript's `Parameters` utility type only takes the *last* signature of an overloaded function. To quote [the docs][1]:

> When inferring from a type with multiple call signatures (such as the type of an overloaded function), inferences are made from the last signature (which, presumably, is the most permissive catch-all case). It is not possible to perform overload resolution based on a list of argument types.

To fix this, we specify the types manually rather than pull them out of `Parameters`. Doing this fixes the potential type error.

I also did a small amount of cleanup on the types in this function.

(This change is primarily to fix an error in our upcoming server branch, but I think it's useful on its own.)

[0]: https://github.com/digidem/comapeo-core/blob/9aaeb37910821a6779ec8eef88af995e94448ba3/src/mapeo-project.js#L140-L141
[1]: https://www.typescriptlang.org/docs/handbook/2/conditional-types.html


gmaclennan (approved):
Looks good. Bear in mind that this now excludes from types passing a Protomux instance, which is supported, however the [return type is different](https://github.com/digidem/digidem-types/blob/86b8e810bd734c178be48c70766b7cb578eda2d2/vendor/hypercore/index.d.ts#L250-L254) when passing a protomux. This is an [upstream bug](https://github.com/holepunchto/hypercore/issues/448), but I think our types are also incorrect (types say return type is Protomux, but I _think_ the return type is NoiseStream).

PR #901, opened 2024-10-10
chore: allow any JSON object to be saved in blob store

EvanHahn:
The blob store will need to store additional metadata, such as audio duration, in the future. This changes its `metadata` type to any JSON object to allow this.

It also adds a test for `blobStore.entry()` to ensure metadata is saved.

This change shouldn't affect the external API because the blob store is not directly exposed. It's used internally by `BlobApi`, which did not change.


gmaclennan (changes requested):
A couple of changes which would be good to include:

- [x] Also change type for `put()` method
- [ ] Override the return of `entry()` to make `Entry.metadata: JsonObject` rather than the current looser `object`.

EvanHahn:
Made `put` use JSON in efdce7776ec0e2b31ee47b53c6e99c4349eab704, and improved the return type in 58dab5b0d6b3ec138ba99c57fc24be62e25c9936.

gmaclennan (approved):

PR #906, opened 2024-10-17
fix: support maps that may be removed or replaced in map server plugin

achou11:
Updates `styled-map-package` to v2.0.0, which adds support for custom maps that may be removed or replaced during the lifetime of the server instance. This addresses a need for the consuming apps, as we have the ability to delete the custom map file at any point.

### DIFF
--- a/src/fastify-plugins/maps.js
+++ b/src/fastify-plugins/maps.js
@@ -1,7 +1,7 @@
 import fs from 'node:fs/promises'
 import path from 'node:path'
 import { fetch } from 'undici'
-import { Server as SMPServerPlugin } from 'styled-map-package'
+import { ReaderWatch, Server as SMPServerPlugin } from 'styled-map-package'
 
 import { noop } from '../utils.js'
 import { NotFoundError, ENOENTError } from './utils.js'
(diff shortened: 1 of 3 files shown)


src/fastify-plugins/maps.js
```diff
@@ -78,13 +82,17 @@ export async function plugin(fastify, opts) {
...
   }
 
+  const fallbackMapReader = new ReaderWatch(opts.fallbackMapPath)
+
+  fastify.addHook('onClose', () => fallbackMapReader.close().catch(noop))
+
   fastify.register(SMPServerPlugin, {
     prefix: FALLBACK_MAP_PREFIX,
-    filepath: opts.fallbackMapPath,
+    reader: fallbackMapReader,
```
achou11, line 95: CoMapeo Mobile doesn't really need this to be set up the same way, but figured it would make more sense to do this anyways for the sake of having the flexibility.

EvanHahn: I like this because I don't want Core to have to rely on Mobile doing things a specific way.

EvanHahn (approved):
I haven't really been keeping up with the `styled-map-package` work lately, but LGTM.

PR #909, opened 2024-10-17
chore: add `omit` utility

EvanHahn:
*This PR is not urgent.*

This change should have no impact on functionality.

This replaces the following pattern:

```javascript
// eslint-disable-next-line no-unused-vars
const { foo, bar, ...rest } = obj
return rest
```

With a new `omit` utility, which I think is clearer:

```javascript
return omit(obj, ['foo', 'bar'])
```

I think this is a useful change on its own, but also makes an upcoming change slightly easier because I wanted this function.

### DIFF
--- a/src/config-import.js
+++ b/src/config-import.js
@@ -516,7 +516,6 @@ function parseIcon(filename, buf) {
   if (!matches) {
     throw new Error(`Unexpected icon filename ${filename}`)
   }
-  /* eslint-disable no-unused-vars */
   const [_, name, size, pixelDensityStr] = matches
   const pixelDensity = Number(pixelDensityStr)
   if (!(pixelDensity === 1 || pixelDensity === 2 || pixelDensity === 3)) {
--- a/src/mapeo-manager.js
+++ b/src/mapeo-manager.js
@@ -34,6 +34,7 @@ import {
   projectKeyToPublicId,
 } from './utils.js'
 import { openedNoiseSecretStream } from './lib/noise-secret-stream-helpers.js'
+import { omit } from './lib/omit.js'
 import { RandomAccessFilePool } from './core-manager/random-access-file-pool.js'
 import BlobServerPlugin from './fastify-plugins/blobs.js'
 import IconServerPlugin from './fastify-plugins/icons.js'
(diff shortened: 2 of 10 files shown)


src/config-import.js
```diff
@@ -516,7 +516,6 @@ function parseIcon(filename, buf) {
   if (!matches) {
     throw new Error(`Unexpected icon filename ${filename}`)
   }
-  /* eslint-disable no-unused-vars */
```
EvanHahn, line 519: Not strictly related to this change. (Worth doing, though.)

src/mapeo-manager.js
```diff
@@ -442,9 +443,10 @@ export class MapeoManager extends TypedEmitter {
 
     // 7. Load config, if relevant
     // TODO: see how to expose warnings to frontend
-    /* eslint-disable no-unused-vars */
```
EvanHahn, line 445: Changed this to two `next-line` disables. The old code disabled the `no-unused-vars` rule _for the rest of the file_, which was probably not the intent.

Not strictly related to this change, though.

src/mapeo-manager.js
```diff
@@ -917,15 +919,8 @@ export class MapeoManager extends TypedEmitter {
...
-    ({
-      // @ts-ignore
-      // eslint-disable-next-line no-unused-vars
-      protomux,
-      ...publicPeerInfo
-    }) => {
-      return publicPeerInfo
-    }
+  return peers.map((peer) =>
+    'protomux' in peer ? omit(peer, ['protomux']) : peer
```
EvanHahn, line 923: The functionality is _ever so slightly_ different here. If the `protomux` property is missing, it will now return the original object instead of a copy. I don't think this causes any actual differences.

test/translation-api.js
```diff
@@ -69,11 +69,8 @@ test('translation api - put() and get()', async () => {
     `we've effectively have fields in spanish`
   )
 
-  /* eslint-disable no-unused-vars */
-  const { schemaName, message: msg, ...docToGet } = doc
```
EvanHahn, line 73: Turns out we didn't need to do this at all, so we can just remove these lines.

achou11 (approved):
lgtm! seeing this reminds me of https://github.com/toss/es-toolkit, which maybe isn't necessary for us anytime soon, but could be worth considering if we keep adding things like this.

PR #932, opened 2024-10-22
feat: Add setIsArchiveDevice() method

### LINKED ISSUE #680: Add a device setting to configure what to sync
We need to persist a users's choice about what they want to sync (current options are "previews only" and "everything"). This is a device-based setting, which might be scoped to project (e.g. support different settings for each project). It shouldn't be written to the project settings of the database, which syncs with other devices.

I think this makes sense to implement on the frontend, and add an API method to change this setting.

Choices to make:

1. Default sync config for a new project instance
    a. All new project instances start with "previews only" until configured otherwise.
    b. All new project instances start with "everything" until configured otherwise.
    c. We add a new option to `api.getProject` for configuring what to sync
    d. Combination of above (option + default)
2. Method name & signature
    a. `project.configureSync('previews' | 'everything')`
    b. `project.setArchiver(true | false)`
    c. `project.setBlobDownloadFilter({ variants: Array<'preview' | 'thumbnail' | 'original'> })`
    d. Other ideas welcome!

gmaclennan:
Does not change sync at this time, but does persist setting.

Fixes #680


gmaclennan:
Currently this throws if trying to call `manager.setIsArchiveDevice()` before first calling `setDeviceInfo()`. Not sure what the alternative is?

src/mapeo-project.js
```diff
@@ -93,6 +95,7 @@ export class MapeoProject extends TypedEmitter {
   #loadingConfig
 
   static EMPTY_PROJECT_SETTINGS = EMPTY_PROJECT_SETTINGS
+  #isArchiveDevice
```
EvanHahn, line 98: Nit: could you move this a few lines up, with the rest of the members?

test-e2e/manager-archive-device.js
```diff
@@ -0,0 +1,140 @@
...
+    async (manager2) => {
+      return manager2.getIsArchiveDevice()
+    }
+  )
+
+  assert.equal(isArchiveDevice1, false)
+  assert.equal(isArchiveDevice2, isArchiveDevice1)
+})
+
+test('migration of localDeviceInfo table', async (t) => {
```
EvanHahn, line 84: I love this. Nit: I think it makes more sense to test this inside a separate migrations file.

test-e2e/manager-archive-device.js
```diff
@@ -0,0 +1,140 @@
...
+
+test('Set & Get isArchiveDevice', async () => {
+  const manager = new MapeoManager({
+    rootKey: KeyManager.generateRootKey(),
+    projectMigrationsFolder,
+    clientMigrationsFolder,
+    dbFolder: ':memory:',
+    coreStorage: () => new RAM(),
+    fastify: Fastify(),
+  })
```
EvanHahn, line 27: Why not use the existing test helpers for creating a manager?

gmaclennan:
I decided to remove the requirement that `setDeviceInfo()` must be called before calling `setIsArchiveDevice()` by migrating the `deviceInfo` column to be nullable, which required manually writing the migration because SQLite does not support `ALTER COLUMN` statements. I added a test for the migration, which checks that any device info set before migration is still readable after migration.

test-e2e/manager-archive-device.js
```diff
@@ -0,0 +1,140 @@
...
+    project[kIsArchiveDevice],
+    true,
+    'Project isArchiveDevice is true initially'
+  )
+
+  assert.strictEqual(
+    manager.getIsArchiveDevice(),
+    true,
+    'isArchiveDevice is true initially'
+  )
```
EvanHahn, line 42: Nit: could you move this above the project creation to ensure that it's set immediately?

test-e2e/manager-archive-device.js
```diff
@@ -0,0 +1,140 @@
...
+  const projectMigrationsFolderPreMigration = fileURLToPath(
+    new URL('../drizzle/project', comapeoCorePreMigrationUrl)
+  )
+
+  const dbFolder = temporaryDirectory()
+  const rootKey = KeyManager.generateRootKey()
+  t.after(() => {
+    fs.rmSync(dbFolder, { recursive: true })
+  })
```
EvanHahn, line 100: Two nits:

1. could we remove this asynchronously (also requires updating the `fs` import to use `node:fs/promises`)?

```suggestion
  t.after(() => fs.rm(dbFolder, { recursive: true }))
```

2. Could we move this `t.after` call to right after we create `dbFolder`?

src/mapeo-project.js
```diff
@@ -625,6 +631,15 @@ export class MapeoProject extends TypedEmitter {
     return deviceInfo.update(existingDoc.versionId, doc)
   }
 
+  /** @param {boolean} isArchiveDevice */
+  async [kSetIsArchiveDevice](isArchiveDevice) {
+    this.#isArchiveDevice = isArchiveDevice
+  }
+
+  get [kIsArchiveDevice]() {
```
EvanHahn, line 639: Nit: I know the return type is trivial, but I'd love if we could add add a return type:

```suggestion
  /** @returns {boolean} */
  get [kIsArchiveDevice]() {
```

src/mapeo-project.js
```diff
@@ -108,6 +111,7 @@ export class MapeoProject extends TypedEmitter {
    * @param {(mediaType: 'blobs' | 'icons') => Promise<string>} opts.getMediaBaseUrl
    * @param {import('./local-peers.js').LocalPeers} opts.localPeers
    * @param {Logger} [opts.logger]
+   * @param {boolean} [opts.isArchiveDevice] Whether this device is an archive device
```
EvanHahn, line 114: Is there ever a time we'd instantiate a project without this option? I think not, so maybe we can make this required?

EvanHahn (changes requested):
LGTM other than a few nitpicks and small questions.

src/mapeo-project.js
gmaclennan: Addressed in 2dc69a6a0351cb1db0ff65b813e1b0efdccaccae

test-e2e/manager-archive-device.js
gmaclennan: Addressed in 2dc69a6a0351cb1db0ff65b813e1b0efdccaccae

EvanHahn (approved):

test-e2e/manager-archive-device.js
gmaclennan: Addressed in 2dc69a6a0351cb1db0ff65b813e1b0efdccaccae

src/mapeo-project.js
gmaclennan: Addressed in 2dc69a6a0351cb1db0ff65b813e1b0efdccaccae

PR #945, opened 2024-10-29
test: basic test that uses IPC

EvanHahn:
CoMapeo Mobile doesn't use `MapeoManager` directly; it uses an IPC bridge.

This adds a test that creates a manager in a separate process and uses `@comapeo/ipc` to test it.

This is something we've been discussing for awhile, and this finally implements it. I think this is useful on its own but also makes a future change easier.


EvanHahn:
Renamed in cc6b1b714790c24cc2bc555af43145749880407f.

Issue #436, opened 2024-01-18
Failure to decode a block (fully or partially) will never be corrected


EvanHahn:
### The problem

If a device fails to fully decode a block, the missing data will be forever missing from the SQLite index, and therefore the app. I think this can happen if there's a *total* or *partial* failure to decode a block.

#### Total failure to decode

The simplest example: when `IndexWriter` fails to decode a block, it discards it:

https://github.com/digidem/mapeo-core-next/blob/75a312d01d0d74ce64640816407a2c5d0a371a2f/src/index-writer/index.js#L71-L78

Imagine that my device is on an older version than yours, and that your device has a data type I've never heard of. `decode()` will throw and the block will be "silently ignore"d.

We will never return to this block. This block will sync to my Hypercore but will be permanently missing from my SQLite index, even if I update.

#### Partial failure to decode

I believe a similar issue can happen when a data type gets a new field. A device will only add fields it understands:

```js
// Snippet from https://github.com/digidem/mapeo-sqlite-indexer/blob/4e9484442204157a792d808da333768f49774c4d/index.js#L74-L77
this.#writeDocSql = db.prepare(
  `REPLACE INTO ${docTableName} (${docColumns.join(',')})
  VALUES (${docColumns.map((name) => `@${name}`).join(',')})`
)
```

If a new field is added, we'll never go back to insert it.

### Possible solutions

I can think of a few solutions:

1. When we detect a version change (e.g., a new field), completely rebuild the index from scratch.
   - Pros: should be simple to drop everything and reset.
   - Cons: will this be slow? Will it open a long-running transaction that blocks other database operations? Is it easy to detect these changes reliably? Will it require a "migrating" screen to be added to CoMapeo? Should there be a way to manually rebuild the index in some advanced settings in case we make a mistake?
1. Store non-decoded blocks and fields for later processing. At some point (e.g., startup of a new app version), reprocess them.
   - Pros: does minimal work, which should be good for performance.
   - Cons: may require [a patch of protobufjs to handle unknown fields][0] depending on our approach. Probably more complex to implement.
1. Some other option I'm not thinking of.
   - Pros: ???
   - Cons: ???

I think this is worth discussing, but I think option 1 is probably best in the short term. It seems fairly easy to implement, and we can always implement option 2 if performance becomes a concern. But I may be wrong!

## Tasks

Assuming we take my preferred approach:

- [ ] Add tests
  - [ ] Test a block that completely fails to decode (e.g., new data type)
[...]

PR #949, opened 2024-10-30
feat: rebuild SQLite when migrations occur

EvanHahn:
See #436.

### DIFF
--- a/src/lib/drizzle-helpers.js
+++ b/src/lib/drizzle-helpers.js
@@ -32,7 +32,7 @@ const tableCountIfExists = (db, tableName) =>
     const existsQuery = sql`
       SELECT EXISTS (
         SELECT 1
-        FROM sqlite_master
+        FROM sqlite_schema
         WHERE type IS 'table'
         AND name IS ${tableName}
       ) AS result
--- a/src/lib/drizzle-helpers.js
+++ b/src/lib/drizzle-helpers.js
@@ -1,0 +1,48 @@
+import { sql } from 'drizzle-orm'
+import { assert } from '../utils.js'
+/** @import { BetterSQLite3Database } from 'drizzle-orm/better-sqlite3' */
+
+/**
+ * @param {unknown} queryResult
+ * @returns {number}
+ */
+const getNumberResult = (queryResult) => {
+  assert(
+    queryResult &&
+      typeof queryResult === 'object' &&
+      'result' in queryResult &&
+      typeof queryResult.result === 'number',
+    'expected query to return proper result'
+  )
+  return queryResult.result
+}
+
+/**
+ * Get the number of rows in a table using `SELECT COUNT(*)`.
+ * Returns 0 if the table doesn't exist.
+ *
+ * @param {BetterSQLite3Database} db
+ * @param {string} tableName
+ * @returns {number}
+ */
+export const tableCountIfExists = (db, tableName) =>
+  db.transaction((tx) => {
+    const existsQuery = sql`
+      SELECT EXISTS (
+        SELECT 1
+        FROM sqlite_master
+        WHERE type IS 'table'
+        AND name IS ${tableName}
+      ) AS result
+    `
+    const existsResult = tx.get(existsQuery)
+    const exists = getNumberResult(existsResult)
+    if (!exists) return 0
+
+    const countQuery = sql`
+      SELECT COUNT(*) AS result
+      FROM ${sql.identifier(tableName)}
+    `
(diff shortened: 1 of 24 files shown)


multi-core-indexer-1.0.0-alpha.10.tgz
EvanHahn, line 1: I plan to remove this once https://github.com/digidem/multi-core-indexer/pull/45 is merged.

src/mapeo-project.js
```diff
@@ -139,11 +144,37 @@ export class MapeoProject extends TypedEmitter {
...
     this.#sqlite = new Database(dbPath)
     const db = drizzle(this.#sqlite)
-    migrate(db, { migrationsFolder: projectMigrationsFolder })
+    const migrationsBefore = tableCountIfExists(db, DRIZZLE_MIGRATIONS_TABLE)
+    migrate(db, {
+      migrationsFolder: projectMigrationsFolder,
+      migrationsTable: DRIZZLE_MIGRATIONS_TABLE,
+    })
+    const migrationsAfter = tableCountIfExists(db, DRIZZLE_MIGRATIONS_TABLE)
+    const reindex = migrationsBefore > 0 && migrationsAfter !== migrationsBefore
```
gmaclennan, line 156: Maybe wrap this in our own migrate function for clarity? e.g. 
```
const didMigrate = migrate(db, { migrationsFolder: projectMigrationsFolder })
```

gmaclennan (changes requested):
Ok looks good apart from one code clarity suggestion. However I think we should also include in this re-indexing of the `projectSettingsTable` (in `MapeoManager`) which I think needs the same treatment?

src/lib/drizzle-helpers.js
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * Get the number of rows in a table using `SELECT COUNT(*)`.
+ * Returns 0 if the table doesn't exist.
+ *
+ * @param {BetterSQLite3Database} db
+ * @param {string} tableName
+ * @returns {number}
+ */
+export const tableCountIfExists = (db, tableName) =>
```
gmaclennan, line 28:
```suggestion
export const countTableRows = (db, tableName) =>
```
Naming of this one is hard! Not sure if this is better. The original name suggested to me that it is counting tables ("tableCount"), and only counting if something exists ("IfExists"), where-as I think what this is doing is counting rows in a table whether it exists or not.

gmaclennan: `safeCountTableRows()`?

src/lib/drizzle-helpers.js
```diff
@@ -0,0 +1,48 @@
...
+  db.transaction((tx) => {
+    const existsQuery = sql`
+      SELECT EXISTS (
+        SELECT 1
+        FROM sqlite_master
```
gmaclennan, line 33: Maybe use `sqlite_schema`? https://sqlite.org/schematab.html

gmaclennan: When you do `PRAGMA table_list` you see `sqlite_schema` as the only table.

EvanHahn: `sqlite_master` is an "historical" alias for `sqlite_schema`. I should be using `_schema`. Fixed in 574dca554b057d4f822c4279f5c47c6e0fa208d2.

src/lib/drizzle-helpers.js
EvanHahn: Renamed it in f93b4a4dfbc45021e937003583c5d96c6c7edc0e. I also stopped exporting it in ae6e0c6d4d97b160f5336b79a578f80bc1aa7085.

src/mapeo-project.js
EvanHahn: Done in ae6e0c6d4d97b160f5336b79a578f80bc1aa7085.

EvanHahn:
> Ok looks good apart from one code clarity suggestion. However I think we should also include in this re-indexing of the `projectSettingsTable` (in `MapeoManager`) which I think needs the same treatment?

Done in 5084bb2e7d853861ca4fb39807c4d9d21761fb72. (I didn't have an easy way to test this so I didn't.)

> Final thoughts, one way of doing this in the future would be to compare just the indexed tables in the `sqlite_schema` table, comparing the `sql` column, to detect changes. That way we would avoid re-indexing if we change the schema of a table that is not coming from our indexer.

That's a much better idea. I'll consider doing that in a followup.

gmaclennan (approved):

PR #950, opened 2024-10-30
fix: memory leak in `DataStore`

EvanHahn:
We have a set of promises waiting to be indexed, and we never removed these promises from the set. This fixes that memory leak.


EvanHahn:
Moved in af184f27f8eb186f0053b1ca6d6cbe593cd9c0fa. I'll land if CI passes.

PR #951, opened 2024-10-30
chore: remove manual type definitions for DataType

EvanHahn:
*This change should have no user impact.*

Before this change, DataType had its own type definitions in `src/datatype/index.d.ts`. This removes that file.

I tested this by manually including it in CoMapeo Mobile and it type-checked fine.


src/datatype/index.js
```diff
@@ -166,28 +200,29 @@ export class DataType extends TypedEmitter {
...
    */
   async getByVersionId(versionId, { lang } = {}) {
     const result = await this.#dataStore.read(versionId)
-    return this.#translate(result, { lang })
+    return this.#translate(deNullify(result), { lang })
   }
 
   /**
-   * @param {MapeoDoc} doc
+   * @param {any} doc
```
EvanHahn, line 223: I was unable to figure out how to get this to type-check without `any`, much to my dismay.

gmaclennan (approved):

Issue #905, opened 2024-10-17
Convert "want hints" into want bitfields


gmaclennan:
A "want hint" is a list of filename filters for the blob hyperdrive(s), that is sent between peers.

A peer receiving a "want hint" should use the hyperdrive blobIndex to convert that want hint into a bitfield for the hyperdrive blob core.

This bitfield(s) can then be passed to the sync API.

EvanHahn:
Done in 3d1c94b36314a32995766e9e43b4ea61504b1026.

PR #956, opened 2024-11-07
feat: handle incoming blob filters

### LINKED ISSUE #682: Support setting "wants" on sync state, rather than the default of "want everything"
Sync state should accept bitfields to define "wants". Once a "want" is received for a given core, sync state should no longer assume that the peer wants everything in the core, instead it only wants what is defined in the want bitfield(s).

EvanHahn:
*I recommend [reviewing this PR one commit at a time](https://github.com/digidem/comapeo-core/pull/956/commits).*

When you receive a blob filter from another peer, this updates their sync states. For example, if you receive a blob filter that says "I only want photo thumbnails", that peer's "wants" bitfield will be reduced.

Closes #682 and #905.

### DIFF
--- a/src/blob-store/downloader.js
+++ b/src/blob-store/downloader.js
@@ -2,8 +2,8 @@ import { TypedEmitter } from 'tiny-typed-emitter'
 import { createEntriesStream } from './entries-stream.js'
 import { filePathMatchesFilter } from './utils.js'
 
-/** @import Hyperdrive from 'hyperdrive' */
 /** @import { BlobFilter } from '../types.js' */
+/** @import { THyperdriveIndex } from './hyperdrive-index.js' */
 
 /**
  * Like hyperdrive.download() but 'live', and for multiple drives.
@@ -26,7 +26,7 @@ import { filePathMatchesFilter } from './utils.js'
  * @extends {TypedEmitter<{ error: (error: Error) => void }>}
  */
 export class Downloader extends TypedEmitter {
-  /** @type {import('./index.js').THyperdriveIndex} */
+  /** @type {THyperdriveIndex} */
   #driveIndex
   /** @type {Set<{ done(): Promise<void>, destroy(): void }>} */
   #queuedDownloads = new Set()
@@ -36,7 +36,7 @@ export class Downloader extends TypedEmitter {
   #shouldDownloadFile
 
   /**
-   * @param {import('./index.js').THyperdriveIndex} driveIndex
+   * @param {THyperdriveIndex} driveIndex
    * @param {object} [options]
    * @param {BlobFilter | null} [options.filter] Filter blobs of specific types and/or sizes to download
    */
--- a/src/blob-store/index.js
+++ b/src/blob-store/index.js
@@ -137,13 +137,7 @@ export class BlobStore extends TypedEmitter {
     const entriesStream = createEntriesStream(this.#driveIndex, { live })
     if (!filter) return entriesStream
     const filterStream = new FilterEntriesStream(filter)
-    const result = pipeline(entriesStream, filterStream, noop)
-    // Destroying the pipeline should destroy the *first* stream, not the
-    // whole pipeline.
-    Object.defineProperty(result, 'destroy', {
-      value: entriesStream.destroy.bind(entriesStream),
-    })
-    return result
+    return pipeline(entriesStream, filterStream, noop)
   }
 
   /**
--- a/src/blob-store/index.js
+++ b/src/blob-store/index.js
@@ -131,7 +131,7 @@ export class BlobStore extends TypedEmitter {
    * @param {object} opts
    * @param {boolean} [opts.live=false] Set to `true` to get a live stream of entries
    * @param {import('./utils.js').GenericBlobFilter | null} [opts.filter] Filter blob types and/or variants in returned entries. Filter is { [BlobType]: BlobVariants[] }.
-   * @returns
+   * @returns {BlobStoreEntriesStream}
    */
   createEntriesReadStream({ live = false, filter } = {}) {
     const entriesStream = createEntriesStream(this.#driveIndex, { live })
--- a/src/blob-store/index.js
+++ b/src/blob-store/index.js
@@ -137,7 +137,13 @@ export class BlobStore extends TypedEmitter {
     const entriesStream = createEntriesStream(this.#driveIndex, { live })
     if (!filter) return entriesStream
     const filterStream = new FilterEntriesStream(filter)
-    return pipeline(entriesStream, filterStream, noop)
+    const result = pipeline(entriesStream, filterStream, noop)
+    // Destroying the pipeline should destroy the *first* stream, not the
+    // whole pipeline.
+    Object.defineProperty(result, 'destroy', {
+      value: entriesStream.destroy.bind(entriesStream),
+    })
+    return result
   }
 
   /**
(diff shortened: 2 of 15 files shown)


src/blob-store/downloader.js
EvanHahn, line 1: Again, I recommend [reviewing this PR one commit at a time](https://github.com/digidem/comapeo-core/pull/956/commits).

src/blob-store/index.js
```diff
@@ -137,7 +137,13 @@ export class BlobStore extends TypedEmitter {
...
     if (!filter) return entriesStream
     const filterStream = new FilterEntriesStream(filter)
-    return pipeline(entriesStream, filterStream, noop)
+    const result = pipeline(entriesStream, filterStream, noop)
+    // Destroying the pipeline should destroy the *first* stream, not the
+    // whole pipeline.
+    Object.defineProperty(result, 'destroy', {
+      value: entriesStream.destroy.bind(entriesStream),
+    })
+    return result
```
gmaclennan, line 146: I don't quite understand why this is needed? `result` here is actually `filterStream`. My understanding of `pipeline` is that it ensures that destroying any stream in the pipeline also destroys the others. This seems to be overriding the `destroy` method of `filterStream` to destroy `entriesStream`. What is the intent of this?

EvanHahn: I don't love this solution, but it was the best I could come up with.

First, as a given: we need to be able to destroy the result of this function. Here's that code:

https://github.com/digidem/comapeo-core/blob/7c810c217fc08e25a5ba1c9ce7ab684981b1cde5/src/mapeo-project.js#L438

We need to be able to destroy `entriesStream` specifically. Calling `destroy()` on the result of `pipeline()` causes an error (I recall something like "premature close").

I considered a few other options, none of which I liked:

- Change the requirements and *don't* destroy the stream. I don't think this is a good idea because it'd leave streams around.
- Call `destroy()` on the result of `pipeline()` and handle the rejection. I don't think this actually cleans things up, but it stops the error, I think?
- Return multiple values; maybe something like `{ resultStream, destroyEntriesStream }` or something. I think that muddies the consumer's API, but it would also work. Also, what would it look like if we didn't include a filter?

Hope this answers the question.

gmaclennan: Thanks, got it. So, a few things (streams are an absolute headache):

1. `ERR_STREAM_PREMATURE_CLOSE` is an expected error from destroying a pipeline before it is finished, this is fine and should be handled.
2. Normally when creating a stream like [you do here](https://github.com/digidem/comapeo-core/blob/7c810c217fc08e25a5ba1c9ce7ab684981b1cde5/src/mapeo-project.js#L438), you would need to attach an `error` listener, to avoid uncaught errors. However, iterating over the stream attaches the error listener for you, and throws with it inside the iterator, so you don't need to attach an error listener here.
3. You do need to wrap the async iterator of `entriesReadStream` with `try {} catch {}` and check `error.code === 'ERR_STREAM_PREMATURE_CLOSE'` - you can ignore that error, but other errors should be logged.
4. You also need to clean up the stream references once the entriesReadStream closes, via `entriesReadStream.on('close', () => if (entriesReadStreams.get(peerId) === entriesReadStream) entriesReadStream.delete(entriesReadStream))`
5. You can remove the checks on `entriesReadStream.destroyed` in the iterator - catching the iterator being thrown (with `ERR_STREAM_PREMATURE_CLOSE`) is the correct way to handle a stream being destroyed.

gmaclennan: TL;DR:

1. Remove the `Object.defineProperty` code
2. Catch errors in the iterator that reads the stream
3. Remove references to closed streams

gmaclennan: Edit: add cleanup after the iterator loop rather than on the close event (it's the same thing, just avoids mixing two syntaxes).

src/sync/core-sync-state.js
```diff
@@ -182,13 +182,23 @@ export class CoreSyncState {
    * blocks/ranges that are added here
    *
    * @param {PeerId} peerId
-   * @param {Array<{ start: number, length: number }>} ranges
+   * @param {number} start
+   * @param {number} length
+   * @returns {void}
    */
-  setPeerWants(peerId, ranges) {
+  addWantRange(peerId, start, length) {
```
gmaclennan, line 189: I was worried about performance (because of the cost of calculating state), but I _think_ the way this is implemented, the throttle on SyncState should avoid state calculation happening too frequently, so I think it is ok to just call this with every range (there will be a lot of calls to this).

gmaclennan (changes requested):
ok looks good, good work on this, just needs some work managing the lifecycle of the entries stream.

src/sync/core-sync-state.js
EvanHahn: I wrote a [simple benchmark script with Deno](https://gist.github.com/EvanHahn/c03d3b3cb88a6dc708782b83f52a6be1). On my machine, I could call `setRange` 7,364 times per second, at least with my simple benchmark. I realize my computer is probably much faster than most phones, but even a 100x slowdown is probably okay here.

I'm going to leave this as is, but let me know if that's wrong.

src/mapeo-project.js
```diff
@@ -430,6 +432,27 @@ export class MapeoProject extends TypedEmitter {
       getReplicationStream,
     })
 
+    /** @type {Map<string, BlobStoreEntriesStream>} */
+    const entriesReadStreams = new Map()
```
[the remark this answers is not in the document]
EvanHahn: We can't use `WeakMap` because it doesn't let you use strings as keys. :(

src/mapeo-project.js
```diff
@@ -430,6 +432,27 @@ export class MapeoProject extends TypedEmitter {
...
+    this.#coreManager.on('peer-download-intent', async (filter, peerId) => {
```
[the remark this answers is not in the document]
EvanHahn: Done in 13f9a3d882604252994f86bbabb956e88d6b2ded.

src/blob-store/index.js
EvanHahn: Thanks for this; I'm bookmarking it for later! Fixed in 13f9a3d882604252994f86bbabb956e88d6b2ded.

src/discovery/local-discovery.js
```diff
@@ -117,7 +118,7 @@ export class LocalDiscovery extends TypedEmitter {
 
     /** @param {Error} e */
     function onSocketError(e) {
-      if ('code' in e && e.code === 'EPIPE') {
+      if (getErrorCode(e) === 'EPIPE') {
```
EvanHahn, line 121: Not strictly related, but uses a new utility.

src/mapeo-project.js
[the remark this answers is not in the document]
gmaclennan: oh yeah, I remember that every time I think I have a use-case for WeakMap

src/sync/core-sync-state.js
gmaclennan: Sorry for the lack of clarity, I was thinking out loud. I was concerned about the performance not because of the cost of this function, but because it triggers a state update, and reading state is expensive (when cores get large, which they will do for the blob core). However, the way we currently trigger state updates, we don't actually calculate state as part of the `update` event, and then in SyncState we throttle the handler for `update` events so that `getState()` is only called ever 200ms, so with the current implementation this is all fine. The reason this was originally `ranges` was to avoid triggering a state-recalculation for every range set, but our throttle in SyncState solves that problem for now. We just need to remember this in the future.

gmaclennan (approved):
great, fantastic to have this working.

PR #959, opened 2024-11-12
feat: `getByDocId` can optionally throw

EvanHahn:
This adds a new option to `DataType.prototype.getByDocId()`, `mustBeFound`. If set to `false` and the document doesn't exist, it will resolve with `null` instead of throwing.

If the option is missing or set to `true`, the current behavior is maintained (throwing if the document is missing).

I think this is a useful feature on its own but will also hopefully make at least one [upcoming change][0] a little easier.

[0]: https://github.com/digidem/comapeo-core/issues/188


src/datatype/index.js
```diff
@@ -186,7 +200,7 @@ export class DataType extends TypedEmitter {
   }
 
   /**
-   * @param {MapeoDoc} doc
+   * @param {any} doc
```
EvanHahn, line 203: This change, and the one below, come from #951.

Issue #188, opened 2023-08-18
Validate Role records


gmaclennan:
## Description

Validating a role record requires tracing a "role chain" back to the project creator auth core (the core whose key is the project key), and checking in each case that the role defined has the capability to assign the role given.

Because roles can change over time (e.g. a device might start as a "member", then be changed to "coordinator", then be changed to "removed"), each role record has a `fromIndex` property, that defines the index from which the role applies in the auth core of the person to whom the role applies.

An example:

1. project creator invites device A, and writes their role to be 'coordinator'
2. device A invites device B, and writes device B's role to be 'member'.

Validating device B's role requires the following steps:

1. The role document for B will be written in device A's auth core. You can get this information from the versionId of the role document.
2. Look up the role of device A to check they have authority to assign a 'member' role:
    1. The role document of device A is written by the project creator, so we trust the "role chain".
    2. The role of A is 'coordinator', which has the capability to assign 'member' roles, so this role chain is valid.

### What is a "Role chain"

A "Role chain" is the chain of role assignments that can be traced back to the project creator. E.g. project creator assigns device A a role, then device A assigns a role to device B, and device B assigns a role to device C. This "chain" of role assignments from creator -> A -> B -> C we call a "role chain" (suggestions for a better name are welcome!). Using the core key of the auth core as the `docId` for each role document, combined with the key of the auth core where the role document is written, can be used to validate this role chain.

### Validation flowchart

This flowchart summarises all the steps to take to validate any role:

```mermaid
flowchart TD
    Invalid([Role chain <br/>is INVALID])
    Valid([Role chain <br/>is VALID])
    Start([Start: want to validate <br/>role of `coreKey`]) --> A
    A[Get Role from coreKey] --> B{Is role written in <br/>project creator core?}
    B ---> |Yes| Valid
    B --> |No| C[Get GrantorRole from <br/>coreKey of role record]
    C --> D{Is index of role <br/>record >= indexFrom <br/>in GrantorRole?}
    D --> |No| E{Does grantorRole link <br/>to previous versions?}
    E ---> |No| Invalid
    E --> |Yes| F[Get previous version <br/>of grantorRole]
    F --> D
    D --> |Yes| G{Does grantorRole <br/>allow assigning role?}
    G ---> |No| Invalid
    G ---> |Yes| H[Now validate <br/>grantorRole as role]
    H --> B
```

### Validating `fromIndex`

[...]

PR #967, opened 2024-11-18
feat: validate role records

EvanHahn:
_Depends on #968._

When fetching a role, we now validate that the role is valid, returning a blocked role if it's not.

See #188 for more details.


EvanHahn:
I made a few changes based on discussions with @gmaclennan. I also did some quick profiling.

I wrote a test that creates 20 managers and then invites them one by one. Manager 1 invites Manager 2, Manager 2 invites Manager 3, and so on.

<details>
<summary>Expand this to see the test code.</summary>

```javascript
// This code is not perfect, but is good enough for this test.
test('perf test', async (t) => {
  const managers = await createManagers(20, t)
  const [creator] = managers

  const projectId = await creator.createProject({ name: 'Mapeo' })

  const disconnect = connectPeers(managers)
  t.after(disconnect)

  console.time('invite chain')
  /** @type {undefined | MapeoManager} */
  let previousManager
  for (const manager of managers) {
    if (previousManager) {
      console.log(previousManager.deviceId, 'inviting', manager.deviceId)
      await invite({
        projectId,
        invitor: previousManager,
        invitees: [manager],
        roleId: COORDINATOR_ROLE_ID,
      })
    }
    previousManager = manager
  }
  console.timeEnd('invite chain')

  const firstManager = managers[0]
  const firstProject = await firstManager.getProject(projectId)

  const lastManager = managers[19]

  console.time('looking up role')
  const lastManagerRole = await firstProject.$member.getById(
    lastManager.deviceId
  )
  assert.equal(lastManagerRole.role.roleId, COORDINATOR_ROLE_ID)
  console.timeEnd('looking up role')
})
```
</details>

_Before_ this change, on my machine:

- doing all the invites: 41.274s
- looking up the role: 0.174ms

_After_ this change:

- doing all the invites: 49.885s
- looking up the role: 6.961ms

Looking up the role is significantly slower, but I still think it's in an acceptable range.

src/roles.js
```diff
@@ -270,31 +276,156 @@ export class Roles extends TypedEmitter {
...
+
+    /** @type {null | ReadonlyDeep<MembershipRecord>} */
+    let currentMembershipRecord = membershipRecord
+
+    while (currentMembershipRecord) {
+      const parentMembershipRecord = await this.#getParentMembershipRecord(
+        currentMembershipRecord
+      )
+      switch (parentMembershipRecord) {
+        case null:
```
gmaclennan, line 336: Maybe be more explicit and return `false` here, since that is what will happen anyway.

src/roles.js
```diff
@@ -270,31 +276,156 @@ export class Roles extends TypedEmitter {
...
+          break
+        case CREATOR_MEMBERSHIP_RECORD:
```
gmaclennan, line 338: I think it's slightly risky relying on strict equality here. We may make a change in the future that means that this is not strictly equal, and not realise the consequences.

I think we should match based on `parentMembershipRecord.roleId === CREATOR_ROLE_ID`.

This means you can't write a switch statement, which is my hidden agenda for this comment... ha.

I do think having 3 `if` statements will make this a bit more readable, with comments for each statement just explaining what that condition means.

src/roles.js
```diff
@@ -270,31 +276,156 @@ export class Roles extends TypedEmitter {
    * @returns {Promise<Role>}
    */
   async getRole(deviceId) {
-    const roleRecord = await this.#dataType
+    const membershipRecord = await this.#getMembershipRecord(deviceId)
+
+    switch (membershipRecord) {
+      case null:
+        return NO_ROLE
+      case CREATOR_MEMBERSHIP_RECORD:
```
gmaclennan, line 284: See comment below about relying on strict equality here.

src/roles.js
```diff
@@ -270,31 +276,156 @@ export class Roles extends TypedEmitter {
...
+   * @param {ReadonlyDeep<MembershipRecord>} membershipRecord
+   * @returns {Promise<boolean>}
+   */
+  async #isRoleChainValid(membershipRecord) {
+    if (membershipRecord.roleId === LEFT_ROLE_ID) return true
+
+    /** @type {null | ReadonlyDeep<MembershipRecord>} */
+    let currentMembershipRecord = membershipRecord
+
+    while (currentMembershipRecord) {
```
gmaclennan, line 331: This should have a max-depth check, to avoid a bug/hack that might create a circular reference so that this never ends.

src/roles.js
```diff
@@ -270,31 +276,156 @@ export class Roles extends TypedEmitter {
...
+      .catch(() => null)
+    if (!assignerDeviceId) return null
+
+    const latestMembershipRecord = await this.#getMembershipRecord(
+      assignerDeviceId
+    )
+
+    /** @type {null | typeof CREATOR_MEMBERSHIP_RECORD | MembershipRecord} */
+    let membershipRecordToCheck = latestMembershipRecord
+    while (membershipRecordToCheck) {
```
gmaclennan, line 390: This should have a max-depth check, to avoid a bug/hack that might create a circular reference so that this never ends.

gmaclennan (commented):
Great work on this, the logic is tricky but clear enough to follow in review.

Can you add additional test(s) for a more complex scenario, e.g. a role chain of length > 2. I think it's ok to add a TODO for testing forked membership records. For forked membership records we also need to change the getWinner for membership records (to match the logic we use for looking up `links` - a doc only has multiple `links` if it was forked and then was subsequently merged.).

I think there are a few other cases that you are handling in the code that are not in the tests, it might be helpful to run a coverage test and see what code paths within the new code are being tested right now.

I am a bit concerned about the slow-down - it's a 40x slow down, and it could be more on phones depending on whether it is CPU limited or disk speed limited. If we start checking roles for reads of all data types, this could introduce a significant overhead. I think it's ok that we do not address this right now, and we think of a solution down the line when performance becomes an issue. Before we do that, we should spend some time adding perf metrics that get reported to Sentry.

EvanHahn (commented):
I spent a little more time thinking about this and I think I need to change the approach.

Currently, if we encounter an invalid membership record, we treat that member as blocked. I think that's not bad because a member could write a bogus membership forever block a user from a project.

For example, imagine the following:

1. Creator invites A and B, both as regular members.
2. A claims that B is a coordinator. This is bogus because it lacks the permission.

Currently, this PR will treat B as *blocked*. Instead, I think we should ignore A's claim; B's role should still be "member".

In other words, I think we should ignore invalid membership records.

I'll try to implement that next week.

PR #983, opened 2025-01-07
fix(maps): enable CORS for style json response

### LINKED ISSUE #981: [maps] Potential CORS issue when proxying online map style jsons
Early tests with the desktop app showed that attempting to use a mapbox style url as the default online style resulted in CORS issues:

Image

Seems like adding `'Access-Control-Allow-Origin': '*',` to https://github.com/digidem/comapeo-core/blob/7639b8dec0588600e5bea482bc3eda11683cae7b/src/fastify-plugins/maps.js#L123 resolved the issue.

achou11:
Fixes #981

### DIFF
--- a/src/fastify-plugins/utils.js
+++ b/src/fastify-plugins/utils.js
@@ -43,16 +43,3 @@ export async function getFastifyServerAddress(server, { timeout } = {}) {
 
   return 'http://' + addr
 }
-
-/**
- * @param {Readonly<Date>} lastModified
- */
-export function createStyleJsonResponseHeaders(lastModified) {
-  return {
-    'Cache-Control': 'max-age=' + 5 * 60, // 5 minutes
-    'Access-Control-Allow-Headers':
-      'Authorization, Content-Type, If-Match, If-Modified-Since, If-None-Match, If-Unmodified-Since',
-    'Access-Control-Allow-Origin': '*',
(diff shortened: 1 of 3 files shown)


src/fastify-plugins/utils.js
```diff
@@ -43,16 +43,3 @@ export async function getFastifyServerAddress(server, { timeout } = {}) {
...
- */
-export function createStyleJsonResponseHeaders(lastModified) {
-  return {
-    'Cache-Control': 'max-age=' + 5 * 60, // 5 minutes
-    'Access-Control-Allow-Headers':
-      'Authorization, Content-Type, If-Match, If-Modified-Since, If-None-Match, If-Unmodified-Since',
-    'Access-Control-Allow-Origin': '*',
-    'Last-Modified': lastModified.toUTCString(),
-  }
-}
```
achou11, line 58: this function is no longer used but does imply that we used to do the right thing 😄 wondering if any of these headers should be added as well 🤔

gmaclennan: not sure, as long as it works without them 🤷 
I think we should (outside this PR) figure out caching headers and last-modified so that if we change the style then it overrides the cache, but I think using without a cache is fine for now.

gmaclennan (approved):
looks good.

PR #984, opened 2025-01-07
fix: restore installation of @node-rs/crc32 optional deps

achou11:
4f33bd0 updated the lockfile to remove the native binaries produced by `@node-rs/crc32`, which prevents me from locally running anything that touches `yauzl-promise` (which uses that module). This has happened in the past and I honestly don't really understand what the expected behavior should be or why this happens in general.

The modules in question are marked as optional deps in the [published package](https://www.npmjs.com/package/@node-rs/crc32?activeTab=code), but not sure how npm decides to add/remove these from the lockfile. And if they're considered optional, not sure why `@node-rs/crc32` complains if the relevant binaries are missing at runtime (i.e. not _truly_ optional by definition).

~Explicitly marking the module as a dependency for us seems to restore the lockfile so that the native binaries are included, but my guess is that this isn't the desired solution to the problem.~ EDIT: I've removed as an explicit dep (see https://github.com/digidem/comapeo-core/pull/984/commits/ce342460270daa0955751694cc43106124e6f0ad) but the lockfile changes are kept in place


achou11:
Ah, a hint!

https://github.com/napi-rs/node-rs/blob/11a9a1f37f9b42899071789f1232556420600983/packages/crc32/index.js#L358-L361

I've definitely tried the node_modules reinstallation approach, but that doesn't seem to work for me (given that the lockfile doesn't specify any of the binaries as deps)

gmaclennan (approved):
looks ok, however can you check the next APK we build with this change, to ensure we aren't unnecessarily shipping a bunch of platform binaries that are included in this change?

gmaclennan:
FYI I continue to have issues locally on Apple silicon where the Darwin arm binaries are not installed and I need to install them manually. Seems to work ok on CI though which I don't understand

achou11:
> looks ok, however can you check the next APK we build with this change, to ensure we aren't unnecessarily shipping a bunch of platform binaries that are included in this change?

confirmed that this doesn't have any effect on the APK that's generated

PR #985, opened 2025-01-22
chore: update comapeo-schema

gmaclennan:
this removes support for migration metadata for now


achou11 (changes requested):
Shouldn't this revert https://github.com/digidem/comapeo-core/commit/2cc85557a8f88b5bd70a6e4e42624126ebfa13c0 as it introduced a database migration that we don't want to include?

Issue #990, opened 2025-02-07
RFC: Improvements to invite API


gmaclennan:
The invite API is proving challenging to use on the client side. The API exposes `getPendingInvites()` to retrieve a list of received invites that have not been responded, and a `invite-removed` event when an invite is removed from this list, as well as a `invite-received` event when an invite is added to that list. The challenge is that because API calls are asynchronous, the client cannot guarantee the relationship between the `invite-received` and `invite-removed` events and the list of pending invites from `getPendingInvites()`, which can result in race-condition edge-cases and complicated client code to handle both events and pending invites.

This is an idea to change the API so to replace `getPendingInvites()` with `getInvites()` that returns all invites including those that have been cancelled, errored or responded to. We should limit the max invites that are stored, and add a TTL (e.g. with an LRU) so that we avoid potential memory issues if a bad client sends multiple invites. The invites returns could include a state:

- `received`
- `accepting` (when an accept response has been sent, but has not received project details back)
- `accepted`
- `already` (already on project - currently `accepted other`)
- `rejected`
- `error` (combining both internal and connection errors)
- `canceled` - should this be treated as an error? That could make a client implementation easier.

An api client needs to handle changes of state differently depending on whether they are initiated by the client, or by the remote peer. E.g. if a remote peer cancels an invite, then the client needs to respond to that event, but an error resulting from trying to accept an invite would be caught in an asynchronous "accept" method called by the client, so the client would not need to respond to the state change in the same way.

We could keep the `invite-received` event, and change `invite-removed` to `invite-updated`. A client could either re-request all invites on every `invite-updated` (e.g. by invalidating a query key if using tanstack query) or by updating a cached invite list (e.g. interacting with the query cache for tanstack query).

This change would mean that the client can rely on `getInvites()` as the "source of truth" of invite state, rather than having to manage state separately by listening to the `invite-removed` event, which contains the "removalReason" which is not available in any other API.

PR #991, opened 2025-02-12
feat!: new invite API with finite invite states

gmaclennan:
The goal of this PR is to improve the invite API as proposed in #990. This PR also switches to modelling invites with a State Machine, implemented with [XState](https://stately.ai/docs), which should hopefully make errors much more unlikely, at the cost of increased verbosity of the code.

Invites are kept in state, and have a state property that can be one of these states:

- `pending`: invite has been received, awaiting either cancelation by sender, or response by receiver.
- `responding`: response is being sent for invite (can be delayed waiting for peers to finish connecting, but normally almost instant)
- `joining`: accept response has been sent and in the process of joining a project. The sub-states are not currently exposed, but include waiting for the project details, adding the project, then doing an initial sync.
- `joined`: receiver has joined the project from the invite.
- `canceled`: invite was canceled by sender
- `error`: an error occurred, which could include a disconnection while trying to send a response, or a timeout waiting for a response (often caused by a disconnect)
- `rejected`: invite was responded with a rejection
- `respondedAlready`: invite was responded informing sender that receiver is already on the project

Any time the sender or receiver try and fail to send a message, the invite moves into the error state. When waiting for a response, or waiting in a state for a user-initiated event, we do not move to the error state if a peer disconnects, because it could re-connect.

The unit tests feel very verbose and are a high maintenance burden. In a later follow-up it would be good to convert these to end-to-end tests where possible, or at least "half end-to-end tests" where one side can be a full CoMapeo instance, and the other can be just an RPC class in order to create unusual states.


RangerMauve (approved):
This looks good but IMO we should add one more check on the inviter side to be certain that the project data was received. Either we need another message from the joiner or some sort of ack at the RPC level

PR #1000, opened 2025-03-12
fix: Get sodium-native working on Linux

RangerMauve:
I was unable to run the project on Linux due to a [glibc versioning issue](https://github.com/holepunchto/sodium-native/issues/214) which got fixed in the latest sodium-native.

After updating the dependency it seems to be working and the tests are passing.

I added the dependency to the top level so we wouldn't need to try updating any of the dependencies that try to pull it in. I checked to make sure there's no duplication and that no other modules got changed in the lockfile.

Is there anything else we should check before merging like whether nodejs-mobile still accepts this version? I tested with node 18 so I'm guessing we're good to go.


achou11:
> Is there anything else we should check before merging like whether nodejs-mobile still accepts this version? I tested with node 18 so I'm guessing we're good to go.

Not sure if worth checking before merging this, but one eventual follow-up task will be to update the the relevant [nodejs-mobile prebuild repo](https://github.com/digidem/sodium-native-nodejs-mobile) to create a prebuild release for this version of sodium-native, so that prebuilds used in the apps actually match the version used in core.

The default branch is a bit outdated in that repo since the tooling has changed. The relevant branch to look at is https://github.com/digidem/sodium-native-nodejs-mobile/tree/shared-workflow. Not sure why the changes there haven't been merged into the main branch, @gmaclennan has more context there.

gmaclennan:
Before merging we need to check that this works on mobile across different devices, ideally by running the e2e mobile tests on browser stack, because native dependency updates in the past have broken builds on mobile for obscure architecture-dependent reasons, and we want to avoid this blocking core updates on mobile.

RangerMauve:
@gmaclennan Is running the e2e tests a good target for me to take on and get acquainted with this stuff?

Also, thoughts on whether the failing tests is being caused by the upgrade:

```
 ...
# BlobStore error HypercoreError: REQUEST_CANCELLED: Request was cancelled
#     at BlockRequest._cancel (/home/runner/work/comapeo-core/comapeo-core/node_modules/hypercore/lib/replicator.js:70:21)
#     at BlockRequest.detach (/home/runner/work/comapeo-core/comapeo-core/node_modules/hypercore/lib/replicator.js:46:10)
#     at Replicator.clearRequests (/home/runner/work/comapeo-core/comapeo-core/node_modules/hypercore/lib/replicator.js:1062:19)
#     at Hypercore._close (/home/runner/work/comapeo-core/comapeo-core/node_modules/hypercore/index.js:427:23)
#     at Hypercore.close (/home/runner/work/comapeo-core/comapeo-core/node_modules/hypercore/index.js:403:25)
#     at Hypercore._closeAllSessions (/home/runner/work/comapeo-core/comapeo-core/node_modules/hypercore/index.js:557:42)
#     at Hypercore.purge (/home/runner/work/comapeo-core/comapeo-core/node_modules/hypercore/index.js:781:16)
#     at CoreManager.deleteOthersData (file:///home/runner/work/comapeo-core/comapeo-core/src/core-manager/index.js:489:34)
#     at file:///home/runner/work/comapeo-core/comapeo-core/src/mapeo-project.js:877:27
#     at Array.flatMap (<anonymous>) {
#   code: 'REQUEST_CANCELLED'
# }
```

It smells like a race condition to me but I'm not sure if this has been seen before.

gmaclennan:
> @gmaclennan Is running the e2e tests a good target for me to take on and get acquainted with this stuff?

We have e2e tests set up to run on mobile, and they run on CI, but you can also trigger them from a local build. Both options could be a rabbit hole:

1. Publish a prerelease with this change, and integrate it into a branch on mobile, and test that via CI (APK will be built on CI and sent to browserstack). Currently comapeo-core is a transitive dep of @comapeo/ipc, so it's a bit of a pain. You could probably just point npm to github branches for the source, since there is no build step in either repo I don't think.
2. use `npm pack` (or `npm link`) locally, and create a local build of the APK, and run browserstack e2e tests locally - it can be a pain setting up APK builds on your machine, but might be needed at some point.

Regarding the failing test, that does smell like a race condition to me, but it hasn't been seen before. It could be a hypercore bug - we are several versions behind, due to breaking changes (we use some hypercore internals) that break our tests. From the logs it looks like hypercore is still trying to replicate something when we are trying to purge it, and for some reason it's throwing an error during that, which is unexpected for me.

RangerMauve:
Reran the tests so it seems the error wasn't from sodium-native upgrades.

I feel like tagging the prerelease and running through CI might be more resilient since my available android devices are more limited (e.g. maybe a specific android version would get buster). Do I have the level of permissions needed to do something like that? I think I still need to be added to browserstack.

gmaclennan:
Because sodium-native is not a direct dependency of comapeo-core, we have been relying on the [package-lock.json in the comapeo-mobile repo](https://github.com/digidem/comapeo-mobile/blob/develop/src/backend/package-lock.json) to avoid dep changes, however we have experienced that to be fragile because it is easy to accidentally update a transitive dep in the lockfile without noticing. However, this does make it easy to check the version change without needing to publish anything - in the mobile repo we can just update `sodium-native` via `cd src/backend && npm update sodium-native` which [I did in this PR](https://github.com/digidem/comapeo-mobile/pull/1039). I also generated [prebuilds of sodium-native@4.3.3](https://github.com/digidem/sodium-native-nodejs-mobile/releases/tag/4.3.3) (generating prebuilds with the new prebuild script is not done via a [workflow dispatch here](https://github.com/digidem/sodium-native-nodejs-mobile/actions/workflows/prebuilds.yml). It's fine to generate as many prebuilds as we want - the mobile build script will download the prebuild relating to the version installed.

As you can see, the e2e tests failed. Looking at the ADB logs from browserstack, I see this:

```
03-14 04:31:05.509 E/nodejs  (24051): file:///data/data/com.comapeo.rc/files/nodejs-project/index.js:29234
03-14 04:31:05.509 E/nodejs  (24051): 	    throw new Error(
03-14 04:31:05.509 E/nodejs  (24051): 	          ^
03-14 04:31:05.509 E/nodejs  (24051): 
03-14 04:31:05.509 E/nodejs  (24051): Error: Cannot find addon '.' imported from 'file:///data/data/com.comapeo.rc/files/nodejs-project/index.js'
03-14 04:31:05.509 E/nodejs  (24051):     at Function.addon (file:///data/data/com.comapeo.rc/files/nodejs-project/index.js:29234:12)
03-14 04:31:05.509 E/nodejs  (24051):     at file:///data/data/com.comapeo.rc/files/nodejs-project/index.js:29283:28
03-14 04:31:05.509 E/nodejs  (24051):     at ModuleJob.run (node:internal/modules/esm/module_job:195:25)
03-14 04:31:05.509 E/nodejs  (24051):     at async ModuleLoader.import (node:internal/modules/esm/loader:337:24)
03-14 04:31:05.509 E/nodejs  (24051):     at async loadESM (node:internal/process/esm_loader:34:7)
03-14 04:31:05.509 E/nodejs  (24051):     at async handleMainPromise (node:internal/modules/run_main:106:12)
03-14 04:31:05.509 E/nodejs  (24051): Node.js v18.20.4
```

So it looks like a change in how sodium-native is built and requires native modules is not compatible with our rollup build process.

Regarding running e2e tests, they are configured on mobile to run for any PR right now. @RangerMauve I've sent an invite to you for BrowserStack (and it looks like someone else has done the same) do you can review the ADB logs yourself.

I won't have time to investigate this more, but hopefully this can get you started.

gmaclennan:
Alternatively, because the sodium-native version for mobile is pinned in the lockfile, if you change this PR to only change the version in package-lock (by running `npm update sodium-native`) without adding it to the package.json, then this change will have no effect on mobile (because mobile will continue to use the v4.3.1 from its package-lock) and it will enable you to run e2e tests.

gmaclennan:
The issue with this not working on mobile comes from the [change in how sodium-native requires the native module](https://github.com/holepunchto/sodium-native/compare/v4.3.1...v4.3.3), and our bundle script which uses a [fragile plugin](https://github.com/digidem/comapeo-mobile/blob/develop/src/backend/scripts/rollup-plugin-native-paths.mjs) to keep the correct path to native modules from the bundled code.

RangerMauve:
I'm going to try updating just the package-lock to see if that helps. I think if I run into more trouble I should put this issue on hold and focus on the RPC reliability for this week with amnual fixes for linux just so I don't get stuck in the weeds.

IMO once there's time for it I should look into the rollup build and make it compatible with this new native version since this will likely block updating hypercore deps otherwise.

gmaclennan:
> IMO once there's time for it I should look into the rollup build and make it compatible with this new native version since this will likely block updating hypercore deps otherwise.

Agreed, well, I think someone should do it, you or @achou11. Can you open an issue for this and add it to the sprint backlog?

gmaclennan (changes requested):
Just trying to clean-up my inbox, and I believe the plan is to change this to only a package-lock change, so that it does not affect the front-end for now, which will continue to use the slightly older sodium-native, because the newer version here is not compatible with how we bundle the code on mobile.

Issue #989, opened 2025-02-07
Improve local peers RPC resilience


gmaclennan:
CoMapeo maintains an [RPC connection with local peers](https://github.com/digidem/comapeo-core/blob/bf52641913baab30a59409a895e9389962ef197c/src/local-peers.js) for sharing device info (name and device type) and sending and receiving invites. Because connections are peer-to-peer, it's possible that each peer connects to the other at the same time, resulting in duplicate connections. This deduplication is [handled in local-discovery.js](https://github.com/digidem/comapeo-core/blob/6b1e97180c9e1cd560adc957ce1ffba451eeb039/src/discovery/local-discovery.js#L207), which closes a duplicate connection in a deterministic way (e.g. both sides will close the same connection). The RPC module needs to handle this duplication, as well as handle connections that are still in the "connecting" state (noise handshake is taking place) and disconnections. Currently this is partially handled in [`chooseDevicePeer()`](https://github.com/digidem/comapeo-core/blob/6b1e97180c9e1cd560adc957ce1ffba451eeb039/src/local-peers.js#L709-L739), however there are possible edge-cases where this is used when sending an RPC message and the connection chosen is closed before the RPC message reaches the destination.

We need to make RPC more resilient to connection deduplication and connections being lost. It seems like on flakey wifi connections, devices can loose their connections with each other. We need to try and make RPC calls more resilient to flakey connections, e.g. by sending to all current connections rather than just picking one, and possibly sending and ACK and retrying delivery.

This issue is to investigate edge-cases when device connections are deduplicating or being closed and re-opened, and add workaround to improve RPC message deliverability.

RangerMauve:
@gmaclennan do we have acks/timeouts/retries in the RPC codebase? I'd probably start with simulating some of these breaks in tests and see what additional mechanisms could help mitigate the issues.

RangerMauve:
Is there a report for how the issue manifested in the field? It'd be good to know the exact steps to reproduce this since it's not obvious where stuff might be going wrong.

My main guess is the logic within [waitForPendingConnections](https://github.com/digidem/comapeo-core/blob/bf52641913baab30a59409a895e9389962ef197c/src/local-peers.js#L575) could be revisited. I'm not sure though. From what I'm seeing the code is pretty resilient against connection duplication issues.

I'm going to dig more into how re-opening works to see if there's something amiss in there.

[...]

Issue #903, opened 2024-10-16
Question: Where to store attachment metadata?


gmaclennan:
We currently have two places where we can store information about an attachment: In an [attachment record](https://github.com/digidem/comapeo-schema/blob/main/proto/observation/v1.proto#L28-L33), which is part of an observation, or using the arbitrary JSON file metadata [supported by hyperdrive](https://github.com/holepunchto/hyperdrive?tab=readme-ov-file#const-ws--drivecreatewritestreampath-options).

Currently in the attachment record we store:

- Attachment name & hyperdrive id (which identify the linked file(s) in hyperdrive)
- Hash of the original attachment
- type of the attachment (photo, video etc)

Previously we stored just the mimeType in hyperdrive metadata, but now we are storing some photo metadata in there too.

I think there is a difference between "an attachment" and "a file/blob". We generate multiple versions of some attachments (e.g. photos), so there is more than one file per attachment.

It feels like the "correct" thing to do is put information about the _file_ in the hyperdrive metadata, and information about the _attachment_ in the attachment record, although I'm not sure there is a clear logical distinction between these two.

The advantage of attachment records is that they are stored with protobuf and we have some guarantees about the structure / type of the data. The metadata from hyperdrive could be arbitrary JSON, so we kind of need to treat it as `unknown` and validate it to get what we want.

Another advantage of attachment records is that the information is available with the observation, it does not require additional requests.

Keeping information in metadata also requires a separate approach for accessing a history of the information in there and validating signatures.

For me it _feels_ like most additional information should be in the attachment record, although I don't feel able to make a strong argument for that. Plus we are currently putting additional metadata into the hyperdrive metadata records... so...

I would welcome feedback and opinions on this! I think it's early enough that we could move what we currently have in hyperdrive metadata into attachments and create a basic fallback.

EvanHahn:
I agree.

Advantages of putting metadata on the `attachments` property of observations:

- Structured data with Protobuf (as you say)
- Fewer database lookups (as you say)
- Metadata is available when the blob isn't available
- Easier querying (e.g., "give me all observations with a thumbnail")

Disadvantages:

- Slightly more difficult in a future where attachments could be part of multiple data types, e.g. tracks (probably not that bad)
- This refactor requires additional work

I would personally opt to put *all* blob metadata onto attachments, even the MIME type, because (1) it's a bit simpler to have all the data in one place (2) you could infer the attachment type from the MIME type. But I don't feel strongly about this detail.

[...]

PR #1001, opened 2025-03-13
feat: blobApi.getMetadata

RangerMauve:
- For the UI being added in the front end
- Checks the return type to best abilities to only accept plain objects
- Adds a test to make sure we're fetching the data


test/blob-api.js
achou11, line 1: thoughts on adding the following cases?

- no metadata
- metadata that doesn't have the expected shape. This one is only relevant if the return type is narrowed down to something more specific than the JSON object return type you currently have implemented. This one may be overcautious, but wondering if there's a case where something arbitrary can be stored as metadata? (e.g. due to a non-conforming client)

achou11 (changes requested):

src/blob-api.js
```diff
@@ -66,6 +67,32 @@ export class BlobApi {
     return base + `${driveId}/${type}/${variant}/${name}`
   }
 
+  /**
+   * Get meatdata for a blob based on its BlobId
+   * @param {BlobId} blobId
+   * @returns {Promise<JsonObject|null>}
```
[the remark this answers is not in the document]
RangerMauve: I was thinking of returning a `Partial<Metadata> | null`. Would the JSON Schemas be the right place to do validation from?

test/blob-api.js
RangerMauve: Yeah I'm thinking a return of `null` either means no metadata or invalid shape. Should I through an error instead? What'd be more useful for the front end?

src/blob-api.js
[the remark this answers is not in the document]
achou11: > I was thinking of returning a `Partial<Metadata> | null`.

That would work 👍 

> Would the JSON Schemas be the right place to do validation from?

I don't think [`comapeo-schema`](https://github.com/digidem/comapeo-schema/) applies for blobs (only references to them in the case of observation attachments, basically), so the validation would be kept here in core. I think for now, doing it in this method seems sufficient. Think the important thing is that the relevant methods from the `BlobApi` for creating and retrieving metadata are aligned.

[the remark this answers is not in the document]
RangerMauve: Are you sure it wouldn't make sense to use it? I think I'll need the same validation as [this](https://github.com/digidem/comapeo-core/blob/main/src/config-import.js#L126) unless I should make something more adhoc.

test/blob-api.js
RangerMauve: Lets leave it like this for now

src/blob-api.js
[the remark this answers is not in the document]
RangerMauve: I think the risk is low overall since we control both the reading and writing code. I've set the return to be `DeepPartial<Metadata>` to future proof for changes in the metadata shape and have the basic checks to see if it is a plain old json object.

RangerMauve:
@achou11 Mind re-reviewing?

src/blob-api.js
```diff
@@ -66,6 +67,32 @@ export class BlobApi {
     return base + `${driveId}/${type}/${variant}/${name}`
   }
 
+  /**
+   * Get metadata for a blob based on its BlobId
+   * @param {BlobId} blobId
+   * @returns {Promise<PartialDeep<Metadata>|null>}
```
achou11, line 73: is it useful to have the `PartialDeep` here? Seems like it'd be easier for frontend to use if it concretely returned `Promise<Metadata | null>`. I think this function takes on the extra work of guaranteeing that shape using validations. If you want the validation code to not be as tedious, you could use something like [`@sinclair/typebox`](https://github.com/sinclairzx81/typebox/tree/0.33.17), as we use this elsewhere:


1. Set up the schemas

```ts
const LocationObjectCoordsSchema = T.Object({
  accuracy: T.Union([T.Number(), T.Null()]),
  altitude: T.Union([T.Number(), T.Null()]),
  altitudeAccuracy: T.Union([T.Number(), T.Null()]),
  heading: T.Union([T.Number(), T.Null()]),
  latitude: T.Number(),
  longitude: T.Number(),
  speed: T.Union([T.Number(), T.Null()]),
})

const LocationObjectSchema = T.Object({
  coords: LocationObjectCoordsSchema,
  mocked: T.Optional(T.Boolean()),
  timestamp: T.Number(),
})

// 🟡 May want to specify `{ additionalProperties: true }` as the second option to `T.Object()` if we want to be more flexible
const MetadataSchema = T.Object({
  mimeType: T.String(),
  timestamp: T.Number(),
  location: T.Optional(LocationObjectSchema),
})

/**
 * @import {Static} from '@sinclair/typebox'
 *
 * @typedef {Static<typeof LocationObjectCoordsSchema>} LocationObjectCoords
 * @typedef {Static<typeof LocationObjectSchema>} LocationObject
 * @typedef {Static<typeof MetadataSchema>} Metadata
 */
```

2. Update implementation of `getMetadata()`

```ts
/**
 * Get metadata for a blob based on its BlobId
 * @param {BlobId} blobId
 * @returns {Promise<Metadata | null>}
 */
async getMetadata(blobId) {
  const entry = await this.#blobStore.entry(blobId)

  if (entry === null) {
    return null
  }

  const metadata = entry.value.metadata

  return Value.Check(MetadataSchema, metadata) ? metadata : null
}
```

RangerMauve: IMO this is a performance hit which won't necessarily help in the short term. All data going in will either be null or conform to the metadata schema.

The reason I was leaning towards using PartialDeep was to "future proof" for the schema changing, e.g. if we add fields then old data will be missing the fields and the front end could be made more robust if we handle it ahead of time.

If that doesn't seem useful right now I'd suggest we remove the DeepPartial wrapper and return `Promise<Metadata|null>` as is.

Does that sound okay or is the deep validation a hard requirement in your opinion?

achou11: > Does that sound okay or is the deep validation a hard requirement in your opinion?

Not a hard requirement for me since it's hard to imagine anything else storing this data without using core right now.

> The reason I was leaning towards using PartialDeep was to "future proof" for the schema changing, e.g. if we add fields then old data will be missing the fields and the front end could be made more robust if we handle it ahead of time.

Yeah this makes sense. Think the part I'm hesitant about is the "deep" aspect and how that affects consuming code. definitely will be more robust but will it be inconvenient?

> If that doesn't seem useful right now I'd suggest we remove the DeepPartial wrapper and return Promise<Metadata|null> as is.

I think I would lean towards omitting the DeepPartial for now, unless the API for storing the metadata is updated to align with it.

RangerMauve: Cool lets remove PartialDeep for now and rervisit when the metadata changes!

gmaclennan (changes requested):
Just catching up on this now. This is a reminder of unresolved work from before that has been hiding in a [buried issue](https://github.com/digidem/comapeo-core/issues/903). Sorry about that Mauve.

Our philosophy for CoMapeo Core has been strict type safety with resilience to whatever is stored in the database. We have not being relying on the assumption that what is read from the hypercores is put there by existing code that we control - we defensively consider that data in the hypercores could come from code that we do not control, or future code by us which changes things.

We have implemented this by implementing a [strict schema](https://github.com/digidem/comapeo-schema) and storing data as protobufs - there is a strict guarantee that whatever you read from the database has the type that it says it has, or if it doesn't then it will not be read at all.

Unfortunately the initial implementation of blob metadata didn't follow this approach, and used the convenient `metadata` option of the hyperdrive API. This doesn't have a schema or type guarantees. This is an approach that we [have planned to change](https://github.com/digidem/comapeo-core/issues/903) but got delayed last year then partially forgotten.

I think that before we build an important feature like photo validation with this data, we should implement #903, and read the metadata from there. I think at this stage it is acceptable to ignore existing metadata that is stored in hyperdrive metadata records, but if we leave it any longer (e.g. with larger groups collecting critical data with CoMapeo) we would have to build a migration strategy to always read the hyperdrive metadata as a backup for the attachment data. In the extreme case where one of our existing 78 users have recorded photos where getting the metadata is very important, then we can handle that on a case-by-case basis.

I think the type safety is particularly important for p2p, even if we assume that only code we control is writing data to the p2p database. For example, with the current implementation, if we do drop using hyperdrive metadata for storage, the type-casting currently assumes that `timestamp` always exists, so if a peer with this code synced with a peer that did not add that `timestamp` data, and there was front-end code assuming that type to be correct, then the app would crash. Similarly we don't have a strict record of when we started adding `timestamp` to the metadata, so we don't know if there is existing data "out there" without it.

In general, I think for any reads we need to fully validate anything we read. Because our current [`@comapeo/schema`](https://github.com/digidem/comapeo-schema) handles all of this already, I think it's best to also include the metadata as part of that.

The only exception I think is the mime-type, which also could be different for different blob variants in the future (imagine audio attachments with a preview with a different low-bandwidth encoding), so I think it makes sense for that to be stored for each _blob_ rather than for each _attachment_. Also, we can assign it as an optional property, and assume it can be gibberish or not a string, and validate it, and it is not valid then we guess the mime-type from the file itself, which [we already do.](https://github.com/digidem/comapeo-core/blob/cbe469649fd9a10f7d909decfee752e3dcd51e40/src/fastify-plugins/blobs.js#L117-L139)

RangerMauve:
Hmm, so what are some options to proceed? I was thinking that exposing `getMetadata` with strict schema validation could be a good start so that I can focus more on the Invite instability #989  before Luandro's workshop while unblocking the front end team. After that we could do #903

Alternately maybe I should do the RPC stuff first and then focus on the metadata stuff after? I just worry that focusing on #903 will take more time since I'll need to onboard onto more parts of the codebase.

RangerMauve:
Gonna put this on hold until after I tinker with the RPC stuff

RangerMauve:
Gonna discuss this more on our call tomorrow but we might want to just expose metadata with validation for now before going with the sqlite option. One other tradeoff is it'd mean way more data being loaded/parsed per observation vs one call on demand once the UI needs it.

RangerMauve:
Closing this in favor of storing in sqlite

PR #1002, opened 2025-03-25
fix: Wait for RPC messages to drain before returning

RangerMauve:
Right now we have a higher risk of false positives for RPC calls which may be related to #989 

This PR adds a check on calls to `channel[message].send()` to wait for data to be drained in the stream if its been queued up. Relevant code in protomux is [here](https://github.com/holepunchto/protomux/blob/main/index.js#L281).

If the connection closes before the write is successful then any inflight RPC calls will throw.

Any tips on how I should go about testing this? Existing tests are passing (outside the hypercore replication error being thrown which gregor addresses in another PR).


src/local-peers.js
```diff
@@ -157,53 +158,91 @@ class Peer {
     this.#disconnectedAt = Date.now()
     // This promise should have already resolved, but if the peer never connected then we reject here
     this.#connected.reject(new PeerFailedConnectionError())
+    for (const listener of this.#drainedListeners) {
+      listener.reject(new Error('RPC Disconnected before sending'))
+    }
```
gmaclennan, line 163:
```suggestion
    }
    this.#drainedListeners.clear()
```

gmaclennan (changes requested):
Looks good. Testing streams is really hard, especially when related to high water marks, but I think you could either pause the replication stream, pipe replication through a passthrough stream that you can pause, or cork the protomux, then assert sending a message throws, assert it hasn't been received, then destroy the streams.

I don't think our current tests have a utility to abruptly terminate the replication stream between peers, we need that too.

RangerMauve:
Yeah lets do a more thorough test suite for disconnects as a new action item.

PR #1012, opened 2025-04-08
chore(ci): update docs workflow to fix deprecation error

achou11:
The docs workflow is currently broken due to our version of `upload-pages-artifact` that uses a deprecated version of the `upload-artifact` action. Updating this action requires us to use a new version of the `deploy-pages` action as well.

Used https://github.com/suzuki-shunsuke/pinact to update the relevant actions to the exact SHA hashes as opposed to using git tags, which has better security implications. Happy to update the other actions in this workflow to do the same if desired.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -26,9 +26,9 @@ jobs:
           npm run doc:publish
       - id: upload-documentation
         name: Upload documentation to Pages
-        uses: actions/upload-pages-artifact@v2
+        uses: actions/upload-pages-artifact@56afc609e74202658d3ffba0e8f6dda462b719fa # v3.0.1
         with:
           path: docs/api/html
       - id: deployment
         name: Deploy documentation to GitHub Pages
-        uses: actions/deploy-pages@v2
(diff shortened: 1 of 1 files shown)


RangerMauve (approved):
Are we using SHA for versions in other workflows?

achou11:
> Are we using SHA for versions in other workflows?

The only other place is https://github.com/digidem/comapeo-core/blob/a3047d0f56881fcde24cb0da1427b531968cf28c/.github/workflows/lockfile.yml#L31, which probably makes more sense given that it's "more" third-party than the ones I've updated here (although it's an action that we maintain 😄 )

RangerMauve:
Fair. IMO if "higher security in github actions" is something we care about it should be done everywhere. Maybe add a TODO for it to the untracked sprint issues?

achou11:
> IMO if "higher security in github actions" is something we care about it should be done everywhere.

We haven't really reached an agreement around this as a team so it could be the case that it's not something we care about. Agree that we should apply this everywhere if it is something we're okay with though 👍

PR #1015, opened 2025-04-10
Feat: RPC acknowledgements, add role after project details are sent

### LINKED ISSUE #1009: Account for Invitee not getting Project Info
## Description

Sometimes when a person is being invited to a project, the message containing project info fails to send. Right now the inviter [adds the invitee](https://github.com/digidem/comapeo-core/blob/bf52641913baab30a59409a895e9389962ef197c/src/member-api.js#L170) to the project before they know that the invitee got the info and can actually join. This results in issues where the invitee needs to clear their app data and start fresh since there is no way for the inviter to send them a request as they see them as added again.

Instead we should make sure the invitee is onboarded before calling addRole. This can be done because the invitee will [wait for a project to sync](https://github.com/digidem/comapeo-core/blob/bf52641913baab30a59409a895e9389962ef197c/src/mapeo-manager.js#L619) as part of receiving the details. This sync happens when the [sync state updates](https://github.com/digidem/comapeo-core/blob/bf52641913baab30a59409a895e9389962ef197c/src/mapeo-manager.js#L694) which should happen when the protomux channel opens on [both sides](https://github.com/digidem/comapeo-core/blob/main/src/sync/sync-api.js#L181). We should also add a timeout to the members-api.accept for waiting on the confirm.

## Tasks

- [x] Add new confirm project details RPC call (0.5d)
- [x] Inviter should wait for it before adding role (0.5d)
- [x] Confirm sync can still work with this new flow (1.0d)
- [x] Add protocol version to device info to skip waiting for the confirm on older versions (1.0d)
- [x] Make sure tests pass, add test for older version peer (2d)
- [ ] Test with two devices, one on older version (1d) (QA team?)

RangerMauve:
- Added RPC acknowledge messages alongside the other RPC types
- It only enables if the other side says they have the "Ack" feature.
- This is facilitated by a new "features" flag in DeviceInfo which has a set of features we can add / remove over time. This is how we can facilitate protocol evolution over time
- This has been in service of knowing that invite actions and project details were actually seen on the other side
- We now only add a new member's role to a project after we know they got the project invite details. Fixes #1009
- Found some race conditions in the tests!

### DIFF
--- a/src/local-peers.js
+++ b/src/local-peers.js
@@ -194,8 +194,8 @@ class Peer {
       listener.reject(new Error('RPC Disconnected before sending'))
     }
     for (const waiters of this.#ackWaiters.values()) {
-      for (const { deffered } of waiters) {
-        deffered.reject(new Error('RPC disconnected before receiving ACK'))
+      for (const { deferred } of waiters) {
+        deferred.reject(new Error('RPC disconnected before receiving ACK'))
       }
     }
     this.#ackWaiters.clear()
@@ -242,13 +242,13 @@ class Peer {
     if (!this.#ackWaiters.has(type)) {
       this.#ackWaiters.set(type, new Set())
     }
-    const deffered = pDefer()
+    const deferred = pDefer()
     this.#ackWaiters.get(type)?.add({
-      deffered,
+      deferred,
       filter,
     })
 
-    await deffered.promise
+    await deferred.promise
   }
 
   /**
@@ -265,7 +265,7 @@ class Peer {
 
     for (const waiter of waiters) {
       if (waiter.filter(ack)) {
-        waiter.deffered.resolve()
+        waiter.deferred.resolve()
         waiters.delete(waiter)
       }
     }
--- a/src/local-peers.js
+++ b/src/local-peers.js
@@ -509,9 +509,6 @@ export class LocalPeers extends TypedEmitter {
    * @param {DeviceInfo} deviceInfo device info to send
    */
   async sendDeviceInfo(deviceId, deviceInfo) {
-    if (!deviceInfo.features) {
-      deviceInfo.features = []
-    }
     await this.#waitForPendingConnections()
     const peer = await this.#getPeerByDeviceId(deviceId)
     await peer.sendDeviceInfo(deviceInfo)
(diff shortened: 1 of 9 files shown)


src/local-peers.js
```diff
@@ -186,6 +224,53 @@ class Peer {
     await onDrain.promise
   }
 
+  /**
+   * Check if RPC Acknowledgement messages are supported by this peer
+   * @returns {boolean}
+   */
+  supportsAck() {
+    return this.#features?.includes(DeviceInfo_RPCFeatures.ack) ?? false
```
achou11, line 232: is the optional chaining necessary here?

```suggestion
    return this.#features.includes(DeviceInfo_RPCFeatures.ack) ?? false
```

achou11 (commented):
Overall the changes read well to me, but I don't feel positioned to be provide very substantial feedback.

Can you clarify how this affects compatibility with devices that are using the older implementation of the RPC? Is this a breaking change such that they can no longer communicate at all, or is it that they still can but this new implementation has to account for the `features` field not being available?

test/local-peers.js
```diff
@@ -307,6 +324,130 @@ test('connected peer has protomux instance', async () => {
...
+test('Device info with ack results in acks sent', async () => {
+  const r1 = new LocalPeers()
+  const r2 = new LocalPeers()
+
+  /** @type {import('../src/generated/rpc.js').DeviceInfo} */
+  const expectedDeviceInfo = {
+    name: 'mapeo',
+    deviceType: 'mobile',
+    features: [DeviceInfo_RPCFeatures.ack],
+  }
```
achou11, line 336: maybe a stylistic thing but was expecting the test to make some assertion on this because of the `expect` prefix in the name. found it a bit confusing that it's only used as an input (albeit the one that affects the result of the test)

src/local-peers.js
RangerMauve: Yes because the field is optional in the device Info

RangerMauve: Good point, it will never be undefined

test/local-peers.js
RangerMauve: renamed to `validDeviceInfo`

src/local-peers.js
```diff
@@ -355,6 +509,9 @@ export class LocalPeers extends TypedEmitter {
    * @param {DeviceInfo} deviceInfo device info to send
    */
   async sendDeviceInfo(deviceId, deviceInfo) {
+    if (!deviceInfo.features) {
+      deviceInfo.features = []
+    }
```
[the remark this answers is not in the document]
RangerMauve: Removed! It was a remnant of that field being optional

RangerMauve:
Regarding backwards compat, we don't enable the ack feature unless the other end sends a DeviceInfo with that feature enabled. Older devices don't see the features field so they ignore it.

achou11 (approved):

PR #1017, opened 2025-04-14
feat: remove server peer

### LINKED ISSUE #999: Remote archive: remove server / Phase 1
Description: This allows the user to remove the serve while online

Add method for removing a server peer. Removing a server peer means changing the peer's role to "blocked".

- Devices should not try to connect to a server that has the blocked role (or any role that does not allow sync of any namespaces), currently we do not check this before making the websocket connection (sync would still not happen, because we do check the role after syncing the auth namespace but before syncing other namespaces).
- Only coordinators should be able to remove the server peer (as defined in their capability to assign the Blocked role)
- The device that removes the server peer needs to sync the new role (server blocked role) to the server, so that other peers that connect to the server can "learn" of the server's removal.
- Should handle the edge-case of the coordinator writing the new "blocked" role for the server, but not having a connection to the server at that time, or loosing the connection before syncing. The device that removes the server should keep trying to connect until it confirms that the blocked role record has synced to the server. This retry should persist across app restarts.
- Devices that connect to the server peer, sync the disconnect record, do not need to disconnect immediately - existing sync code should stop data syncing once the removal role has synced - but they should not connect again once they do disconnect.

Background and user requirements here [internal access only]: https://www.notion.so/digidem/Remote-Archive-Remove-Server-14b1b08162d580cbac69e5ab52eaeebb?pvs=4

[Some test code](https://github.com/digidem/comapeo-core/blob/883d14bf802859b91f7371365e5ca50298d98f3d/test-e2e/server.js#L228-L286) has already been written for this.

RangerMauve:
Closes #999

Some specific stuff I'd like feedback on:
1. Right now this only allows removing while we're replicating. Should I initialize a replication stream if one doesn't exist instead of erroring?
2. Is removing the server by the base URL the easiest for the front end? Should we use the deviceId or some other identifier instead?
3. Are there any events we should be adding to let the front end know the server is removed, or is the role update for the server peer enough?
4. Any other tests I should include here?

### DIFF
--- a/src/member-api.js
+++ b/src/member-api.js
@@ -63,9 +63,7 @@ export class MemberApi extends TypedEmitter {
   #projectKey
   #rpc
   #getReplicationStream
-  #isConnectedToServer
   #waitForInitialSyncWithPeer
-  #waitForSyncAndDisconnect
   #dataTypes
 
   /** @type {Map<string, { abortController: AbortController }>} */
@@ -82,8 +80,6 @@ export class MemberApi extends TypedEmitter {
    * @param {import('./local-peers.js').LocalPeers} opts.rpc
    * @param {() => ReplicationStream} opts.getReplicationStream
    * @param {(deviceId: string, abortSignal: AbortSignal) => Promise<void>} opts.waitForInitialSyncWithPeer
-   * @param {(deviceId: string) => Promise<void>} opts.waitForSyncAndDisconnect
-   * @param {(deviceId: string) => boolean} opts.isConnectedToServer
    * @param {Object} opts.dataTypes
    * @param {Pick<DeviceInfoDataType, 'getByDocId' | 'getMany'>} opts.dataTypes.deviceInfo
    * @param {Pick<ProjectDataType, 'getByDocId'>} opts.dataTypes.project
@@ -99,8 +95,6 @@ export class MemberApi extends TypedEmitter {
     getReplicationStream,
     waitForInitialSyncWithPeer,
     dataTypes,
-    isConnectedToServer,
-    waitForSyncAndDisconnect,
   }) {
     super()
     this.#ownDeviceId = deviceId
@@ -111,9 +105,7 @@ export class MemberApi extends TypedEmitter {
     this.#projectKey = projectKey
     this.#rpc = rpc
     this.#getReplicationStream = getReplicationStream
-    this.#isConnectedToServer = isConnectedToServer
     this.#waitForInitialSyncWithPeer = waitForInitialSyncWithPeer
-    this.#waitForSyncAndDisconnect = waitForSyncAndDisconnect
     this.#dataTypes = dataTypes
   }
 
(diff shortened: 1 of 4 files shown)


achou11:
Haven't looked at the PR just yet but answering where I can...

---

> Should I initialize a replication stream if one doesn't exist instead of erroring?

I think so? From what I remember, we have 2 kinds of sync: "initial" and "data" sync. In theory, each one could be enabled/disabled at will, but in practice I think we always have the "initial" sync enabled and allow the "data" sync to be toggled (see `SyncApi.start()` and `SyncApi.stop()`). Removal of a remote archive from a project affects the `auth` namespace, which falls under the "initial" sync. If we don't have anything exposed to the client in terms of starting/stopping the "initial" sync, not sure what actions the client could take if you throw an error when attempting to remove a server.

Apologies if this isn't relevant to your question! "replicating" feels like it could mean a bunch of different things and this was the thing that first came to mind.

--- 

> Right now this only allows removing while we're replicating.

Does this detail from the original issue conflict with this implementation detail? 

> should handle the edge-case of the coordinator writing the new "blocked" role for the server, but not having a connection to the server at that time, or loosing the connection before syncing. The device that removes the server should keep trying to connect until it confirms that the blocked role record has synced to the server. This retry should persist across app restarts.

Could be tangential but just wanted to check.

---

> Is removing the server by the base URL the easiest for the front end? Should we use the deviceId or some other identifier instead?

I _think_ `deviceId` would be slightly easier for the frontend since remote archives are project members. The general sequence from the frontend would be:

  1. Get all members of the project
  2. Filter out the non-archive members
  3. Select one of the remaining members and call `project.$member.removeServerPeer(member.deviceId)`
  
To me, it'd be nice if we could get closer to something like `MemberApi.remove(deviceId)`, where it only actually does something for remote archives for now, but could eventually be used for non-archive members too (if that's a future need we have).

---

> Are there any events we should be adding to let the front end know the server is removed, or is the role update for the server peer enough?

I think the role update for the server peer is sufficient for our needs. After removing the server, we'd do a refetch of the project members and see that the role has been updated that way.

RangerMauve:
Very useful feedback, ty. Going to do some refactoring based on it. I agree `.remove()` could be handy down the line.

Regarding the edge case, this will happen in a future PR. Going to handle the online case first in order to unblock folks, then add resilience to remove while offline later

RangerMauve:
Simplified the code a bunch.

- fixed `MembersAPI.getById` not supporting the self hosted server device info
- We always establish a connection on remove and wait for initial sync, bypassing the rest of the sync API.
- Added a test to make sure the server knows that it is blocked after we sync

src/member-api.js
```diff
@@ -312,6 +317,46 @@ export class MemberApi extends TypedEmitter {
...
+   * @param {object} [options]
+   * @param {boolean} [options.dangerouslyAllowInsecureConnections] Allow insecure network connections. Should only be used in tests.
+   * @returns {Promise<void>}
+   */
+  async removeServerPeer(
+    serverDeviceId,
+    { dangerouslyAllowInsecureConnections = false } = {}
+  ) {
+    // Get device ID for URL
+    // Prase through URL to ensure end pathname if missing
```
achou11, line 333: parse?
```suggestion
    // Parse through URL to ensure end pathname if missing
```

achou11 (approved):

PR #1018, opened 2025-04-17
feat: update @comapeo/schema to 1.5.0 to introduce updated project settings schema

### LINKED ISSUE #1014: Add Properties to `ProjectSettings`
## Description
In anticipation for [Multi Projects](https://www.notion.so/digidem/Multiple-projects-ce7d9913220342ed968311ac9e8a523b?pvs=4#1881b08162d581d09885dd956a39afd4) the following properties need to be added to ProjectSettings:

- `projectColor:string` - this could be an optional field. In an ideal world this would be a required field, and we would add the default color to those that do not already have this field. But if that is too difficult we can make this optional, and the front end can add the color in the app. This will always be used and required, so its not ideal, but overall pretty minor for the front end to do
- `projectDescription?:string`. This should be optional as we do not want to force the user to write a project Description if they do not want to
- ~`projectIcon?:{
        mimeType: "image/png";
        size: Size;
        pixelDensity: 1 | 2 | 3;
        blobVersionId: BlobVersionId;
    } | {
        size: Size;
        mimeType: "image/svg+xml";
        blobVersionId: BlobVersionId;
    }` - This should also be optional as we are currently not using it, but is going to be used in the future. I'm basing this typing off the `Icon` typing from schema~ I am realizing that we need to know how project icons will be added before we know the typing. There is potential that they will be simple react components, in which case we might just need to save them as strings.

## Tasks

- [ ]

achou11:
Closes #1014 

Surface area feels kind of large at the moment - I'm open to breaking it up if it makes sense to (with some guidance).

Summary of changes:

- Adds a database migration for the client database schema to add support for the `projectColor` and `projectDescription` fields as part of the `projectInfo` column.

- Updates the invite RPC message to add `projectColor` and `projectDescription` fields (both optional).
    
    - ~**Question**: should the `projectColor` be a required field? `projectName` is a required field but not sure if the color should follow suite. Based on UX needs, maybe it makes sense to, but can also imagine it being okay to omit and letting the app handle using a fallback.~ EDIT: at the RPC level, it has to be optional for compat reasons. at the app api level, we decided to leave it as optional and let the frontend handle it.
    
- Updates the `MapeoManager` to support creating a project with a color and description specified (both optional).

- Updates the `MapeoManager` to support the color and description fields when adding a project (mostly relevant for when accepting an invite).

- Updates the `MemberApi` to include the project color and description when sending an invite (both optional).

- Updates affected tests. Didn't put too much emphasis on trying to introduce new tests, but open to doing so with some guidance on what's considered worthy of adding.

### DIFF
--- a/proto/rpc.proto
+++ b/proto/rpc.proto
@@ -9,6 +9,9 @@ message Invite {
   optional string roleName = 4;
   optional string roleDescription = 5;
   string invitorName = 6;
+  // TODO: Should this be required?
+  optional string projectColor = 7;
+  optional string projectDescription = 8;
 }
 
 message InviteCancel {
--- a/proto/rpc.proto
+++ b/proto/rpc.proto
@@ -9,7 +9,6 @@ message Invite {
   optional string roleName = 4;
   optional string roleDescription = 5;
   string invitorName = 6;
-  // TODO: Should this be required?
   optional string projectColor = 7;
   optional string projectDescription = 8;
 }
--- a/src/member-api.js
+++ b/src/member-api.js
@@ -155,8 +155,6 @@ export class MemberApi extends TypedEmitter {
       const projectInviteId = projectKeyToProjectInviteId(this.#projectKey)
       const project = await this.#dataTypes.project.getByDocId(projectId)
       const projectName = project.name
-
-      // TODO: Require projectColor to be set as well?
       assert(projectName, 'Project must have a name to invite people')
 
       const projectColor = project.projectColor
--- a/src/member-api.js
+++ b/src/member-api.js
@@ -155,7 +155,12 @@ export class MemberApi extends TypedEmitter {
       const projectInviteId = projectKeyToProjectInviteId(this.#projectKey)
       const project = await this.#dataTypes.project.getByDocId(projectId)
       const projectName = project.name
+
+      // TODO: Require projectColor to be set as well?
       assert(projectName, 'Project must have a name to invite people')
+
+      const projectColor = project.projectColor
+      const projectDescription = project.projectDescription
 
       abortSignal.throwIfAborted()
 
@@ -163,6 +168,8 @@ export class MemberApi extends TypedEmitter {
         inviteId,
         projectInviteId,
         projectName,
+        projectColor,
+        projectDescription,
         roleName,
         roleDescription,
         invitorName,
--- a/test-e2e/members.js
+++ b/test-e2e/members.js
@@ -116,6 +116,27 @@ test('getting yourself after adding project (but not yet synced)', async (t) => 
   )
 })
 
+test('cannot invite when project is missing critical details', async (t) => {
+  const managers = await createManagers(2, t)
+  const [invitor, invitee] = managers
+  const disconnectPeers = connectPeers(managers)
+  t.after(disconnectPeers)
+
+  // Note: no name for created project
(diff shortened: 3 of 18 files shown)


proto/rpc.proto
```diff
@@ -9,6 +9,9 @@ message Invite {
   optional string roleName = 4;
   optional string roleDescription = 5;
   string invitorName = 6;
+  // TODO: Should this be required?
+  optional string projectColor = 7;
```
achou11, line 13: highlighting this as related to my question about making this a required field

src/member-api.js
```diff
@@ -155,14 +155,21 @@ export class MemberApi extends TypedEmitter {
       const projectInviteId = projectKeyToProjectInviteId(this.#projectKey)
       const project = await this.#dataTypes.project.getByDocId(projectId)
       const projectName = project.name
+
+      // TODO: Require projectColor to be set as well?
```
achou11, line 159: highlighting as it's related to my question about whether to make the `projectColor` field in an invite required or not

test-e2e/members.js
```diff
@@ -116,6 +116,27 @@ test('getting yourself after adding project (but not yet synced)', async (t) =>
   )
 })
 
+test('cannot invite when project is missing critical details', async (t) => {
```
achou11, line 119: noticed that this test was notably missing. not strictly necessary to add this as part of this PR, but figured it could be if we decide to make the project color a required field for invites as well.

RangerMauve (changes requested):
Looks solid overall, just some comments regarding having the fields be `required`

proto/rpc.proto
achou11: Makes sense! Kind of hesitant to commit to using a default value in core though - is that because of the app versioning mismatch issue?

src/member-api.js
achou11: Similar to other comments, I feel slightly uncomfortable with committing to a default color to use at the core level.

UX tradeoffs I'm thinking about in regards to clients with pre-existing projects that upgrade to this new API:

1. If we require color to be set, they'll receive an error about missing a color despite having been able to send invites before. The UI would need to expose a way to set this color (think that's already the plan).

2. Otherwise, we'll implicitly use the default color when sending new invites (until they explicitly set the color).

(1) feels more correct to me despite it seeming more inconvenient, but curious to get additional thoughts.

achou11: But yeah I think making the color required here generally makes sense

RangerMauve: IMO it'd be better to have it optional for backwards compat and have the UI set a default if one is lacking. 🤷 Eventually it won't matter once everyone is updated.

RangerMauve:
Talked more with Andrew and we think it'd be better to have the fields be optional up the whole stack. It makes the code cleaner and accounts for backwards compat in a straightforward way. Also at the RPC level we pretty much *must* have the fields optional for backwards compat

src/member-api.js
achou11: based on discussion, decided to not require project color here

proto/rpc.proto
achou11: based on discussion, this has to be optional for compat reasons. todo comment is no longer necessary

RangerMauve (approved):

PR #1023, opened 2025-04-24
Fix/ensure invite config sync

### LINKED ISSUE #1022: Invites: Account for initial sync failing
## Description

Sometimes inviting fails in the middle of syncing the project after it's been accepted and the invitee's role was set. We should account for this failure.

## Tasks

- [x] Have inviter wait for initial sync before marking the invite as complete
- [x] On failure, mark the role as LEFT

RangerMauve:
fix #1022 

- inviter waits for initial sync before considering the accepted invite to be a success
- inviter sets the member as "left" if the initial sync failed
- inviters can now safely try again and both sides will know the invite failed
- updated unexpected disconnect test since it's more reliable now

I'm actually pretty happy with this outcome! This doesn't auto-retry but it makes sure the inviter knows it fails and resets the member to a state where they can try inviting them again.

Sadly I wasn't as aware of how roles worked two weeks ago and wasn't aware this fix was an option

### DIFF
--- a/src/errors.js
+++ b/src/errors.js
@@ -31,7 +31,7 @@ export class InviteAbortedError extends Error {
 
 export class ProjectDetailsSendFail extends Error {
   /** @param {string} [message] */
-  constructor(message = 'Project Details Failed to send') {
+  constructor(message = 'Project details failed to send') {
     super(message)
     this.name = 'ProjectDetailsSendFail'
   }
--- a/src/errors.js
+++ b/src/errors.js
@@ -29,35 +29,35 @@ export class InviteAbortedError extends Error {
   }
 }
 
-export class ProjectDetailsSendFail extends Error {
+export class ProjectDetailsSendFailError extends Error {
   /** @param {string} [message] */
   constructor(message = 'Project details failed to send') {
     super(message)
-    this.name = 'ProjectDetailsSendFail'
+    this.name = 'ProjectDetailsSendFailError'
   }
 }
 
-export class InviteInitialSyncFail extends Error {
+export class InviteInitialSyncFailError extends Error {
   /** @param {string} [message] */
   constructor(message = 'Failed to sync config for invite') {
     super(message)
-    this.name = 'InviteInitialSyncFail'
+    this.name = 'InviteInitialSyncFailError'
   }
 }
 
-export class RPCDisconnectBeforeSending extends Error {
+export class RPCDisconnectBeforeSendingError extends Error {
   /** @param {string} [message] */
   constructor(message = 'RPC disconnected before sending') {
     super(message)
-    this.name = 'RPCDisconnectBeforeSending'
+    this.name = 'RPCDisconnectBeforeSendingError'
   }
 }
 
-export class RPCDisconnectBeforeAck extends Error {
+export class RPCDisconnectBeforeAckError extends Error {
   /** @param {string} [message] */
   constructor(message = 'RPC disconnected before receiving acknowledgement') {
     super(message)
-    this.name = 'RPCDisconnectBeforeAck'
+    this.name = 'RPCDisconnectBeforeAckError'
   }
 }
 
--- a/src/errors.js
+++ b/src/errors.js
@@ -55,7 +55,7 @@ export class RPCDisconnectBeforeSending extends Error {
 
 export class RPCDisconnectBeforeAck extends Error {
   /** @param {string} [message] */
-  constructor(message = 'RPC disconnected before recieving acknowledgement') {
+  constructor(message = 'RPC disconnected before receiving acknowledgement') {
     super(message)
     this.name = 'RPCDisconnectBeforeAck'
   }
--- a/src/errors.js
+++ b/src/errors.js
@@ -29,6 +29,38 @@ export class InviteAbortedError extends Error {
   }
 }
 
+export class ProjectDetailsSendFail extends Error {
+  /** @param {string} [message] */
+  constructor(message = 'Project Details Failed to send') {
+    super(message)
+    this.name = 'ProjectDetailsSendFail'
+  }
+}
+
+export class InviteInitialSyncFail extends Error {
+  /** @param {string} [message] */
+  constructor(message = 'Failed to sync config for invite') {
+    super(message)
+    this.name = 'InviteInitialSyncFail'
+  }
+}
+
+export class RPCDisconnectBeforeSending extends Error {
+  /** @param {string} [message] */
+  constructor(message = 'RPC disconnected before sending') {
+    super(message)
+    this.name = 'RPCDisconnectBeforeSending'
+  }
+}
+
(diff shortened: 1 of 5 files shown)


src/errors.js
```diff
@@ -29,6 +29,38 @@ export class InviteAbortedError extends Error {
   }
 }
 
+export class ProjectDetailsSendFail extends Error {
+  /** @param {string} [message] */
+  constructor(message = 'Project Details Failed to send') {
```
achou11, line 34: nitpicking but seems like sentence-casing is reasonable for this?
```suggestion
  constructor(message = 'Project details failed to send') {
```

src/errors.js
```diff
@@ -29,6 +29,38 @@ export class InviteAbortedError extends Error {
...
+  /** @param {string} [message] */
+  constructor(message = 'RPC disconnected before sending') {
+    super(message)
+    this.name = 'RPCDisconnectBeforeSending'
+  }
+}
+
+export class RPCDisconnectBeforeAck extends Error {
+  /** @param {string} [message] */
+  constructor(message = 'RPC disconnected before recieving acknowledgement') {
```
achou11, line 58:
```suggestion
  constructor(message = 'RPC disconnected before receiving acknowledgement') {
```

src/errors.js
achou11, line 1: is it worth suffixing with names of these custom classes with `Error`? Think I'm more used to the names of custom error classes being very explicit about that.

test-e2e/manager-invite.js
achou11, line 1: is the todo comment still relevant?

test-e2e/manager-invite.js
```diff
@@ -637,48 +637,41 @@ test('disconnect before invite accept', async (t) => {
...
-      'accepting invite times out when project details are not received'
-    )
-    await disconnectPeers()
-    // Run down the timeout waiting for project details
-    clock.runAll()
-    await Promise.all([assertInviteRejectsPromise, assertAcceptRejectsPromise])
-  }
-)
+// peers disconnect, so can't expect specific errors.
+test('disconnect before sending project join details', async (t) => {
```
achou11, line 641: if it's not already tested elsewhere, I think it would be helpful to extend this test to make sure that re-inviting can work after the failed attempt simulated here

src/member-api.js
```diff
@@ -200,16 +205,28 @@ export class MemberApi extends TypedEmitter {
...
 
           // Only add after we know they got the details
           // Otherwise the joiner will be stuck unable to join
           await this.#roles.assignRole(deviceId, roleId)
 
+          try {
+            await this.#waitForInitialSyncWithPeer(deviceId, abortSignal)
+          } catch {
+            // Mark them as "left" so we can retry the flow
+            await this.#roles.assignRole(deviceId, LEFT_ROLE_ID)
```
achou11, line 226: just noting that I really don't have the best idea about whether this introduces any unintended side effects. think it should be okay but just wanted to be honest about it 😄

achou11 (changes requested):
Couple of small suggestions and one request to make sure a key behavior is represented in the affected test.

test-e2e/manager-invite.js
RangerMauve: Good idea!

test-e2e/manager-invite.js
RangerMauve: Yeah we still can't absolutely guarantee when the disconnect happens in the cycle, sadly.

test-e2e/manager-invite.js
RangerMauve: Good catch, we got an error saying we can't assign the LEFT role to another device.

RangerMauve: Yeah looks like I need a new role

src/member-api.js
```diff
@@ -511,6 +536,19 @@ export class MemberApi extends TypedEmitter {
...
+   * @returns {Promise<Array<MemberInfo>>}
+   */
+  async getActive() {
+    const members = await this.getMany()
+    return members.filter(({ role }) =>
+      [MEMBER_ROLE_ID, COORDINATOR_ROLE_ID, CREATOR_ROLE_ID].includes(
+        role.roleId
+      )
+    )
+  }
```
achou11, line 550: I'd be in favor of adding this if it actually narrowed the return type, but it doesn't due to existing TS limitations as far as I know. 

Is there a strong need to introduce this right now? Maybe open a separate PR for this specifically to consider its addition as opposed to this one.

achou11 (changes requested):
Some small changes needed before merging. Also not sure about introducing `MemberApi.getActive()` as part of this PR

achou11:
Also just to check, the existing `No Role` role can't be used instead?

>  This is the role assumed for a device when no role record can be found. This
>  can happen when an invited device did not manage to sync with the device that
> invited them, and they then try to sync with someone else. We want them to be
> able to sync the auth and config store, because that way they may be able to
> receive their role record, and they can get the project config so that they
> can start collecting data.

RangerMauve:
So, `no_role` doesn't work well here for a few reasons:

- it's not a role that can be assigned to other members (there's a check)
- it signals they should be able to sync, whereas here we want to mark them as not being succesfully added

I initially tried going the no_role route but it would have required changing a bunch of code related to how it is validated which IMO is a scarier change than a new role.

src/member-api.js
achou11: > Why would narrowing the type be necessary for this to be useful?

I don't think it's necessary but it would make it less confusing/misleading when using with TypeScript. Without the narrowing, the consuming code still has to add checks to consider every role even though usage of the method implies that it's filtered. Basically the question that would come up when using this is "Why do I have to make sure that the member's role isn't blocked/left/failed if I'm using data that's supposed to omit those?"

```tsx
function Example() {
  // 🟠 Theoretical hook that is implemented in core-react that wraps `MemberApi.getActive()`
  const { data: activeMembers } = useActiveMembers()
  
  const activeMember = activeMembers[0]

  // 🟠 This kind of check would be needed to appease TypeScript but it feels odd because the source of the data communicates something to the contrary.
  if (activeMember.role.roleId === BLOCKED_ROLE_ID) {
    // ...
  }
}
```

To clarify, I think having something like this is generally useful. The thing to note is that additions like these immediately become part of the public API due to how we consume it in the apps so just want to be thoughtful about these kinds of additions when possible. The suggestion to move as a separate PR is to provide isolation between a bug fix and a new tangential feature being added in this PR.

achou11: Tangential but worth referencing https://github.com/digidem/comapeo-core/issues/431, which alluded to ideas around how to get a filtered list of members in a project. Might be able to just close that issue if this is added!

achou11 (approved):
Changes strictly related to the fix lgtm! Still have some reservations about the addition to the MemberApi but would consider it non-blocking at this point.

PR #1029, opened 2025-05-05
feat: export GeoJSON stream

RangerMauve:
- [x] Export observations/trails as GeoJSON stream
- [x] Test bunch of observations
- [x] Test trail with bunch of observations
- [x] Test bunch of trails with bunch of observations
- [x] Export to file from a path
- [x] Export zip stream with GeoJSON + attachments
- [x] Export zip stream to file
- [ ] Clean up files if there's an error during export?


RangerMauve:
Probs gonna use this for the zip stream generation: https://www.npmjs.com/package/async-zip-stream

Or might use zip-stream directly with a promise wrapper

test-e2e/project-export.js
```diff
@@ -0,0 +1,361 @@
...
+const DEFAULT_OBSERVATIONS = 2
+const DEFAULT_TRACKS = 2
+const OBSERVATIONS_PER_TRACK = 2
+
+const BLOB_FIXTURES = fileURLToPath(
+  new URL('../test/fixtures/blob-api/', import.meta.url)
+)
+
+test('Project export empty GeoJSON to stream', async () => {
+  const manager = await setupManager()
```
achou11, line 30: no await needed here (and for other tests in this file using same setup)
```suggestion
  const manager = setupManager()
```

test-e2e/project-export.js
```diff
@@ -0,0 +1,361 @@
...
+  assert(wholeBuffer.length > 0, 'Exported buffer not empty')
+
+  const asString = b4a.toString(wholeBuffer, 'utf-8')
+
+  const parsed = JSON.parse(asString)
+
+  return parsed
+}
+
+// TODO: Create tracks and assign them observations
```
achou11, line 251: this doesn't seem to be relevant anymore

src/mapeo-project.js
```diff
@@ -749,6 +761,389 @@ export class MapeoProject extends TypedEmitter {
...
+      }
+
+      if (missingAttachments.length) {
+        this.#l.log(`Found missing attachments during export`)
+        const missingContent = missingAttachments
+          .map((attachment) => JSON.stringify(attachment))
+          .join('\n')
+
+        await archive.entry(missingContent, {
+          name: mediaFolder + 'missing.ndjson',
```
achou11, line 1083: is the idea that this file could eventually describe other kinds of content aside from attachments? just wondering if a more explicit file name is helpful or not.

src/mapeo-project.js
```diff
@@ -749,6 +761,389 @@ export class MapeoProject extends TypedEmitter {
...
+
+    // @ts-expect-error
+    await archive.entry(geoJSONStream, { name: geoJSONFileName })
+
+    const missingAttachments = []
+    // Attachments
+    if (attachments) {
+      const mediaFolder = this.#exportPrefix('Media') + '/'
+      for (const attachment of seenAttachments.values()) {
+        const stream = await this.#tryExportBlob(attachment)
```
achou11, line 1066: for clarity, the output for an exported attachment is a singular blob associated with the variant of preference, where the order of that preference is original > preview > thumbnail?

wondering if we'll need to eventually export all variants of an attachment at some point, which accounts for situations where the blob types used between variants might be notably different for some reason (example provided in https://github.com/digidem/comapeo-core/issues/903#issuecomment-2428707408). probably introduces a lot of complexity in terms of implementation but seeing this reminded me of that scenario.

src/mapeo-project.js
```diff
@@ -749,6 +761,389 @@ export class MapeoProject extends TypedEmitter {
...
+      first = false
+      yield b4a.from(
+        `${comma}\n      ` +
+          JSON.stringify({
+            type: 'Feature',
+            properties: observation,
+            geometry: {
+              type: 'Point',
+              // TODO: Altitude?
+              coordinates: [lon, lat],
```
achou11, line 798: If `lat` or `lon` are undefined, this becomes encoded as `[null, null]`. From what I can tell, the GeoJSON spec says that this field should be an array of numbers.

Not sure if this would cause issues for whomever's consuming the export but wondering if it's something that needs to be accounted for.

achou11 (commented):
Mostly very trivial suggestions and a couple of bigger picture questions.

Would also generally recommend installing [`@types/geojson`](https://www.npmjs.com/package/@types/geojson) and using it to ensure that the data that's being generated conforms to the expected shapes.

src/mapeo-project.js
```diff
@@ -749,6 +761,389 @@ export class MapeoProject extends TypedEmitter {
...
+  }
+
+  /**
+   * @param {boolean} observations
+   * @param {boolean} tracks
+   * @returns {Promise<string>}
+   */
+  async [kGeoJSONFileName](observations, tracks) {
+    let exportType = ''
+    if (observations) exportType += 'Obsvns'
```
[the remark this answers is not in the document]
RangerMauve: This was in the spec on notion so I'd ask the design team. 😅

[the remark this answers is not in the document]
achou11: I won't question it if that's what was decided. Just seems like an odd abbreviation to me 😅

src/mapeo-project.js
RangerMauve: The spec doesn't get into it too deeply, it's just supposed to contain the missing items.

Originally it was described as `missing.txt` but IMO an ndjson file would be better for machine parsing and rendering missing attachment info. Not obvious how this could mix with existing geojson tools.

https://www.notion.so/digidem/Data-Export-mobile-fe0a747ef76e42c5a3ca32736bd4e82b

src/mapeo-project.js
RangerMauve: IMO the API surface can be extended to allow for more options once we get a requirement for it. So far the spec says this on attachment export:

> When including attachments, make a “best effort” - try for original variant first, and fallback to preview, then thumbnail, if not available.

One thing this made me notice though is I wasn't adding the variant suffix! I'll make the fix.

src/mapeo-project.js
RangerMauve: Shoot yeah the spec didn't mention what to do when lat and lon are missing. Is this actually possible in how observations are constructed?

ErikSin: If there is no `lat` or `lon`, can we just make the entire `geometry` object be `null`. `geometry` can either be of type [geometry or JSON null](https://datatracker.ietf.org/doc/html/rfc7946#section-3.2).

RangerMauve: Done!

achou11:
> Since I'm dynamically generating a stream of JSON text I don't think the typescript package would help. I suppose I could include a JSON Schema to validate against in the tests though?

yeah it's mostly for the sake of catching potential type issues at development time. for example, the potential issue around missing lat/lon when you're creating the point feature could've been caught by using the types they provide.

achou11 (approved):

src/mapeo-project.js
```diff
@@ -749,6 +761,410 @@ export class MapeoProject extends TypedEmitter {
...
+      let longitude = lon
+      let altitude = null
+      const position = observation?.metadata?.position?.coords
+      if (position) {
+        latitude = position.latitude
+        longitude = position.longitude
+        if (position.altitude !== undefined) {
+          altitude = position.altitude
+        }
+      }
```
gmaclennan, line 799: The lat and lon properties of an observation are the source of truth, not the position. The main situation where a different might arise is where a user has manually entered the coordinates for an observation. The position record is useful for data analysis and validation, to check any discrepancies between the location recorded by a user and the GPS data from the phone at the time. I think it's reasonable to use the position data as a fallback if lat and lon are null.

src/mapeo-project.js
```diff
@@ -749,6 +761,410 @@ export class MapeoProject extends TypedEmitter {
...
+
+      if (missingAttachments.length) {
+        this.#l.log(`Found missing attachments during export`)
+        const missingContent = missingAttachments
+          .map((attachment) => JSON.stringify(attachment))
+          .join('\n')
+
+        await archive.entry(missingContent, {
+          name: mediaFolder + 'missing.ndjson',
```
gmaclennan, line 1104: We need to validate this with some user research, but I think in many cases a human-readable file is going to be most useful, because the user story for export is someone without programming skills exporting data to use in GIS tools or another system, and they will want to review what data is missing from the export to check whether they need to coordinate users to sync data to get all the missing data. We maybe need further work to identify what information is needed in this list of missing data that will help a user identify what needs to be done to get the data.

src/mapeo-project.js
```diff
@@ -749,6 +761,410 @@ export class MapeoProject extends TypedEmitter {
...
+  }
+
+  /**
+   * @param {Attachment} attachment
+   * @returns {Promise<null | BlobId>}
+   */
+  async #tryGetBlobId(attachment) {
+    // Audio must not have variants
+    for (const variant of VARIANT_EXPORT_ORDER) {
+      const blobId = buildBlobId(attachment, variant)
```
gmaclennan, line 1044: I believe `buildBlobId` will throw if an unknown attachment type is encountered, and this is not caught anywhere and will lead to a failed export. We should export unknown attachment types without error. The user story for this is a new version of that app adds a new attachment type, syncs with a user with an older version of the app, and the user with the older version exports data. They would expect to get all data, including the unknown attachment types, in their export.

src/mapeo-project.js
```diff
@@ -749,6 +761,410 @@ export class MapeoProject extends TypedEmitter {
...
+      yield b4a.from(
+        `${comma}\n      ` +
+          JSON.stringify({
+            type: 'Feature',
+            properties: track,
+            geometry: {
+              type: 'LineString',
+              coordinates,
+            },
+          }) +
```
gmaclennan, line 866: It's unfortunately that there is no standard JSON format for storing timestamps or other event data for each position in a track. It might be useful to add as a possible follow-up feature storing the start and end datetime for a track, since we can't export the timestamps for each position. The only export format that supports this I think is GPX (an XML format). I found this old discussion https://gis.stackexchange.com/questions/162081/representing-time-in-geojson and I remember other attempts to address this in the past, but none came to anything.

src/mapeo-project.js
```diff
@@ -749,6 +761,410 @@ export class MapeoProject extends TypedEmitter {
...
+        hadTracks = true
+        yield chunk
+      }
+    }
+
+    if (observations) {
+      if (hadTracks) {
+        yield b4a.from(',')
+      }
+      const rows = await this.#dataTypes.observation.getMany({ lang })
```
gmaclennan, line 922: Can we put a pin in a possible follow up to expose an iterable getMany from bettersqlite3, to avoid needing to load all data into memory as part of the export.

src/mapeo-project.js
```diff
@@ -749,6 +761,410 @@ export class MapeoProject extends TypedEmitter {
...
+
+  /**
+   * @param {Object} [options={}]
+   * @param {boolean} [options.observations=true] Whether observations should be exported
+   * @param {boolean} [options.tracks=true] Whether all tracks and their observations should be exported
+   * @param {SeenAttachments} [options.seenAttachments]
+   * @param {string} [options.lang]
+   * @returns {AsyncIterable<Buffer | Uint8Array>}
+   */
+  async *#exportGeoJSONIterator({
```
gmaclennan, line 892: It would be worth consulting users once this feature is released whether it is complicated mixing tracks and observations in the same geojson file. From my knowledge when importing into GIS software, mixing geometry types in a file can sometimes make data processing more complicated.

gmaclennan (commented):
Looks good overall, but I think the question around errors from unknown attachment types needs to be addressed. Most other comments are for follow-up.

A couple of design comments: I wonder whether responsibility for the filename should lie with the API consumer, who can pass a filepath rather than a folder? Also wondering about how we expose missing attachments which might need some user research.

For the tests, I wonder why we are exposing some internal methods via symbol properties? Our e2e tests should only rely on public methods IMO, and internal methods should only be used as a temporary measure if tests are needed before a feature is complete. I'm not sure we need the symbol method wrappers at all?

Finally with testing I think it's important to test with large numbers of observations and a lot of attachment data. We often come into issues when tests only use small sample data sets. I also think it's important that tests include tests for missing variants, since this is very common in real-life use and we should handle it correctly.

src/mapeo-project.js
```diff
@@ -749,6 +761,410 @@ export class MapeoProject extends TypedEmitter {
...
+
+        // @ts-expect-error
+        await archive.entry(stream, { name })
+      }
+
+      if (missingAttachments.length) {
+        this.#l.log(`Found missing attachments during export`)
+        const missingContent = missingAttachments
+          .map((attachment) => JSON.stringify(attachment))
+          .join('\n')
```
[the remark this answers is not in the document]
RangerMauve: At the moment it only registers attachments that are entirely missing.

I misread the requirement as "if the attachment is missing and all variants are missing then add attachment to missing.txt". Should it instead be "for each variant that is missing add the variant type and attachment info to missing.txt"? or should it be more along the lines of "If the original is missing, add its information to missing.txt, for each variant that is missing also add it to missing.txt"?

src/mapeo-project.js
RangerMauve: The current TS types require a variant to be specified in `SUPPORTED_BLOB_VARIANTS`. Should I make it possible to bypass this inside the blob store API?

IMO it's not the end of the world if an old app skips data it doesn't understand.

cc @achou11 on ideas for how to make the TS types happy with unknown variants.

```javascript
const SUPPORTED_BLOB_VARIANTS = /** @type {const} */ ({
  photo: ['original', 'preview', 'thumbnail'],
  audio: ['original'],
  video: ['original'],
})
```

src/mapeo-project.js
RangerMauve: Yeah my original implementation was to do a stream of the observations but I found out that sqlite doesn't support iterating over a query. https://orm.drizzle.team/docs/select#iterator

It's whack that streaming isn't a first class feature of all the adapters 😅

We could make one by using `SKIP`/`LIMIT` to do pagination. Thoughts on page size? I was thing 512 or 1024.

src/mapeo-project.js
gmaclennan: Good point and agreed. I think given that we have no immediate plans to add new blob types or variants, it's ok to skip data that is not understood. I think the fix that is needed to stop unknown data throwing, and instead ignore it.

src/mapeo-project.js
RangerMauve: Tracked https://github.com/digidem/comapeo-core/issues/1038

src/mapeo-project.js
RangerMauve: From the UI you choose only one or the other right now.

src/mapeo-project.js
gmaclennan: Ah interesting. It's supported by bettersqlite3, but yeah it does seem that drizzle doesn't support this yet. I don't think this is an urgent issue, since we load everything into memory for reads anyway. Let's punt this and address it when we add pagination to reads in the API, and then we can use that pagination for export too.

RangerMauve:
>  I wonder whether responsibility for the filename should lie with the API consumer, who can pass a filepath rather than a folder?

The main reason I have core generating the file name is that I need to reuse filename generation logic for the geojson file within the zip file, per the notion doc. I think this has been working well for @ErikSin so far from front end DX

> Also wondering about how we expose missing attachments which might need some user research.

Yeah IMO this function is a bit under specced in the notion doc. Happy to implement whatever would be most useful. My gut feel is to release *something* and then release an update with the better format once we get feedback.

> I'm not sure we need the symbol method wrappers at all?

True, I can make temporary dirs for the tests and keep those methods private.

src/mapeo-project.js
RangerMauve: Added catch for this

Issue #1037, opened 2025-05-21
Correct export of attachments


ErikSin:
## Description
When creating a zip file with the observation geoJSON and the associated attachements, the attachment are not saving with a proper file extension. As well the are being saved in a folder named `[object Promise]/original`.

This image shows the file structure and the associated attachments. The second file is an audio file and the other ones are images. 

<img width="699" alt="Image" src="https://github.com/user-attachments/assets/9341e067-93f2-403d-a0b0-eaed7814af1f" />

## Tasks

- [ ]

RangerMauve:
Done! Will be part of the next release of core

PR #1039, opened 2025-05-22
Fix/export extensions

RangerMauve:
Fix for #1037 as well as improvements fro Gregor's comments in #1029

### DIFF
--- a/test-e2e/project-export.js
+++ b/test-e2e/project-export.js
@@ -13,7 +13,6 @@ import { fileURLToPath } from 'node:url'
 import { createReadStream } from 'node:fs'
 
 import { MapeoManager } from '../src/mapeo-manager.js'
-import { blobMetadata } from '../test//helpers/blob-store.js'
 
 /** @import { Readable } from 'streamx' */
 


achou11 (approved):
Couple of questions and some non-blocking suggestions, but otherwise changes lgtm!

src/mapeo-project.js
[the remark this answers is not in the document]
achou11: you may need to install `@types/geojson` as an explicit dev dep (and use that for the type declaration in the example)

src/mapeo-project.js
```diff
@@ -786,40 +789,43 @@ export class MapeoProject extends TypedEmitter {
...
+            typeof altitude === 'number'
+              ? [metadataCoords.longitude, metadataCoords.latitude, altitude]
+              : [metadataCoords.longitude, metadataCoords.latitude]
         }
       }
 
-      const coordinates = [longitude, latitude]
-      if (typeof altitude === 'number') {
-        coordinates.push(altitude)
+      /** @type {import('geojson').Feature<import('geojson').Point | null>} */
```
achou11, line 815: mentioned [here](https://github.com/digidem/comapeo-core/pull/1039#discussion_r2127215297) but could've been lost, but may need to use `@types/geojson` and have it installed as an explicit dev dep.

```suggestion
      /** @type {import('@types/geojson').Feature<import('@types/geojson').Point | null>} */
```

src/mapeo-project.js
```diff
@@ -1036,15 +1042,34 @@ export class MapeoProject extends TypedEmitter {
...
+        const entry = await this.#blobStore.entry(blobId)
+        if (!entry) continue
+        const metadata = entry.value.metadata
+        if (!metadata || typeof metadata !== 'object') continue
+        if (
+          !('mimeType' in metadata && typeof metadata.mimeType === 'string')
+        ) {
+          this.#l.log('Blob missing mime type', blobId, entry)
+          continue
+        }
```
gmaclennan, line 1061: I don't think we should not export attachments without metadata. This field is optional, and early versions of comapeo might not have metadata or mimetype, and I think it's best to lean towards getting all data out.

src/mapeo-project.js
```diff
@@ -1079,16 +1104,29 @@ export class MapeoProject extends TypedEmitter {
...
+          missingAttachments.push(attachment)
+          continue
+        }
+
+        const { blobId, mimeType } = ref
+        const extension = mime.getExtension(mimeType)
+
+        if (!extension) {
           missingAttachments.push(attachment)
+          this.#l.log('Got unknown mime type in attachment blob', attachment)
```
gmaclennan, line 1120: similarly here I kind of think we should still include the attachment even if we don't know the mimetype.

gmaclennan (approved):
Looks ok to me, but it would be I kind of think we should export attachments even if we don't know their mimetype or they don't have metadata, or if it's an unknown mimetype. Theoretically there should not be blobs in the database without a mimetype, but if there are, we would want to get them out to look at them probably. I could be convinced otherwise though!

PR #1052, opened 2025-06-17
Exclude left projects from project list

RangerMauve:
fixes https://github.com/digidem/comapeo-mobile/issues/1235

### DIFF
--- a/test-e2e/project-leave.js
+++ b/test-e2e/project-leave.js
@@ -219,6 +219,10 @@ test('Member can leave project if creator exists', async (t) => {
     ROLES[LEFT_ROLE_ID],
     'creator can still retrieve info about member who left'
   )
+
+  const list = await member.listProjects()
+
+  assert.equal(list.length, 0, 'member has no projects after leaving')
 })
 
 test('Data access after leaving project', async (t) => {


src/mapeo-manager.js
```diff
@@ -658,6 +663,26 @@ export class MapeoManager extends TypedEmitter {
         )
       }
 
+      /** @type {"projectSettings"} */
+      const schemaName = 'projectSettings'
+
+      /** at-type import('@comapeo/schema').ProjectSettingsValue */
+      const settingsDoc = {
+        schemaName,
```
achou11, line 671: accounting for the mistyped type annotation, can adjust this a bit:

```suggestion
      const settingsDoc = {
        schemaName: /** @type {const} */ ('projectSettings'),
```

src/mapeo-manager.js
```diff
@@ -89,6 +89,9 @@ export const DEFAULT_FALLBACK_MAP_FILE_PATH = require.resolve(
 export const DEFAULT_ONLINE_STYLE_URL =
   'https://demotiles.maplibre.org/style.json'
 
+// Oldest possible time, ensure it gets overwritten with any updates
+export const UNIX_EPOCH_DATE = new Date(0).toISOString()
```
achou11, line 93: exporting this doesn't seem necessary

```suggestion
const UNIX_EPOCH_DATE = new Date(0).toISOString()
```

achou11 (changes requested):
Might be worth adding a note about the nature of the changes - it took some questioning from me elsewhere to get a fuller idea of how they relate to the issue being addressed.

Aside from that, pointed out a couple of very small changes that probably should be done

gmaclennan:
I pushed the commit with the type fix so that I can run CI tests again to see the log output (the tests had previously run long enough ago that I could not see logs)

gmaclennan:
In terms of the failing test, I think the issue is that now `addProject` could complete before sync is done, because `#waitForInitialSync` no longer waits for projectSettings to sync. This will be rare, but I think in some tests it will mean that subsequent actions happen before that initial sync is complete, which will cause errors. This is probably an issue only with tests, because it is unlikely that this would be replicated in a real-world situation, however knowing our luck it could be. Do we actually want to not wait for project settings in the `#waitForInitialSync`? We are still writing that data and we want it to sync in ideal conditions. When project settings are not written due to an error, would a timeout from sync be an ok result?

RangerMauve:
TY for digging into it, I'll figure out another signal for the isProjectSettingSynced flag, probably from the updatedAt timestamp of the projectSettings being set to the Unix 0.

gmaclennan (commented):
I think maybe we need to change our handling of an existing project now, because it is more likely that a failed invite will leave an existing project around, and I think then you would be stuck, because it would "half exist" and you would be unable to add it again. Could you just check the logic on that.

PR #1078, opened 2025-07-29
fix: No outgoing connections to existing connections

### LINKED ISSUE #1076: Remove redundant existing peer connection check
Currently `localDiscovery.connectPeer()` has a [check for existing connections](https://github.com/digidem/comapeo-core/blob/86f7731d524abd034adca1021c4d3d85fa0e8740/src/discovery/local-discovery.js#L110-L113). However it uses the mDNS service name, which is a random string, and it checks it against a map which uses deviceIds as the key, so it will just not work. This check is not necessary anyway because of the dedupe code once the noise connection is made, and if it did work it could create race conditions where one peer uses this check, and the other peer uses the deterministic dedupe once the noise connection is made, which could result in both connections being removed.

This is non-urgent, because I believe this code currently just does nothing.

RangerMauve:
Open questions:

- Should we wait for our outgoing connection to "connect" before considering it outgoing, or should we ignore any attempts to connect while we're in the process of connecting too?
- Similar for incoming connections should we destroy incoming sockets if we already have one in progress from that IP/Port?

This could affect the reconnect logic once we add it.

Closes #1076

### DIFF
--- a/src/discovery/local-discovery.js
+++ b/src/discovery/local-discovery.js
@@ -39,6 +39,10 @@ export class LocalDiscovery extends TypedEmitter {
   #server
   /** @type {Map<string, OpenedNetNoiseStream>} */
   #noiseConnections = new Map()
+  /** @type {Set<string>} */
+  #connectionsToName = new Set()
+  /** @type {Set<string>} */
+  #connectionsFromIP = new Set()
   #sm
   #log
   /** @type {(e: Error) => void} */
@@ -107,8 +111,13 @@ export class LocalDiscovery extends TypedEmitter {
   connectPeer({ address, port, name }) {
     if (this.#name === name) return
     this.#log('peer connected', name.slice(0, 7), address, port)
-    if (this.#noiseConnections.has(name)) {
+    if (this.#connectionsToName.has(name)) {
       this.#log(`Already connected to ${name.slice(0, 7)}`)
+      return
+    }
+    const ipId = `${address}:${port}`
+    if (this.#connectionsFromIP.has(ipId)) {
+      this.#log(`Already connected to ${ipId}`)
       return
     }
     const socket = net.connect({
@@ -118,6 +127,8 @@ export class LocalDiscovery extends TypedEmitter {
     })
     socket.on('error', this.#handleSocketError)
     socket.once('connect', () => {
+      this.#connectionsToName.add(name)
+      socket.once('close', () => this.#connectionsToName.delete(name))
       this.#handleTcpConnection(true, socket)
     })
   }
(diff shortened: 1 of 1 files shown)


src/discovery/local-discovery.js
```diff
@@ -141,11 +152,16 @@ export class LocalDiscovery extends TypedEmitter {
...
+    const { remoteAddress, remotePort } = socket
     if (!remoteAddress || !isPrivate(remoteAddress)) {
       socket.destroy(new Error('Invalid remoteAddress ' + remoteAddress))
       return
     }
+
+    // TODO: What if we already have an incoming connection from them?
+    const ipId = `${remoteAddress}:${remotePort}`
+    this.#connectionsFromIP.add(ipId)
+
```
gmaclennan, line 164: This won't work - the remote port is random and changes for each connection. e.g. run this and you'll see the port is different on each connection:

```js
import { once } from 'node:events'
import net from 'node:net'

const server = net.createServer()
server.listen(0, '0.0.0.0')

server.on('connection', (socket) => {
  console.log(`${socket.remoteAddress}:${socket.remotePort} connected`)
})

await once(server, 'listening')

net.connect({ port: server.address().port, host: '127.0.0.1' })
net.connect({ port: server.address().port, host: '127.0.0.1' })
```

src/discovery/local-discovery.js
```diff
@@ -107,17 +111,24 @@ export class LocalDiscovery extends TypedEmitter {
...
+      return
+    }
     const socket = net.connect({
       ...TCP_KEEP_ALIVE_OPTIONS,
       host: address,
       port,
     })
     socket.on('error', this.#handleSocketError)
     socket.once('connect', () => {
+      this.#connectionsToName.add(name)
```
gmaclennan, line 130: I think we shouldn't add this until the noise handshake and deduplication is complete, to avoid a possible race condition with the deduplication logic where one side dedupes here (first connection wins) and the other side dedupes with the dedupe code after noise stream connection (1st connection sorted by public key wins). This could happen with unusual network conditions, would be rare, but extremely hard to find and debug.

gmaclennan (changes requested):
Unfortunately there is no way to uniquely identify incoming connections (TCP connections work by connecting from a randomly assigned port) until we do the noise handshake.

For outgoing connections, there is a risk that someone could theoretically stop a client connecting by publishing a duplicate service name to mdns, but I think that would not have significant security consequences and would be hard to do, so I think ok. I think we should wait until the noise handshake and de-dupe though, to avoid a rare race condition conflict between the dedupe of outgoing based on name, and the dedupe that happens later.

RangerMauve:
Closing for now!

PR #1081, opened 2025-07-30
fix: turn off debug logging for indexing and sync

gmaclennan:
Logging these events produces a lot of output, which has a performance overhead. We need debug logging for debugging other issues, and currently we don't have a way to set log levels to selectively log this, so turning this off for now until we can filter logs by level.


RangerMauve (approved):
Yeah this will make my life a lot easier outside of the perf gains 😅

PR #1084, opened 2025-08-04
chore: add explicit type exports

gmaclennan:
These are the types currently used by
dependencies, and are currently imported from
source files, which is liable to break if we
refactor code. Dependencies should be updated to
use these exports from the main entry point.


achou11 (approved):
Nice!

Reminds me: as a separate PR, can the role IDs be exported as top-level exports to avoid doing [this](https://github.com/digidem/comapeo-mobile/blob/da7a76ede0da57557cbceaa04488c47d3970717a/src/frontend/sharedTypes/index.ts#L37-L43)? Importing from the direct source file didn't work for some reason (maybe due to missing type exports or something similar)

PR #1092, opened 2025-08-07
feat: Index docs in a worker

gmaclennan:


RangerMauve:
Why was it necessary to migrate the coresTable for this?

gmaclennan:
Not sure, I have a feeling that they are not. Easiest way to check is to comment out the migration code and see if any test fails.

gmaclennan:
This would only be noticeable in this sequence:

1. Devices A and B
2. Device A creates some data, syncs with B
3. Devices disconnect
4. Device B destroys its instance and creates a new instance with the new core manager storage
5. Device B re-indexes data
6. Does data that was synced from A still exist?

You would only notice that cores are missing (eg they haven't been persisted) if device B reindexes, because the data will be in the indexes for B. Alternatively, if B syncs with a new device C after migration, are A's cores shared (eg does C get A's data from B)

RangerMauve:
Fixed the issue with manager stop freezing. Seems sync is busted now. 😅 Will progress on it next time I have a moment

PR #1104, opened 2025-09-03
Feat/better sqlite3 11

RangerMauve:
This should fix the ci failing on node 22 with the older core dep pulling in an outdated better-sqlite3

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -21,7 +21,7 @@
         "@sindresorhus/merge-streams": "^4.0.0",
         "b4a": "^1.6.3",
         "bcp-47": "^2.1.0",
-        "better-sqlite3": "^8.7.0",
+        "better-sqlite3": "^11.10.0",
         "big-sparse-array": "^1.0.3",
         "bogon": "^1.1.0",
         "compact-encoding": "^2.12.0",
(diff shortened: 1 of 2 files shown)


achou11 (commented):
A bit unfamiliar with how override definitions work for modules that are consumed as dependencies. Does this mean that any sub dependency that's installed when running `npm i @comapeo/core` will be using the override? Seems dangerous but seems like the only direct dependency that's affected by this is `@mapeo/sqlite-indexer`, which has a [pretty relaxed peer dep declaration](https://github.com/digidem/mapeo-sqlite-indexer/blob/ebb2de1fb2cc772c1ccb47bc2f8676687acd69a7/package.json#L25) on better-sqlite3.

gmaclennan:
Can you follow-up with https://github.com/digidem/mapeo-sqlite-indexer/issues/29, so we can ensure those tests pass with v11.

RangerMauve:
The error this override fixes isn't coming from sqlite-indexer. It's from the dev dependency on [@comapeo/core@2.0.1](https://github.com/digidem/comapeo-core/actions/runs/17444262995/job/49534620104#step:3:168)

I'm not sure if the override gets used when we install core as a dependency or not but I can check. What would the issue be with core forcing an sqlite version it its dependencies?

achou11:
> What would the issue be with core forcing an sqlite version it its dependencies?

In most cases you run the risk of sub dependencies using a version of a dep that they haven't been run and tested. More of a potential issue when the override is changing the major version such as what's being done here. Luckily better-sqlite3 (and sqlite) have been pretty good about not actually introducing runtime breaking changes between major versions.

achou11 (approved):

PR #1111, opened 2025-09-08
feat: Derive createdBy and updatedBy for dataTypes

### LINKED ISSUE #136: Derive `createdBy` and `updatedBy` fields for data returned to client
### Description

This is a relative priority because our current workaround — [`project.$originalVersionIdToDeviceId()`](https://github.com/digidem/comapeo-core/blob/2e4a2933b1bccb17d1f14089ba0639be22e50745/src/mapeo-project.js#L696) — results in a large number of RPC requests, which is slowing down the app significantly due to current issues with RPC overhead. Including these fields in records in core will reduce the need for additional queries by the client.

We don't store the deviceId of the device that created or edited (updated) any record, but we do store the core ID (public key for core) where the record is written in the versionId. The `versionId` is a derived field so it can't be "hacked" by a user. The creator is tracked with an `originalVersionId` field written to each record. A "bad" client could theoretically change this, but it can be checked by tracing the record history and comparing the `versionId` or the "root" record with `originalVersionId`. It's not worth doing this validation for every read, and it's an edge-case with few practical applications for a bad actor, so we don't need to address this here, but can add a "verify database" function in the future that can check these things.

The `versionId` and `originalVersionId` can be mapped to coreIds via the @comapeo/schema `parseVersionId()` method. The coreIds can be mapped to a deviceId with `coreOwnsership.getOwner(coreId)`.

### Tasks

This task is to use core ownership records to derive the device ID of the creator and updator, and append them to each returned document/record as `createdBy` and `updatedBy` fields. Type exports should be updated to include these derived fields. Best implemented in the `DataType` class to work for all record/document types.

### Previous discussions 

- https://github.com/digidem/comapeo-schema/issues/83
- https://github.com/digidem/comapeo-core/issues/153

RangerMauve:
Fixes #136 

Feels like I should do something different for the core ownership case, or maybe have some sort of in-memory cache?


gmaclennan (approved):
Looks solid to me.

Could you also just add a [`@deprecated`](https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#deprecated) tag to  `$originalVersionIdToDeviceId()`?

As you say, we'd get some benefit to caching this, we can wrap in [`p-memoize`](https://www.npmjs.com/package/p-memoize) with a [`quick-lru`](https://github.com/sindresorhus/quick-lru) cache with maybe 1000 entries to start - should be plenty for most projects. Happy to add that later since this already is much faster than the current RPC round-trip made for every value.

PR #1124, opened 2025-09-23
wip: Upgrade to hypercore11 cinematic universe

### LINKED ISSUE #1042: Upgrade to hypercore 11
## Description

Hypercore came out with a new version that uses rocksdb for storage. this should be faster and will hopefully avoid some issues with filesystem specific quirks of the sparse files storage system. We should adopt this sooner than later as it will require migrating data over which could require double the storage temporarily.

## Tasks

- [ ] Review migration code [here](https://github.com/holepunchto/hypercore-storage/tree/main/migrations/0)
- [ ] Test that storage migration to rocksdb works
- [ ] Add check that we have enough storage on the device
  - [ ] Do we need platform specific APIs for this?
  - [ ] Send signal to front end about needing time to migrate?
- [ ] Test sync compatibility with older versions

RangerMauve:
Closes #1042


RangerMauve:
Closing in favor of #1246

PR #1125, opened 2025-09-25
fix: Rework initial settings storage on invitee v2

gmaclennan:
This is a rework of #1123, because during review I realised it was not necessary to maintain a separate table because we already store project info in the keys table.

This changes how leaving a project is tracked: it is now written to the project keys table _before_ clearing project data.

A previous fix to ensure data is cleared after leaving a project used the change of role to acertain if the project had been left, however because this check was made every time a project instance was created, including when adding a new project from an invite, it was actually creating a big delay, because a project instance would not be returned until getOwnRole returned, which waited for initial sync.

This adds a `status` prop to the `manager.listProjects()` function to be explicit about when a project is in the "joining" state, and adds an additional "left" status.

There is a change of behaviour here where $getProjectSettings() will still return the basic project info (name, description, color, sendStats) even after leaving a project, which makes it consistent with what is returned from `listProjects()`.

The test for "partly-left projects" should be more robust now too - it doesn't "pretend" to partly leave a project, but instead prematurely exits the process to more realistically simulate what would happen in real usage if leave project fails.

Also fixes a small type bug for the `projectInviteId` column of the projectKeysTable, which was defined as `unknown` because it was defaulting to a JSON column I think.

### DIFF
--- a/drizzle/client/0004_fresh_pestilence.sql
+++ b/drizzle/client/0004_fresh_pestilence.sql
@@ -1,5 +1,6 @@
 CREATE TABLE `backupProjectInfo` (
 	`projectId` text PRIMARY KEY NOT NULL,
 	`projectName` text NOT NULL,
-	`projectDescription` text
+	`projectDescription` text,
+	`sendStats` integer DEFAULT false NOT NULL
 );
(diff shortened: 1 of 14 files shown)


drizzle/client/0004_fresh_pestilence.sql
```diff
@@ -0,0 +1,6 @@
+CREATE TABLE `backupProjectInfo` (
```
RangerMauve, line 1: I think we should delete these two files from the journal and run another db:generate:client so that we can reduce the number of migrations happening

RangerMauve (changes requested):
LGTM overall, cleaner than before

PR #1162, opened 2025-10-23
Feat/members remove

### LINKED ISSUE #258: implement `member.remove()` method in the MemberApi
## Description

Extracted from #133 . Note that this is currently descoped for the MVP release.

## Tasks

- [x] assign role to BLOCKED
- [x] Allow blocked role to sync auth cores still
- [x] Test the role assignment propagates and the blocked user sees it
- [x] Test what happens if the removed member is disconnected when the remove happens and then reconnects (fix in separate PR if this is an issue)
- [x] Add event `project.on('own-role-update')` to listen for own role changing
- [x] Test new methods

RangerMauve:
Closes #258 

- Add `project.$member.remove` API
- Add `project.on('own-role-change', {role})` event
- Allow BLOCKED_ROLE to sync auth cores
- Test for remove while connected
- Test for remove while disconnected

### DIFF
--- a/test-e2e/members.js
+++ b/test-e2e/members.js
@@ -270,7 +270,11 @@ test('roles - creator role and role assignment', async (t) => {
   const project = await manager.getProject(projectId)
   const ownRole = await project.$getOwnRole()
 
-  assert.deepEqual(ownRole, CREATOR_ROLE, 'Project creator has creator role')
+  assert.equal(
+    ownRole.roleId,
+    CREATOR_ROLE.roleId,
+    'Project creator has creator role'
+  )
 
   const deviceId = randomBytes(32).toString('hex')
   await project.$member.assignRole(deviceId, MEMBER_ROLE_ID)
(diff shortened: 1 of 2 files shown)


gmaclennan:
> > What's the plan for handling being blocked on the backend?
> 
> So far I was thinking of leaving it to the front end to close or leave the project to help account for race conditions but I'm not sure what the reqirements are on that side. cc @ErikSin @cimigree If the project gets removed right after the event is fired, can the front end handle that?

I think it might result in indeterminate behaviour, and possible race conditions, because of the various things that are wired up to the backend that may call methods on the (now closed) project after it's blocked. I think it would be good to eventually do this all on the backend, but short-term I think what you propose is better.

However, this requires the frontend to call `manager.leaveProject()` when receiving the role change event to a blocked user, and currently [that will throw](https://github.com/digidem/comapeo-core/blob/335296d40b0741dcf2f0c1e7837c5ad8b8b54745/src/mapeo-project.js#L1286) if the user is already blocked, and the error being thrown will mean that [data is not cleared](https://github.com/digidem/comapeo-core/blob/335296d40b0741dcf2f0c1e7837c5ad8b8b54745/src/mapeo-project.js#L1318-L1319).

It's important that blocking a device results, on the blocked device:

1. Encryption keys for all namespaces other than auth are deleted
2. Data is deleted

It may well be that user requirements evolve to require different deletion behaviour for blocked devices vs. when a device leaves, so in the future we will want to support that, so we should eventually ensure this architecture will enable that.

RangerMauve:
Fixed by allowing leaveProject to run on blocked users. We just don't write the role update if we're blocked.

RangerMauve:
TODO:
- add reason to get own role
- add reason to members get many
- next month revisit membership records (add status? How do we determine membership?)

PR #1171, opened 2025-11-03
fix: Should be able to reinvite after blocking

RangerMauve:
Needed for member removal to make sense.

Currently test is freezing up

### DIFF
--- a/test-e2e/project-leave.js
+++ b/test-e2e/project-leave.js
@@ -463,6 +463,7 @@ test('Member can join project again after leaving', async (t) => {
     roleId: MEMBER_ROLE_ID,
   })
 
+  // We need to re-get the project since it was closed
   const reMemberProject = await member.getProject(projectId)
 
   assert.notEqual(reMemberProject, memberProject, 'new project on rejoin')


RangerMauve:
Test for the case where member has left also freezes up. Likely there's a condition in the invite flow. Will debug if on inviter or invitee side and fix tomorrow. 👍

src/mapeo-manager.js
```diff
@@ -1064,6 +1069,9 @@ export class MapeoManager extends TypedEmitter {
     const project = await this.getProject(projectPublicId)
 
     await project[kProjectLeave]()
+
+    // Sync any role changes from prpoject leave
```
achou11:
```suggestion
    // Sync any role changes from project leave
```

src/invite/invite-api.js
achou11: trying to confirm my understanding...

do the changes in this file account for when someone removes a member and tries to invite them back BEFORE the removed member acknowledges it and calls `leaveProject()` on their end? Seems like that's the case but maybe an explicit test would be helpful?

RangerMauve: Current tests don't cover re-inviting before project-leave happens. I can add a test. Expected behavior? I propose auto-reject.

achou11: > Expected behavior? I propose auto-reject

That seems reasonable to me!

test-e2e/project-leave.js
```diff
@@ -420,4 +416,66 @@ test('leaving a project before PR#1125 persists after PR#1125', async (t) => {
...
+    'creator successfully added from creator perspective'
+  )
+
+  await waitForSync(projects, 'initial')
+
+  await member.leaveProject(projectId)
+
+  // Close the project after you leave and sync
+  // This clears up resources so we can be reinvited
+  await memberProject.close()
```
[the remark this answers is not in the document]
RangerMauve: I think it broke the sync when I tried to do it before but having the wait for initial sync inside close might make that irrelevant

src/invite/invite-api.js
RangerMauve: Will get on this tomorrow morn

PR #1173, opened 2025-11-05
test: Check project isnt listed after member is removed

RangerMauve:
Added a check to make sure project is marked as left after a member is removed from it.

Seems to be working?

cc @ErikSin @cimigree @achou11


ErikSin:
I also don't know the internal of the core, so perhaps this is what happens under the hood when a user is removed from a project?

ErikSin:
Sorry, perhaps i don't fully understand the test. But it looks like to me that the `invitor` is removing the invitee. Then the `invitee` is leaving the project, then you are checking the output of `listProjects`.

Is that correct?

If so, we need to take out that middle step of the invitee leaving the project, unless we are misunderstanding how this API should be used?

PR #1182, opened 2025-11-26
fix: Auto close active project when being invited back into it

### LINKED ISSUE #1179: Enable re-initializing a project after leaveProject is called and we get an invitw
## Description

We need a way to be able to rejoin a project that has been left (blocked or self removed) but is still loaded in memory. Currently we auto reject invites for projects that are actively loaded, and we don't yet have a way to re-hydrate deleted data from a left project.

## Tasks

- [ ] Add new code path for when we get an invite with an active project that is left
- [ ] Re-init things unloaded by `leaveProject`
- [ ] Add unit tests for re-joining without calling `project.close()`

RangerMauve:
Fixes #1179 

Easier than re-hydrating the project instance. We just close it if we know we left it.

Instead of erroring saying we have an active project, we error on having the project key and not having left.

With this we no longer need to close projects after leaving them and can leave them active to sync.

### DIFF
--- a/src/mapeo-manager.js
+++ b/src/mapeo-manager.js
@@ -679,7 +679,7 @@ export class MapeoManager extends TypedEmitter {
   ) => {
     const projectPublicId = projectKeyToPublicId(projectKey)
 
-    // 3. Check if the project exists in the project keys table
+    // 1. Check if the project exists in the project keys table
     // If it does, that means the project has already been either created or added before
     const projectId = projectKeyToId(projectKey)
     const projectInviteId = projectKeyToProjectInviteId(projectKey)


src/mapeo-manager.js
```diff
@@ -679,14 +679,7 @@ export class MapeoManager extends TypedEmitter {
...
 
-    // 1. Check for an active project
-    const activeProject = this.#activeProjects.get(projectPublicId)
-
-    if (activeProject) {
-      throw new Error(`Project with ID ${projectPublicId} already exists`)
-    }
-
-    // 2. Check if the project exists in the project keys table
+    // 3. Check if the project exists in the project keys table
```
gmaclennan:
```suggestion
    // 1. Check if the project exists in the project keys table
```

PR #1186, opened 2025-12-02
feat!: only list active members

### LINKED ISSUE #431: Should members who have left a project be listed as members in end-user APIs?
## Description

We implemented `project.leave()` which updates a project member's role to `LEFT`. There's an open question as to whether listing these members when calling `MemberAPI.getMany()` should omit them by default (currently they're included). 

Gregor's potential suggestion is to omit them by default and include an option for the method e.g. `includeLeft` that would include them as an opt-in behavior.

Initial feeling is that we should implement this as described by the suggestion

RangerMauve:
Closes #431 

By default `project.$member.getMany()` will now only return active members unless you add the `includeInactive` flag.

I also tried to make the JSDoc types reflect that omitting the flag only yields active members. Hopefully that works and is helpful?

I adjusted some of the tests that were expecting inactive members to be in there. Also added a test to make sure the flag was working as expected outside of the existing tests that depend on it implicitly.

Review requests:

- Does the naming make sense? Feel free to propose other names.
- Are the types looking good?

### DIFF
--- a/src/member-api.js
+++ b/src/member-api.js
@@ -590,85 +590,45 @@ export class MemberApi extends TypedEmitter {
 
     const deviceInfoByConfigCoreId = keyBy(allDeviceInfo, ({ docId }) => docId)
 
-    if (includeLeft) {
-      /**
-       * @type {Array<Promise<MemberInfo>>}
-       */
-      const memberInfoPromises = []
+    /**
+     * @type {Array<Promise<MemberInfo>>}
+     */
+    const activeMemberInfoPromises = []
 
-      for (const [deviceId, role] of allRoles.entries()) {
-        const getMemberInfo = async () => {
-          /** @type {MemberInfo} */
-          const memberInfo = { deviceId, role }
-
-          try {
-            const configCoreId = await this.#coreOwnership.getCoreId(
-              deviceId,
-              'config'
-            )
-
-            const deviceInfo = deviceInfoByConfigCoreId.get(configCoreId)
-
-            memberInfo.name = deviceInfo?.name
-            memberInfo.deviceType = deviceInfo?.deviceType
-            memberInfo.joinedAt = deviceInfo?.createdAt
-            memberInfo.selfHostedServerDetails =
-              deviceInfo?.selfHostedServerDetails
-          } catch (err) {
-            // Attempting to get someone else may throw because sync hasn't occurred or completed
-            // Only throw if attempting to get themself since the relevant information should be available
-            if (deviceId === this.#ownDeviceId) throw err
-          }
-
-          return memberInfo
-        }
-
-        memberInfoPromises.push(getMemberInfo())
+    for (const [deviceId, role] of allRoles.entries()) {
+      if (!includeLeft && !isActiveMemberRole(role)) {
+        continue
       }
 
-      return Promise.all(memberInfoPromises)
-    } else {
-      /**
-       * @type {Array<Promise<ActiveMemberInfo>>}
-       */
-      const activeMemberInfoPromises = []
+      const getMemberInfo = async () => {
+        /** @type {MemberInfo} */
+        const memberInfo = { deviceId, role }
 
-      for (const [deviceId, role] of allRoles.entries()) {
-        if (!isActiveMemberRole(role)) {
-          continue
+        try {
+          const configCoreId = await this.#coreOwnership.getCoreId(
+            deviceId,
+            'config'
+          )
+
+          const deviceInfo = deviceInfoByConfigCoreId.get(configCoreId)
+
+          memberInfo.name = deviceInfo?.name
+          memberInfo.deviceType = deviceInfo?.deviceType
+          memberInfo.joinedAt = deviceInfo?.createdAt
+          memberInfo.selfHostedServerDetails =
+            deviceInfo?.selfHostedServerDetails
+        } catch (err) {
+          // Attempting to get someone else may throw because sync hasn't occurred or completed
+          // Only throw if attempting to get themself since the relevant information should be available
+          if (deviceId === this.#ownDeviceId) throw err
         }
 
-        const getMemberInfo = async () => {
-          /** @type {ActiveMemberInfo} */
-          const memberInfo = { deviceId, role }
-
-          try {
-            const configCoreId = await this.#coreOwnership.getCoreId(
-              deviceId,
-              'config'
-            )
-
-            const deviceInfo = deviceInfoByConfigCoreId.get(configCoreId)
-
-            memberInfo.name = deviceInfo?.name
-            memberInfo.deviceType = deviceInfo?.deviceType
-            memberInfo.joinedAt = deviceInfo?.createdAt
-            memberInfo.selfHostedServerDetails =
-              deviceInfo?.selfHostedServerDetails
-          } catch (err) {
-            // Attempting to get someone else may throw because sync hasn't occurred or completed
-            // Only throw if attempting to get themself since the relevant information should be available
-            if (deviceId === this.#ownDeviceId) throw err
-          }
-
-          return memberInfo
-        }
-
-        activeMemberInfoPromises.push(getMemberInfo())
+        return memberInfo
       }
-
-      return Promise.all(activeMemberInfoPromises)
+      activeMemberInfoPromises.push(getMemberInfo())
     }
+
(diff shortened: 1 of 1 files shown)


src/member-api.js
```diff
@@ -56,6 +60,10 @@ import {
  * @prop {string} selfHostedServerDetails.baseUrl
  */
 
+/**
+ * @typedef {MemberInfo & {role: ROLES[MEMBER_ROLE_ID] | ROLES[COORDINATOR_ROLE_ID] | ROLES[CREATOR_ROLE_ID]}} ActiveMemberInfo
```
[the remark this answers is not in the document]
RangerMauve: I was thinking this could be convenient for the consumer side.

cc @achou11 @ErikSin @cimigree Thoughts on whether this would be nice to have?

src/member-api.js
```diff
@@ -561,9 +569,23 @@ export class MemberApi extends TypedEmitter {
...
+   * @param {object} [opts]
+   * @param {false} [opts.includeLeft=false]
+   * @returns {Promise<Array<ActiveMemberInfo>>}
+   */
+  /**
+   * List members in the project. By default only active Members and Coordinators are returned
+   * @param {object} [opts]
+   * @param {boolean} [opts.includeLeft=false] Set to true to list removed members
+   */
+  async getMany({ includeLeft = false } = {}) {
```
achou11: haven't wrapped my head around a potential fix but the type inference isn't working correctly when the param isn't specified:

```tsx
// Raises a type error

/** @type {Array<import('../src/member-api.js').ActiveMemberInfo>} */
const activeMembers = await project.$member.getMany()
```

```tsx
// This works

/** @type {Array<import('../src/member-api.js').ActiveMemberInfo>} */
const activeMembers = await project.$member.getMany({ includeLeft:false })
```

RangerMauve: I think I just need another overload for where opts isn't provided

RangerMauve: Fixed!

src/member-api.js
```diff
@@ -572,31 +596,35 @@ export class MemberApi extends TypedEmitter {
...
-          if (deviceId === this.#ownDeviceId) throw err
-        }
-
-        return memberInfo
-      })
+      [...allRoles.entries()]
+        .filter(
+          ([_, { roleId }]) => includeLeft || ACTIVE_ROLE_IDS.includes(roleId)
+        )
+        .map(async ([deviceId, role]) => {
```
achou11: will the size of `allRoles` ever get to a point where it's worth doing more optimal iteration instead of the 4(?) array allocations happening here? i guess `$member.getMany()` may not be a hot path, but what's here stood out to me as something that can be achieved with a single for loop 😄

src/member-api.js
achou11: took a while but think I was able to update the type declaration and implementation such that TypeScript actually checks that the return type is correctly inferred based on the input type. In order to get stronger guarantees, it requires separating the internal implementation based on the input parameter:


```diff
diff --git a/src/member-api.js b/src/member-api.js
index 77924fd..48fe62b 100644
--- a/src/member-api.js
+++ b/src/member-api.js
@@ -36,7 +36,7 @@ const ACTIVE_ROLE_IDS = [CREATOR_ROLE_ID, MEMBER_ROLE_ID, COORDINATOR_ROLE_ID]
  *   DeviceInfo,
  *   DeviceInfoValue,
  *   ProjectSettings,
- *   ProjectSettingsValue
+ *   ProjectSettingsValue,
  * } from '@comapeo/schema'
  */
 /** @import { Promisable } from 'type-fest' */
@@ -570,22 +570,17 @@ export class MemberApi extends TypedEmitter {
 
   /**
    * @overload
-   * @param {object} opts
-   * @param {true} opts.includeLeft
    * @returns {Promise<Array<MemberInfo>>}
    */
   /**
+   * @template {boolean} [T=false]
+   *
    * @overload
-   * @returns {Promise<Array<ActiveMemberInfo>>}
-   */
-  /**
-   * @overload
-   * @param {object} opts
-   * @param {false} opts.includeLeft
-   * @returns {Promise<Array<ActiveMemberInfo>>}
-   */
-  /**
-   * List members in the project. By default only active Members and Coordinators are returned
+   * @param {Object} opts
+   * @param {T} [includeLeft=false]
+   *
+   * @returns {Promise<T extends true ? Array<MemberInfo> : Array<ActiveMemberInfo>>}
+   *
    */
   async getMany({ includeLeft = false } = {}) {
     const [allRoles, allDeviceInfo] = await Promise.all([
@@ -595,12 +590,14 @@ export class MemberApi extends TypedEmitter {
 
     const deviceInfoByConfigCoreId = keyBy(allDeviceInfo, ({ docId }) => docId)
 
-    return Promise.all(
-      [...allRoles.entries()]
-        .filter(
-          ([_, { roleId }]) => includeLeft || ACTIVE_ROLE_IDS.includes(roleId)
-        )
-        .map(async ([deviceId, role]) => {
+    if (includeLeft) {
+      /**
+       * @type {Array<Promise<MemberInfo>>}
+       */
+      const memberInfoPromises = []
+
+      for (const [deviceId, role] of allRoles.entries()) {
+        const getMemberInfo = async () => {
           /** @type {MemberInfo} */
           const memberInfo = { deviceId, role }
 
@@ -624,8 +621,54 @@ export class MemberApi extends TypedEmitter {
           }
 
           return memberInfo
-        })
-    )
+        }
+
+        memberInfoPromises.push(getMemberInfo())
+      }
+
+      return Promise.all(memberInfoPromises)
+    } else {
+      /**
+       * @type {Array<Promise<ActiveMemberInfo>>}
+       */
+      const activeMemberInfoPromises = []
+
+      for (const [deviceId, role] of allRoles.entries()) {
+        if (!isActiveMemberRole(role)) {
+          continue
+        }
+
+        const getMemberInfo = async () => {
+          /** @type {ActiveMemberInfo} */
+          const memberInfo = { deviceId, role }
+
+          try {
+            const configCoreId = await this.#coreOwnership.getCoreId(
+              deviceId,
+              'config'
+            )
+
+            const deviceInfo = deviceInfoByConfigCoreId.get(configCoreId)
+
+            memberInfo.name = deviceInfo?.name
+            memberInfo.deviceType = deviceInfo?.deviceType
+            memberInfo.joinedAt = deviceInfo?.createdAt
+            memberInfo.selfHostedServerDetails =
+              deviceInfo?.selfHostedServerDetails
+          } catch (err) {
+            // Attempting to get someone else may throw because sync hasn't occurred or completed
+            // Only throw if attempting to get themself since the relevant information should be available
+            if (deviceId === this.#ownDeviceId) throw err
+          }
+
+          return memberInfo
+        }
+
+        activeMemberInfoPromises.push(getMemberInfo())
+      }
+
+      return Promise.all(activeMemberInfoPromises)
+    }
   }
 
   /**
@@ -638,6 +681,15 @@ export class MemberApi extends TypedEmitter {
   }
 }
 
+/**
+ * @param {import('./roles.js').Role} role
+ *
+ * @returns {role is ActiveMemberInfo['role']}
+ */
+function isActiveMemberRole(role) {
+  return ACTIVE_ROLE_IDS.includes(role.roleId)
+}
+
 /**
  * @param {string} baseUrl
  * @param {object} options
****
```

admittedly one of those unfortunate cases where you're writing code to appease TypeScript a bit. can do a bit more to reuse some logic but otherwise this is the kind of work needed to get TypeScript to be useful 😅 - hopefully helpful to at least see the diff!

Technically, there's still some level of trust in the implementer because we could make the `isActiveMemberRole` helper lie to us, but unlikely given how small in scope that narrowing function is.

achou11 (commented):
In case it's helpful and informs any additional changes to this PR

achou11 (approved):
Couple of non-blocking questions but otherwise changes lgtm

src/member-api.js
```diff
@@ -561,43 +569,106 @@ export class MemberApi extends TypedEmitter {
...
+
+      return Promise.all(memberInfoPromises)
+    } else {
+      /**
+       * @type {Array<Promise<ActiveMemberInfo>>}
+       */
+      const activeMemberInfoPromises = []
+
+      for (const [deviceId, role] of allRoles.entries()) {
+        if (!isActiveMemberRole(role)) {
```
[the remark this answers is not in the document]
RangerMauve: Cleaned up and the types should still work.

gmaclennan (approved):

PR #1219, opened 2026-02-04
feat: Add map share rpc protobufs

gmaclennan:
This PR replaces (adds to) #1198

I adjusted the protobuf to have a single `mapShareUrls` field, to match the map server implementation, and I added a `receiverDeviceId` field - each map share is created for a specific receiver.

I cleaned up some of the typescript and validation function.


RangerMauve (approved):
LGTM

PR #1220, opened 2026-02-04
Feat/standardize errors

### LINKED ISSUE #84: Custom / consistent errors
I think we need to define the errors we raise with consistent error codes. This will enable the front end to display the correct UI to the user depending on the type of error.

Some errors will be expected, e.g.

- Blob or Record not found
- Peer disconnects

The simple way to do this is with instance checks e.g.

```js
class ErrNotFound extends Error {
  constructor(message) {
    super(message)
  }
}

if (err instanceof ErrNotFound) {
  //...
}
```

However this doesn't work over RPC. I recommend defining error codes, then each error should have a `.code` property. [`rpc-reflector`](https://github.com/gmaclennan/rpc-reflector/blob/master/package.json) uses [`serialize-error`](https://github.com/sindresorhus/serialize-error) to serialize errors over RPC, so it will preserve `error.code` properties.

I was just looking through the [Node code for how they define errors](https://github.com/nodejs/node/blob/7cb09f40a69846fd5283626a095eb1b995d529fc/lib/internal/errors.js#L366), which seems like a good start. I would have thought there would be a neat library on npm to define errors, but it's a hard thing to find the right search terms.

RangerMauve:
Closes #84

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -34,6 +34,7 @@ export { MapeoManager } from './mapeo-manager.js'
 /**
  * @namespace MemberApi
  * @typedef {import('./member-api.js').MemberInfo} MemberApi.MemberInfo
+ * @typedef {import('./member-api.js').ActiveMemberInfo} MemberApi.ActiveMemberInfo
  * @typedef {import('./roles.js').RoleId} MemberApi.RoleId
  * @typedef {import('./roles.js').RoleIdForNewInvite} MemberApi.RoleIdForNewInvite
  * @typedef {import('./roles.js').RoleIdAssignableToOthers} MemberApi.RoleIdAssignableToOthers
--- a/src/member-api.js
+++ b/src/member-api.js
@@ -39,18 +39,22 @@ import {
 import { wsCoreReplicator } from './lib/ws-core-replicator.js'
 import {
   BLOCKED_ROLE_ID,
+  COORDINATOR_ROLE_ID,
+  CREATOR_ROLE_ID,
   LEFT_ROLE_ID,
   MEMBER_ROLE_ID,
   ROLES,
   isRoleIdForNewInvite,
 } from './roles.js'
 
+const ACTIVE_ROLE_IDS = [CREATOR_ROLE_ID, MEMBER_ROLE_ID, COORDINATOR_ROLE_ID]
+
 /**
  * @import {
  *   DeviceInfo,
  *   DeviceInfoValue,
  *   ProjectSettings,
- *   ProjectSettingsValue
+ *   ProjectSettingsValue,
  * } from '@comapeo/schema'
  */
 /** @import { Promisable } from 'type-fest' */
@@ -72,6 +76,10 @@ import {
  * @prop {DeviceInfo['createdAt']} [joinedAt]
  * @prop {object} [selfHostedServerDetails]
  * @prop {string} selfHostedServerDetails.baseUrl
+ */
+
+/**
+ * @typedef {Omit<MemberInfo, 'role'> & {role: import('./roles.js').Role<typeof MEMBER_ROLE_ID | typeof COORDINATOR_ROLE_ID | typeof CREATOR_ROLE_ID>}} ActiveMemberInfo
  */
 
 export class MemberApi extends TypedEmitter {
@@ -571,9 +579,20 @@ export class MemberApi extends TypedEmitter {
   }
 
   /**
-   * @returns {Promise<Array<MemberInfo>>}
+   * @overload
+   * @returns {Promise<Array<ActiveMemberInfo>>}
    */
-  async getMany() {
+  /**
+   * @template {boolean} [T=false]
+   *
+   * @overload
+   * @param {Object} opts
+   * @param {T} [opts.includeLeft=false]
+   *
+   * @returns {Promise<T extends true ? Array<MemberInfo> : Array<ActiveMemberInfo>>}
+   *
+   */
+  async getMany({ includeLeft = false } = {}) {
     const [allRoles, allDeviceInfo] = await Promise.all([
       this.#roles.getAll(),
       this.#dataTypes.deviceInfo.getMany(),
@@ -581,8 +600,17 @@ export class MemberApi extends TypedEmitter {
 
     const deviceInfoByConfigCoreId = keyBy(allDeviceInfo, ({ docId }) => docId)
 
-    return Promise.all(
-      [...allRoles.entries()].map(async ([deviceId, role]) => {
+    /**
+     * @type {Array<Promise<MemberInfo>>}
+     */
+    const activeMemberInfoPromises = []
+
+    for (const [deviceId, role] of allRoles.entries()) {
+      if (!includeLeft && !isActiveMemberRole(role)) {
+        continue
+      }
+
+      const getMemberInfo = async () => {
         /** @type {MemberInfo} */
         const memberInfo = { deviceId, role }
 
@@ -608,8 +636,11 @@ export class MemberApi extends TypedEmitter {
         }
 
         return memberInfo
-      })
-    )
+      }
+      activeMemberInfoPromises.push(getMemberInfo())
+    }
+
+    return Promise.all(activeMemberInfoPromises)
   }
 
   /**
@@ -620,6 +651,15 @@ export class MemberApi extends TypedEmitter {
   async assignRole(deviceId, roleId) {
     return this.#roles.assignRole(deviceId, roleId)
   }
+}
+
+/**
+ * @param {import('./roles.js').Role} role
+ *
+ * @returns {role is ActiveMemberInfo['role']}
+ */
+function isActiveMemberRole(role) {
+  return ACTIVE_ROLE_IDS.includes(role.roleId)
 }
 
 /**
--- a/test-e2e/members.js
+++ b/test-e2e/members.js
@@ -103,7 +103,7 @@ test('getting yourself after adding project (but not yet synced)', async (t) => 
     'has expected member info with no role'
   )
 
-  const members = await project.$member.getMany()
+  const members = await project.$member.getMany({ includeLeft: true })
   const { joinedAt: _, ...member } = members[0]
 
   assert.equal(members.length, 1)
@@ -377,7 +377,7 @@ test('roles - getMany() on newly invited device before sync', async (t) => {
 
   const expected = { [deviceId]: NO_ROLE }
 
-  const allMembers = await project.$member.getMany()
+  const allMembers = await project.$member.getMany({ includeLeft: true })
 
   /** @type {Record<string, import('../src/roles.js').Role>} */
   const actual = {}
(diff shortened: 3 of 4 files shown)


package.json
achou11: Need to run `npm run build:errors` here first

achou11:
```suggestion
    "test": "npm-run-all lint test:prettier build:errors build:types type test:unit test:e2e test:types",
```

Alternatively can define a `build` script like: `"build": "npm-run-all build:*"` can use that instead:

```suggestion
    "test": "npm-run-all lint test:prettier build type test:unit test:e2e test:types",
```

achou11: Need to run `build:errors` script here first. see https://github.com/digidem/comapeo-core/pull/1220/changes#r2794756653 for suggestions

RangerMauve:
CI is freezing somewhere, been debugging it, should be ready for final review after

scripts/build-error-codes.js
achou11: think it'd be helpful to add a comment to the generated file that mentions that the file is generated from this script.

src/fastify-plugins/icons.js
```diff
@@ -153,6 +159,6 @@ function assertValidSize(value) {
  */
 function assertValidPixelDensity(value) {
   if (!VALID_PIXEL_DENSITIES.includes(/** @type {any} */ (value))) {
-    throw new Error(`${value} is not a valid icon pixel density`)
+    throw new InvalidIconPixelDensityError(value)
   }
 }
```
achou11: This should be fixed up to tighten up the custom error implementation. The input is only ever a number so can be updated as follows:

```ts
/**
 * @param {number} value
 * @returns {asserts value is import('../icon-api.js').BitmapOpts['pixelDensity']}
 */
function assertValidPixelDensity(value) {
  if (
    !VALID_PIXEL_DENSITIES.includes(
      /** @type {import('../icon-api.js').BitmapOpts['pixelDensity']} */ (value)
    )
  ) {
    throw new InvalidIconPixelDensityError(value)
  }
}
```

This allows updating the constructor param type of `InvalidIconPixelDensityError` to use `number`

src/errors.js
```diff
@@ -1,63 +1,723 @@
...
+  constructor(value) {
+    super(`'${value}' is not a valid icon size`)
+    this.name = 'InvalidIconSizeError'
+    this.code = 'INVALID_ICON_SIZE_ERROR'
+    this.status = 400
+  }
+}
+
+export class InvalidIconPixelDensityError extends Error {
+  /** @param {any} [density] */
```
achou11: See other comment about fixing up the implementation where this error is used.

```suggestion
  /** @param {number} density */
```

src/errors.js
```diff
@@ -66,6 +726,162 @@ export class PeerNotFoundError extends Error {
...
+    super(message)
+    this.name = 'AlreadyInvitingError'
+    this.code = 'ALREADY_INVITING_ERROR'
+    this.status = 409
+  }
+}
+
+export class InvalidRoleIDForNewInviteError extends Error {
+  constructor() {
+    super('Invalid role ID for new invite')
```
achou11: Should this surface the role ID?

src/errors.js
```diff
@@ -66,6 +726,162 @@ export class PeerNotFoundError extends Error {
...
+    super(message)
+    this.name = 'InvalidMapShareError'
+    this.code = 'INVALID_MAP_SHARE_ERROR'
+    this.status = 400
+  }
+}
+
+export class InvalidResponseBodyError extends Error {
+  constructor() {
+    super('InvalidResponseBodyError: Response body is not valid')
```
achou11:
```suggestion
    super('Response body is not valid')
```

src/errors.js
```diff
@@ -66,6 +726,162 @@ export class PeerNotFoundError extends Error {
...
+    this.code = 'INVALID_PROJECT_SECRET_KEY_ERROR'
+    this.status = 400
+  }
+}
+
+export class InvalidProjectJoinDetailsError extends Error {
+  /** @param {string} [message] */
+  constructor(
+    message = 'InvalidProjectJoinDetailsError: Invalid project join details'
+  ) {
```
achou11: Based on usage, think it's okay to make the param required


```suggestion
  /** @param {string} message */
  constructor(message) {
```

src/errors.js
```diff
@@ -66,6 +726,162 @@ export class PeerNotFoundError extends Error {
...
+    super(message)
+    this.name = 'UnexpectedError'
+    this.code = 'UNEXPECTED_ERROR'
+    this.status = 500
+  }
+}
+
+export class AutoStopTimeoutError extends Error {
+  constructor(
+    message = 'AutoStopTimeoutError: auto-stop timeout must be Infinity or a positive integer between 0 and the largest 32-bit signed integer'
```
achou11:
```suggestion
    message = 'Auto-stop timeout must be Infinity or a positive integer between 0 and the largest 32-bit signed integer'
```

src/errors.js
```diff
@@ -66,6 +726,162 @@ export class PeerNotFoundError extends Error {
...
+    super(message)
+    this.name = 'InvalidDrizzleQueryError'
+    this.code = 'INVALID_DRIZZLE_QUERY_ERROR'
+    this.status = 400
+  }
+}
+
+export class InvalidDrizzleJournalError extends Error {
+  /** @param {string} [message] */
+  constructor(message = 'invalid journal') {
```
achou11:
```suggestion
  constructor(message = 'Invalid journal') {
```

src/errors.js
```diff
@@ -1,63 +1,723 @@
...
+    super(message)
+    this.name = 'PeerDisconnectedError'
+    this.code = 'PEER_DISCONNECTED_ERROR'
+    this.status = 504
+  }
+}
+
+export class PeerFailedConnectionError extends Error {
+  /** @param {string} [message] */
+  constructor(message = 'PeerFailedConnectionError') {
```
achou11: This should be more human-oriented and not just the error class name.

src/errors.js
```diff
@@ -1,63 +1,723 @@
...
+    super('Cannot run multiple category imports at the same time')
+    this.name = 'MultipleCategoryImportsError'
+    this.code = 'MULTIPLE_CATEGORY_IMPORTS_ERROR'
+    this.status = 409
+  }
+}
+
+export class UnknownPeerError extends Error {
+  /** @param {string} [message] */
+  constructor(message = 'UnknownPeerError') {
```
achou11: This should be more human-oriented and not just the error class name.

src/errors.js
```diff
@@ -1,63 +1,723 @@
...
+    super(message)
+    this.name = 'InvalidDeviceInfoError'
+    this.code = 'INVALID_DEVICE_INFO_ERROR'
+    this.status = 400
+  }
+}
+
+export class RoleAssignError extends Error {
+  /** @param {string} [message] */
+  constructor(message = 'Cannot assign this role to this device') {
```
achou11: Based on usage, seems okay to make this a required param
```suggestion
  /** @param {string} message */
  constructor(message) {
```

package.json
```diff
@@ -4,24 +4,36 @@
...
   "main": "src/index.js",
   "types": "dist/index.d.ts",
+  "exports": {
+    ".": {
+      "import": "./src/index.js",
+      "types": "./dist/index.d.ts"
+    },
+    "./errors.js": {
+      "import": "./src/error_codes.js",
+      "types": "./dist/error_codes.d.ts"
```
achou11: Need to switch the ordering here (https://publint.dev/rules#exports_types_should_be_first).

Also file name seems to be incorrect? (should be hyphen instead of underscore)

```suggestion
      "types": "./dist/error-codes.d.ts",
      "import": "./src/error-codes.js"
```

achou11 (changes requested):
The implementation of the custom error classes seems a bit inconsistent so most of the comments are related to cleaning that up a bit. A good number are decently nitpicky.

Also some adjustments needed for the package exports changes.

src/errors.js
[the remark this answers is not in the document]
achou11: part of me thinks that it isn't useful for these errors to define their own numeric status to begin with. feels sufficient to let the context in which they're used (e.g. an http server) to set the appropriate status, which will mostly be the more common http codes. clients can differentiate between the different error types based on the `code` field you've defined here.

maybe i'm missing something about how having these could be useful?

RangerMauve:
TY for the thorough review! Gonna finish this off on Monday

package.json
```diff
@@ -4,24 +4,36 @@
   "description": "Offline p2p mapping library",
   "main": "src/index.js",
   "types": "dist/index.d.ts",
+  "exports": {
+    ".": {
+      "types": "./dist/index.d.ts",
+      "import": "./src/index.js"
+    },
+    "./errors.js": {
```
achou11: just checking: the name of this export is desired over something like `./error-codes.js`? I think it's fine to keep what you have - just want to make sure that's an intentional choice.

package.json
```diff
@@ -4,24 +4,36 @@
...
   "scripts": {
     "lint": "eslint --cache .",
     "format": "prettier . --write",
-    "test": "npm-run-all lint test:prettier build:types type test:unit test:e2e test:types",
+    "test": "npm-run-all lint test:prettier build:errors build:types type test:unit test:e2e test:types",
     "test:prettier": "prettier --check .",
     "test:unit": "node --test",
     "test:e2e": "node --test test-e2e/*.js test-e2e/**/*.js",
     "test:types": "tsc -p test-types/tsconfig.json",
+    "build": "npm-run-all build:*",
```
achou11: Realizing my suggestion to use the wildcard here is a bit misguided because you need to make sure that the errors script is run before the types script. Instead of relying on order in package.json (which I think is what's done internally when using a wildcard), probably safest to just name the exact scripts to run sequentially:

```suggestion
    "build": "npm-run-all build:errors build:types",
```

src/import-categories.js
```diff
@@ -349,19 +350,17 @@ function appliesToToGeometry(appliesTo) {
...
  * @throws {TypeError} if `map` is not a Map
  * @throws {Error} if `key` is not in `map` (with `msgOrError` as message or the default message)
  */
-function getOrThrow(map, key, msgOrError) {
+function getOrThrow(map, key) {
   if (!(map instanceof Map)) throw new TypeError('map must be a Map')
   if (!map.has(key)) {
-    if (msgOrError instanceof Error) throw msgOrError
-    throw new Error(msgOrError ?? `key ${key} not found in map`)
+    throw new KeyNotFoundError(key)
```
gmaclennan: Rather than make this a custom error, I think we would be better wrapping anywhere we use this in a try...catch and throwing a more descriptive custom error, with the cause being the plain error here. `KeyNotFoundError()` doesn't really help us in application code.

src/mapeo-project.js
```diff
@@ -658,7 +668,8 @@ export class MapeoProject extends TypedEmitter {
             index: entry.index,
           })
 
-          assert(doc.schemaName === 'translation', 'expected a translation doc')
+          // assert(doc.schemaName === 'translation', 'expected a translation doc')
+          if (doc.schemaName !== 'translation') continue
```
gmaclennan: I prefer throwing an error rather than `continue` because it allows us to add logging more easily (in the `catch`) in the future for debugging.

src/schema/json-schema-to-drizzle.js
```diff
@@ -20,14 +23,20 @@ export function jsonSchemaToDrizzleSqliteTable(
...
   /** @type {Record<string, any>} */
   const columns = {}
   for (const [key, value] of Object.entries(schema.properties)) {
     if (typeof value !== 'object') continue
     if (isArray(value.type) || typeof value.type === 'undefined') {
-      throw new Error('Cannot process JSONSchema as SQL table')
+      throw new InvalidComapeoSchemaFormatError(
+        tableName,
+        'Columns must have single known type'
+      )
```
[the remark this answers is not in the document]
RangerMauve: It's used inside `client.js` so I assume that gets loaded at runtime? It should really error at build time before runtime gets released with whatever schema in the first place.

RangerMauve:
@gmaclennan @achou11 Regarding the `message` param, I'm actually leaning more towards removing it from any error types that don't explicitly need it.

src/blob-store/index.js
```diff
@@ -7,7 +7,9 @@ import { noop } from '../utils.js'
 import { TypedEmitter } from 'tiny-typed-emitter'
 import { HyperdriveIndexImpl as HyperdriveIndex } from './hyperdrive-index.js'
 import { Logger } from '../logger.js'
-import { getErrorCode, getErrorMessage } from '../lib/error.js'
+import { BlobNotFoundError, getErrorCode } from '../errors.js'
+import { BlobsNotFoundError, DriveNotFoundError } from '../errors.js'
```
gmaclennan: merge to single import

src/local-peers.js
```diff
@@ -830,14 +836,14 @@ export class LocalPeers extends TypedEmitter {
...
-      throw new UnknownPeerError('Unknown peer ' + deviceId.slice(0, 7))
+      throw new UnknownPeerError(deviceId.slice(0, 7))
     }
     const peer = chooseDevicePeer(devicePeers)
     if (peer) return peer
     return new Promise((resolve, reject) => {
       const timeoutId = setTimeout(() => {
         this.off('peers', onPeers)
-        reject(new UnknownPeerError('Unknown peer ' + deviceId.slice(0, 7)))
+        reject(new UnknownPeerError(deviceId.slice(0, 7)))
```
gmaclennan:
```suggestion
        reject(new UnknownPeerError({ deviceId: deviceId.slice(0, 7) }))
```

PR #1230, opened 2026-02-12
Feat/invitor sets device info

RangerMauve:
- New flag in manager to skip writing own device info to config core (will enable in comapeo-core maybe?)
- Invitors write deviceInfo keyed to deviceId to account for invitees that don't / can't write their own info
- When getting member info we fallback to fetching it by their device ID
- Added test for case where deviceInfo is only written by invitor

TODO: How do we validate the deviceID based case? Initially we could check if the writer is a Coordinator, but this won't work for when coordinators are removed. This speaks to needing vector clocks keyed to membership state being written to auth docs so we can guarantee permissions at that point in time.

### DIFF
--- a/src/datatype/index.js
+++ b/src/datatype/index.js
@@ -9,6 +9,7 @@ import {
   InvalidDocError,
   InvalidDocFormatError,
   NotFoundError,
+  nullIfNotFound,
 } from '../errors.js'
 import { TypedEmitter } from 'tiny-typed-emitter'
 import { setProperty, getProperty } from 'dot-prop-extra'
@@ -195,7 +196,7 @@ export class DataType extends TypedEmitter {
    * @returns {Promise<TDoc & DerivedDocFields>}
    */
   async [kCreateOrUpdateWithDocId](docId, value) {
-    const existing = await this.getByDocId(docId).catch(noop)
+    const existing = await this.getByDocId(docId).catch(nullIfNotFound)
     if (existing) {
       return this.update(existing.versionId, value)
     } else {


RangerMauve:
Regarding the invitor setting the name/deviceType, I've added it as an optional param which also triggers the behaviour of the invitor writing the info and setting the flag in the invite.

src/member-api.js
```diff
@@ -221,6 +233,22 @@ export class MemberApi extends TypedEmitter {
           }
           await this.#roles.assignRole(deviceId, roleId)
 
+          if (peerInfo && peerInfo.deviceType && peerInfo.name) {
```
gmaclennan: Fix the type above and this can be just `if (peerInfo)`. Without it we still have a path where deviceInfo could neither be written by the invitor nor the invitee.

src/member-api.js
```diff
@@ -221,6 +233,22 @@ export class MemberApi extends TypedEmitter {
...
+              deviceType: deviceType,
+              selfHostedServerDetails: undefined,
+              schemaName: /** @type {const} */ ('deviceInfo'),
+            }
+            await this.#dataTypes.deviceInfo[kCreateWithDocId](deviceId, doc)
+          } else {
+            this.#l.log(
+              'ERROR: Could not assign peer device info due to lack of device info in RPC',
+              peerInfo
+            )
```
gmaclennan: Let's remove this as a "no error" path for when `invitorWroteDeviceInfo` is false, because in that case neither the invitor nor invitee would write the deviceInfo, and we will have a device on the project without a deviceId.

src/member-api.js
```diff
@@ -560,6 +581,28 @@ export class MemberApi extends TypedEmitter {
...
 
+  /**
+   * @param {string} deviceId
+   * @returns {Promise<DeviceInfo>}
+   */
+  async #getDeviceInfo(deviceId) {
+    try {
+      const deviceInfo = await this.#dataTypes.deviceInfo.getByDocId(deviceId)
+
+      return deviceInfo
```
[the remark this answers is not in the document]
RangerMauve: This caused a bug because not awaiting the promise first meant we returned right away and the catch never got triggered :P

PR #1236, opened 2026-03-05
chore: remove file extension from the errors export

RangerMauve:
Should be a bit cleaner and more standard

e.g.

```javascript
import { SomeError } from `@comapeo/core/errors`
```

instead of

```javascript
import { SomeError } from `@comapeo/core/errors.js`
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
       "types": "./dist/index.d.ts",
       "import": "./src/index.js"
     },
-    "./errors.js": {
+    "./errors": {
       "types": "./dist/errors.d.ts",
       "import": "./src/errors.js"
     }


achou11 (commented):
Personally I prefer having the file extension included and it seems like we do that for some of our other more recently worked-on modules.

Also worth noting that there can be a technical benefit to including the file extension (at least in Node): https://nodejs.org/api/packages.html#extensions-in-subpaths

> With [import maps](https://github.com/WICG/import-maps) now providing a standard for package resolution in browsers and other JavaScript runtimes, using the extensionless style can result in bloated import map definitions. Explicit file extensions can avoid this issue by enabling the import map to utilize a [packages folder mapping](https://github.com/WICG/import-maps#packages-via-trailing-slashes) to map multiple subpaths where possible instead of a separate map entry per package subpath export. This also mirrors the requirement of using [the full specifier path](https://nodejs.org/api/esm.html#mandatory-file-extensions) in relative and absolute import specifiers.

PR #1254, opened 2026-03-24
feat: Invite over internet

### LINKED ISSUE #1207: Invite over Internet
## Description

We need a way to add members to a project when they aren't on the same wifi network but have a means to send each other messages through whatsapp or email.

We will do this by making direct p2p connections over the internet using hyperswarm. We'll send the invite flow over those connections using the usual RPC methods

RangerMauve:
Closes #1207


src/discovery/remote-discovery.js
```diff
@@ -0,0 +1,113 @@
...
+   * Connect to another peer by their NOISE public key
+   * @param {string} publicKey
+   * @param {object} [opts]
+   * @param {number} [opts.timeout]
+   */
+  async connectPeer(publicKey, { timeout = 60_000 } = {}) {
+    const swarm = await this.#ensureSwarm()
+    const noisePublicKey = Buffer.from(publicKey, 'hex')
+
+    const onConnected = pEvent(this, 'connection', {
```
gmaclennan, line 244: Although currently unlikely, we should add defensive code for when the client is already connected to the peer when connectPeer is called.

src/mapeo-manager.js
```diff
@@ -285,6 +287,11 @@ export class MapeoManager extends TypedEmitter {
       logger,
     })
     this.#localDiscovery.on('connection', this.#replicate.bind(this))
+    this.#remoteDiscovery = new RemoteDiscovery({
+      identityKeypair: this.#keyManager.getIdentityKeypair(),
```
gmaclennan, line 344: I think we need to use a different keypair for hyperswarm connections, to avoid leaking identities (e.g. the "identity keypair" public key) to the public DHT swarm. This means that ultimately for over-the-internet sync we will need to store the "remote identity keypair" in the membership records or deviceInfo records in addition the existing deviceId, so that we can validate / authorize exchange with devices over the internet. We can add this later, but sooner-the-better because every device will need this updated membership record for internet exchange to work.

src/mapeo-manager.js
```diff
@@ -285,6 +287,11 @@ export class MapeoManager extends TypedEmitter {
...
+      logger,
+    })
+    this.#remoteDiscovery.on('connection', this.#replicate.bind(this))
```
gmaclennan, line 349: This is where we need an authentication guard on the server, validating that the incoming connection has the shared inviteId secret, but also with the manual PIN validation based on the connection handshake hash and verified out-of-band, because otherwise we are giving any incoming connection (which could come from anywhere) access to the RPC channel, which is not secure - it could send invites or other nefarious stuff.

src/member-api.js
```diff
@@ -85,90 +84,196 @@ const ACTIVE_ROLE_IDS = [CREATOR_ROLE_ID, MEMBER_ROLE_ID, COORDINATOR_ROLE_ID]
...
+  /**
+   * Start inviting somone over the internet. Returns a URL for the recipient to load.
+   * @param {InviteOptions} opts
+   */
+  async inviteOverInternet(opts) {
+    const inviteId = opts.__testOnlyInviteId || crypto.randomBytes(32)
+    const inviteIdString = inviteId.toString('hex')
+
+    const url =
+      INTERNET_INVITE_PAGE + `?i=${inviteIdString}&d=${this.#ownDeviceId}`
```
gmaclennan: used derived deviceId here (different keypair for hyperswarm connections)

gmaclennan: Store as `#` rather than query string to avoid passing to the server

gmaclennan: We're not sending project title, description and role, but I think that's ok if we are launching into the existing invite flow after connection and validation. So for the invitee the flow would be:

1. Open invite link -> opens app
2. Starts connecting immediately to invitor (or has a confirmation step)
3. Shows "connected" status with validation code. Waits until invitor confirms and sends invite (invitee would can ignore this screen if needed - it will be dismissed when the invitor sends the actual invite)
4. Invitor sends invite, flow as with local invites.
5.

src/member-api.js
```diff
@@ -85,90 +84,196 @@ const ACTIVE_ROLE_IDS = [CREATOR_ROLE_ID, MEMBER_ROLE_ID, COORDINATOR_ROLE_ID]
...
+
+    this.#pendingInvitesOverInternet.set(url, {
```
gmaclennan: Keying this by the `url` seems fragile. Maybe use the `inviteId` instead, since this is effectively what is used for the lookup anyway (the lookup currently constructs the URL deterministically from the `inviteId`).

gmaclennan (commented):
I left some comments on the approach. I'm not sure the division of responsiblities between the member-api and the remote-discovery class is quite right, but it depends on how you resolve the validation / authentication of incoming connections, which is the most important challenge to address.

src/mapeo-manager.js
RangerMauve: Mind elaborating on what the benefits are to not reusing the identity on the DHT?

src/mapeo-manager.js
RangerMauve: Mind elaborating on the security issues you expect? Inviting internet connected peers to other projects seems like a potential use case in the future.

src/mapeo-manager.js
gmaclennan: The public key is visible on the DHT. If it's the identity key pair / device id then anyone who learns of a device id can monitor the activity of the device online (eg when it connects and from which IP). Using a separate key pair makes it harder to associate the key with the device. An alternative is to be very careful about how we expose device ids outside the app, eg we could always hash them in exports (although we already have quite a lot of exports and data with device ids)

RangerMauve: > anyone who learns of a device id can monitor the activity of the device online

This assumes the DHT is compromised via a sybil attack and someone is trying to deanonymize participants, right?

Would avoiding advertising on our own device ID and sticking to topics derived from the invite ID help?

FWIW we could remove the deviceId from the URL and instead have a hash of the deviceId then filter the connection based on the hash of the remote public key to get the proper device Id.

gmaclennan: Actually I have limited understanding of the internal workings of the DHT. I was making an assumption that it's possible for devices that are part of the DHT to see the public keys of other devices that are on the DHT, but that may be incorrect.

src/mapeo-manager.js
gmaclennan: The attack vector I am concerned about:
1. I am a remote attacker
2. I portscan your device, find the open port, and initiate a noise connection.
3. I can now send RPC messages to your device like DoS invite to project - only the `invite-over-internet-redeemed` has a check on it.

Of course a similar vulnerability exists prior to this change from an attacker on the local network - our security review picked up on that - however this is a smaller and more acceptable risk.

RangerMauve:
I've got the connection and invite flow going. We want to make it harder to track specific peers via the DHT, therefore we want to use a fresh DHT keypair each time. The difficult part is that the keypair is used for the `deviceId` during the RPC API and that's tied with the member record. We'd need to either tie all members with an additional keypair, or somehow override the deviceID at the RPC layer, or something along those lines.

Gregor proposed we have something like this to send the *real* identity as the first packet down the noise stream after handshaking.

```
// Stable identity — persisted across sessions
const stableKeyPair = DHT.keyPair(someSeed)

// Ephemeral DHT presence — rotated each session
const swarm = new Hyperswarm({ keyPair: DHT.keyPair() })

swarm.join(topic, { server: true, client: true })

swarm.on('connection', async (conn, peerInfo) => {
  try {
    const remote = await handshakeIdentity(conn, stableKeyPair)
    console.log('Authenticated:', b4a.toString(remote.publicKey, 'hex'))
  } catch (err) {
    conn.destroy()
  }
})

function handshakeIdentity (conn, keyPair) {
  return new Promise((resolve, reject) => {
    // Sign the Noise handshake hash with our stable key
    const sig = b4a.allocUnsafe(64)
    sodium.crypto_sign_detached(sig, conn.handshakeHash, keyPair.secretKey)

    // Send stable public key + proof in a single message
    conn.write(encodeHandshake({
      publicKey: keyPair.publicKey,
      signature: sig
    }))

    conn.once('data', (data) => {
      const msg = decodeHandshake(data)

      const valid = sodium.crypto_sign_verify_detached(
        msg.signature,
        conn.handshakeHash,  // same hash on both sides
        msg.publicKey
      )

      if (!valid) return reject(new Error('Invalid identity proof'))
      resolve({ publicKey: msg.publicKey })
    })

    setTimeout(() => reject(new Error('Auth timeout')), 10000)
  })
}
```

This could leave the option to join by remote public key instead of a fresh topic.

We will also need some sort of access control step to have the invite ID before we expose the protomux channel for RPC.

src/mapeo-manager.js
RangerMauve: We rotate the DHT keypair whenever we restart core now. This should be a bit less fragile than constantly re-genning the key at the expense of being slightly more trackable.

RangerMauve:
Been wrestling with some sort of race condition for several hours now. I think I traced it down to the protomux channel recieving data before it's fully opened.

src/discovery/remote-discovery.js
```diff
@@ -0,0 +1,152 @@
...
+
+    // Send stable public key + proof in a single message
+    socket.write(
+      SwarmHandshake.encode({
+        publicKey: keyPair.publicKey,
+        signature: Buffer.from(sig),
+      }).finish()
+    )
+
+    const data = await firstData
```
gmaclennan: No guarantee this will come as a single chunk. Maybe needs to be length delimited or something? Or rely on it being a fixed length.

src/discovery/remote-discovery.js
```diff
@@ -0,0 +1,152 @@
...
+      socket.handshakeHash, // same hash on both sides
+      msg.publicKey
+    )
+
+    if (!valid) {
+      throw new Error('Invalid identity proof')
+    }
+
+    // @ts-ignore
+    socket.handshakePublicKey = msg.publicKey
```
gmaclennan: Could just override remotePublicKey to avoid changes in the rest of the code.

gmaclennan: Maybe a bit hacky

gmaclennan (commented):
I still think we need an authentication step before connecting to the RPC channel:

Validating the inviteID _and_ allowing the user (invitor) to confirm before connecting, which means queueing the connection somewhere until the user confirms

src/discovery/remote-discovery.js
RangerMauve: Length delimited is probably the safest option. uint32 to be safe?

RangerMauve:
Been blocked getting the tests to work with the handshake logic. Seems the connection is breaking when we try to write the public key down the wire and isn't draining.

RangerMauve:
Might do a flag in local-peers to ignore incoming RPC messages until we get authenticated? I think this extra handshake step is making everything more fragile.

RangerMauve:
Been trying a bunch of ways to rework this and I'm getting a bit stuck. Gonna write out some thoughts on approaches here.

## Initial approach:

- Invitor creates invite id
- Invitor starts swarm to listen on connections with same identity keypair (with deviceId ad publicKey) as local connections
- Invitor sends URL with own deviceId and the inviteId to invitee out of band
- Invitee takes deviceID and inviteId from URL
- Invitee starts swarm with own identity keypair
- Invitee calls "joinPeer" with invitor deviceId and waits for a connection to be established to them
- Invitee calls RPC API to send inviteID to the invitor deviceID to redeem the invite
- Meanwhile, the invitor was replicating all incoming swarm connections to localPeers and was waiting for an invite redeem call
- Invitor runs regular invite flow to the deviceID if the inviteID is valid
- Invitee auto-accepts invite if it is from the invitor deviceID
- Invitee becomes part of the project

This worked fine

## Ephemeral keys approach

- Invitor initializes a fresh ephemeral keypair for the swarm
- Invitor creates invite id
- Invitor starts swarm to listen on connections with swarm keypair
- Invitor sends URL with swarm public key and the inviteId to invitee out of band
- Invitee takes deviceID and inviteId from URL
- Invitee starts swarm with own ephemeral swarm keypair
- Invitee calls "joinPeer" with invitor swarm id and waits for a connection to be established to the ephemeral keypair
- On both sides we wait for the first "network packet" and fetch a proof for them owning a deviceId keypair (this is their identity key). This Id is used to identify them in local-peers instead of the remotePublic key of their ephemeral swarm keypair in the connection
- While waiting for this proof, we generate our own proof and send it down the wire
- Invitee calls RPC API to send inviteID to the invitor deviceID to redeem the invite
- Invitor runs regular invite flow
- Invitee auto-accepts invite if it is from the invitor deviceID
- Invitee becomes part of the project

This didn't work because the connection was "breaking" after sending the first packet. I think it's got something to do with how we're reading a packet of the stream before sending it off elsewhere, but it's been really difficult to debug.

**Edit: I got this working! I needed to pause the stream after getting the first packet, else we'd drop an event.**

## Queue up connections before RPC

In order to guard the RPC methods from spam, we should prompt the user to verify them before replicating anything. If we queue up the connections (at the manager level) we need a new place to track them that isn't local peers and would need to somehow tie them to the project they're trying to join via the member API. This extra book keeping and connection management is IMO not ideal. We'd likely need to add the invite ID into the handshake after establishing hyperswarm, which would also mean getting the remote discovery module to ask each project's `member-api` if this is a valid invite or not. Kinda leaky IMO

## Alternate: Disable RPC until connection is verified

Instead we could mark peers inside localPeers as being `untrusted` which limits which types of RPC events they are allowed to send. We could keep the invite redeem event open and give member-api a new `trustConnection(peerId)` method which would unlock the RPC channels. On the invitee side we can mark the connection as `trusted` right off the bat since we validate the remote ID. I think I'll go with this approach first since it's got the least amount of code changes and back-and-forths.

src/discovery/remote-discovery.js
RangerMauve: On the other hand I'm like 99% sure that the first packet will always be a single UDP packet passed through UDX

RangerMauve:
TODO: Test all the edge cases

RangerMauve:
TODO: Convert all `new Error` to proper Error classes inside `errors.js`

src/mapeo-manager.js
```diff
@@ -190,7 +196,8 @@ export class MapeoManager extends TypedEmitter {
     super()
     this.#keyManager = new KeyManager(rootKey)
     this.#deviceId = getDeviceId(this.#keyManager)
-    this.#defaultConfigPath = defaultConfigPath
+    ;(this.#swarmIdentityKeypair = KeyManager.generateProjectKeypair()),
+      (this.#defaultConfigPath = defaultConfigPath)
```
gmaclennan: Why the use of the comma operator here?

gmaclennan:
Good progress here. I think tracking peer state (`isTrusted`) makes sense vs. a waterfall approach where we try to authenticate first. Remember the requirement about user intervention to confirm a connection - do you have a plan for how to expose this in the API? 
As I understand the code so far, the ephemeral keypair used for swarm connection is generated on app startup, and the invite incorporates this and uses it to connect back to the invitor (e.g. not using the topic approach that we discussed). This makes sense to me, however two caveats:
1. On some devices, we may find that Android aggressively shuts down the app in the background, which means that the user experience will be that unless they keep the app open from the moment of sending the invite to the invitee responding, it may not work because the keypair will have rotated.
2. On other devices, we could find that the app remains open, potentially for multiple days, which creates a risk of tracking (stable public key on the DHT over multiple days), and equally we would have no way of controlling that.
One way of addressing this would be to deterministically generate the swarm keypair with the current day (local time), so we know it is stable within the same day, and will always change at the end of the day (e.g. midnight). I think that is easy to understand for users ("your invite is valid until midnight").

src/mapeo-manager.js
RangerMauve: I think I misclicked something in my editor that messed with the formatting.

RangerMauve:
Deterministic swarm key creation is a great idea!

RangerMauve:
I'll wait on the key rotation changes pending answers to these questions:

1. Is `identity-${year}-${month}-${day}` sufficient for the key name derivation?
2. If we want to be able to recover the swarm ID after reloading the app does that mean that we should be persisting invite over internet data to a DB so that we can reload it upon project load?

src/member-api.js
```diff
@@ -6,12 +6,10 @@ import { pEvent } from 'p-event'
 import { InviteResponse_Decision } from './generated/rpc.js'
 import {
   noop,
-  projectKeyToId,
   projectKeyToProjectInviteId,
   projectKeyToPublicId,
 } from './utils.js'
 import { Logger } from './logger.js'
-import { keyBy } from './lib/key-by.js'
```
RangerMauve, line 14: TODO: Shall we delete this util if it's no longer being used?

RangerMauve:
TODO: We should be able to reuse the invite after a failed  invite accept, keep the invite until we succeed

TODO: Track when invite was created and delete it if the swarm public key has changed since it was created

src/mapeo-manager.js
```diff
@@ -624,6 +643,22 @@ export class MapeoManager extends TypedEmitter {
           .where(eq(projectKeysTable.projectId, projectId))
           .get()?.projectInfo
       },
+      setShouldListenOverInternet: (shouldListen) => {
+        if (shouldListen) return this.#remoteDiscovery.start()
+        else return this.#remoteDiscovery.stop()
```
gmaclennan: Because this remoteDiscovery instance is shared by all projects, this could lead to some edge-case bugs if a user is switching between projects. The UI limitations _probably_ mean this will be ok in most circumstances, but down the line we could forget the restriction and make it easy for someone to have pending invites across multiple projects, and this will bite us. It's not that easy to fix - we need a ref count I think, plus we need to handle start() and stop() being async, and we should handle project.close() too. Maybe there is a different architecture that removes all the code needed for these edge-cases?

src/member-api.js
```diff
@@ -805,3 +1039,33 @@ async function parseAddServerResponse(response) {
...
+ *
+ * @param {string} inviteIdString
+ * @param {string} deviceId
+ * @returns {string}
+ */
+export function makeInviteURL(inviteIdString, deviceId) {
+  const url = INTERNET_INVITE_PAGE + `#i=${inviteIdString}&d=${deviceId}`
+
+  return url
+}
```
gmaclennan: I think maybe rename `deviceId` here and whereever it's used to `swarmPublicKey` - otherwise we a future maintainer could get in a muddle and try to use this as a deviceId.

src/member-api.js
```diff
@@ -85,90 +95,344 @@ const ACTIVE_ROLE_IDS = [CREATOR_ROLE_ID, MEMBER_ROLE_ID, COORDINATOR_ROLE_ID]
...
+  async #handlePeerRemove(peer) {
+    for (const [
+      pendingInviteId,
+      pendingInvite,
+    ] of this.#pendingInvitesOverInternet.entries()) {
+      if (pendingInvite.inviteeDeviceId === peer.deviceId) {
+        // TODO: Handle errors? Emit event saying this happened?
+        this.#cancelInviteOverInternetById(pendingInviteId).catch((e) => {
+          this.#l.log('Error: Unable to cancel invite', pendingInviteId, e)
+        })
```
gmaclennan: I'm not sure we should cancel the invite when the peer is removed? With unstable connectivity, a disconnect during an attempted invite connect would not be uncommon, and with the current code the reovery is to coordinate re-sending a new invite every time connections fail. We should allow retries.

src/pending-invites-api.js
```diff
@@ -0,0 +1,154 @@
...
+    if (!row) return undefined
+
+    if (!isRoleIdForNewInvite(row.roleId)) {
+      throw new Error(`Invalid roleId in database: ${row.roleId}`)
+    }
+
+    return /** @type {PendingInviteRecord} */ ({
+      ...row,
+      inviteeDeviceId: row.inviteeDeviceId ?? undefined,
+    })
```
gmaclennan: use `denullify()` here and remove the cast - `null` values will create subtle bugs.

src/pending-invites-api.js
```diff
@@ -0,0 +1,154 @@
...
+    const rows = this.#sql.getAll.all()
+
+    return rows.map((row) => {
+      if (!isRoleIdForNewInvite(row.roleId)) {
+        throw new Error(`Invalid roleId in database: ${row.roleId}`)
+      }
+      return /** @type {PendingInviteRecord} */ ({
+        ...row,
+        inviteeDeviceId: row.inviteeDeviceId ?? undefined,
+      })
```
gmaclennan: Use denullify - see above.

src/member-api.js
```diff
@@ -85,90 +95,344 @@ const ACTIVE_ROLE_IDS = [CREATOR_ROLE_ID, MEMBER_ROLE_ID, COORDINATOR_ROLE_ID]
...
+        inviteeDeviceId
+      )
+
+      // If they aren't connected after redeeming, we should mark it as cancelled
+      if (!stillConnected) {
+        await this.cancelInviteOverInternet(pendingInviteId)
+        throw new PeerDisconnectedSinceRedeemingInviteError()
+      }
+      if (!opts.roleName) delete opts.roleName
+      if (!opts.roleDescription) delete opts.roleDescription
```
gmaclennan: I think this is mainly a workaround for possible `null` values here - I think you can drop these once you `denullify()` the returned values, you won't have the bugs from this (the bugs happen because if roleName is `null`, the default destructuring in https://github.com/digidem/comapeo-core/blob/6c0122cc6b5839bf245430968da1e81b381cb4cc/src/member-api.js#L441-L442 doesn't work as expected)

src/member-api.js
```diff
@@ -85,90 +95,344 @@ const ACTIVE_ROLE_IDS = [CREATOR_ROLE_ID, MEMBER_ROLE_ID, COORDINATOR_ROLE_ID]
...
+        this.emit('internet-invite-cancelled', inviteId)
+      }
+      this.#pendingInvitesOverInternet.clear()
+      await this.#pendingInvitesApi.deleteAll()
+      await this.#setShouldListenOverInternet(false)
+      return
+    }
+    const { inviteIdString } = parseInviteURL(url)
+    await this.#cancelInviteOverInternetById(inviteIdString)
+  }
```
gmaclennan: It's an edge-case, but the ordering here can cause bugs - the event is emitted before the state is updated (client could read stale state after seeing event), and atomicity issues if there is a crash before the SQL is updated - the latter probably isn't an issue because of how it's handled in `_open`, but worth fixing maybe in case it bites us in a future code change? Ordering should be:

1. Persist to SQL
2. Mutate `this.#pendingInvitesOverInternet` (reflects SQL)
3. Update `setShouldListenOverInternet`
4. Emit events (listeners see consistent state in step 2)

src/discovery/remote-discovery.js
```diff
@@ -0,0 +1,326 @@
...
+  UnableToReadHandshakeError,
+} from '../errors.js'
+
+import { openedNoiseSecretStream } from '../lib/noise-secret-stream-helpers.js'
+
+/** @import {OpenedNoiseStream} from '../lib/noise-secret-stream-helpers.js' */
+/** @import {Keypair} from './local-discovery.js' */
+/** @import {Duplex, Readable} from "streamx" */
+
+/** @typedef {OpenedNoiseStream & {handshakePublicKey:Buffer, isTrusted: boolean}} RemoteAuthedNoiseStream */
```
gmaclennan: I think having two shapes of OpenedNoiseStream being passed around and needing to check for property presence (`'handshakePublicKey' in ...` and `'isTrusted' in ...`) and branch everywhere we use them is fragile for future maintainers. Better would be to have a single type `AuthedNoiseStream` emitted by both LocalDiscovery and RemoteDiscovery:

```
type AuthedNoiseStream = OpenedNoiseStream & {
  authenticatedPublicKey: Buffer  // the key downstream code should treat as deviceId
  isTrusted: boolean
}
```

test-e2e/invite-over-internet.js
```diff
@@ -0,0 +1,267 @@
...
+  assert.equal(reason, InviteResponse_Decision.ACCEPT)
+
+  const gotProjectId = await onInvited
+
+  assert.equal(gotProjectId, projectId, 'joined expected project')
+
+  // TODO: Test that initial sync happened
+})
+
+test.only('invite over internet, close, reopen, and join from URL', async (t) => {
```
gmaclennan: remove `.only()`?

src/discovery/remote-discovery.js
```diff
@@ -0,0 +1,326 @@
...
+ * @param {number} length
+ * @returns {Promise<Uint8Array>}
+ */
+async function readChunk(stream, length) {
+  let data = stream.read()
+
+  if (!data) {
+    try {
+      await pEvent(stream, 'readable', {
+        timeout: 10_00,
```
gmaclennan: Typo?
```suggestion
        timeout: 10_000,
```

src/member-api.js
```diff
@@ -85,90 +95,344 @@ const ACTIVE_ROLE_IDS = [CREATOR_ROLE_ID, MEMBER_ROLE_ID, COORDINATOR_ROLE_ID]
...
+  async #handleRedeemInviteOverInternet(peerId, { inviteId }) {
+    const inviteIdString = inviteId.toString('hex')
+    this.#l.log('Got incoming invite redeem', inviteIdString.slice(0, 7))
+
+    try {
+      for (const [
+        pendingInviteId,
+        pendingInvite,
+      ] of this.#pendingInvitesOverInternet.entries()) {
+        if (pendingInviteId !== inviteIdString) continue
```
gmaclennan: Do you think we need to use `timingSafeEqual` (and on the check below)? Probably not an actual exploitable issue, but we do use it everywhere else we check keys, so it's more a consistency issue than anything else.

test-e2e/invite-over-internet.js
```diff
@@ -0,0 +1,267 @@
...
+  // We should either take the first 4-8 bytes from the deviceID or derive something visual like emoji
+  const reason = await project.$member.acceptRedeemedInvite(inviteId)
+
+  assert.equal(reason, InviteResponse_Decision.ACCEPT)
+
+  const gotProjectId = await onInvited
+
+  assert.equal(gotProjectId, projectId, 'joined expected project')
+
+  // TODO: Test that initial sync happened
```
gmaclennan: This is important to address I think? How do you think we handle "invite should fail without initial sync" requirement? Could maybe be a separate PR? Otherwise we need to handle the "invited without initial sync" state.

src/local-peers.js
```diff
@@ -784,47 +869,52 @@ export class LocalPeers extends TypedEmitter {
     if (!peer) return // TODO: report error - this should not happen
     switch (type) {
```
gmaclennan, line 969: Which methods should we gate on `peer.isTrusted`? Maybe default to gating on that, and have exceptions if needed? I want to avoid accidentally opening security holes down the line.

gmaclennan (commented):
I think it could be good to switch the encoding of the keys on the URL to `base64url` - it would reduce the URL to within the 160-char SMS limit, and keeping the URLs shorter is nicer UX I think.

After sitting with this for a while, I've been reconsidering how we handle invite expiry. I think we should separate two concerns: expiring invites and rotating keys on the swarm. 

1) Rotating swarm identity: when creating an invite, check for existing non-expired invites. If none, generate a new keypair and perist it - reused while there are pending invites.
2) Separately add an `expiresAt` field to invites, and check it whenever reading invites. Add an "inviteExpired" error.

I think 24h is a good expiry - the current behaviour (I know I suggested it) I think is too confusing for users, and is strange across timezones. It's clearer to know "this invite is valid for 24h", and it's easier to change in the future.

src/mapeo-manager.js
RangerMauve: Architecture change: We track the pending invites at the manager level instead of projects. This then triggers the start stop behavior based on the global state.

src/local-peers.js
RangerMauve: I think we should gate everything but the device info and invite redeem

test-e2e/invite-over-internet.js
RangerMauve: I have a feeling the errors we had from this were related to the cache invalidation issue you fixed last week

src/local-peers.js
gmaclennan: And we should gate outgoing device info on trusted, so that an invitor does not just send its device info to any incoming connection.

RangerMauve:
Consideration:
Instead of rotating keys per day, we'll gate key rotation based on unexpired pending invites.

- For the first pending invite created, take not of the current time for a new `seedTimestamp` field.
- When creating subsequent invites, get the seedTimestamp from the oldest unexpired pending invite
- Use the oldest unexpired seedTimestamp for the swarm identity

This avoids us accidentally rotating the swarm key while there's still pending invites that depend on it since an invite link is static and has a hardcoded swarm public key

RangerMauve:
TODO: Check sender device ID when waiting for invite after redeeming invite code

src/discovery/remote-discovery.js
```diff
@@ -0,0 +1,326 @@
...
+    this.#identityKeypair = identityKeypair
+    this.#deriveSwarmIdentityKeypair = deriveSwarmIdentityKeypair
+  }
+
+  async #initSwarm() {
+    this.#l.log('Initializing swarm')
+
+    const swarm = new Hyperswarm({
+      keyPair: this.#deriveSwarmIdentityKeypair(),
+      maxPeers: 4,
```
gmaclennan: This seems arbitrary? What is the error when more than this number of peers attempt to connect? Is there a test for that?

src/discovery/remote-discovery.js
```diff
@@ -0,0 +1,326 @@
...
+      this.#swarm = await this.#loading
+    }
+    return this.#swarm
+  }
+
+  /**
+   * Start listening for incoming connections
+   */
+  async start() {
+    // TODO: Use start stop state machine
```
gmaclennan: Do you want to address this before merging?

src/discovery/remote-discovery.js
```diff
@@ -0,0 +1,326 @@
...
+  return b4a.concat([data, remainingData])
+}
+
+/**
+ *
+ * @param {Readable|Duplex} stream
+ * @returns {Promise<Buffer>}
+ */
+export async function readHandshakeBuffer(stream) {
+  const handshakeLengthBytes = await readChunk(stream, LENGTH_BYTES_LENGTH)
```
gmaclennan, line 377: Should have a max length here, so stop a connecting peer causing an OOM error.

src/invite/invite-links-api.js
```diff
@@ -0,0 +1,406 @@
...
+    if (!oldest) return
+    const expiresAt = oldest.createdAt + this.#expiryMs
+    const delay = Math.max(0, expiresAt - Date.now())
+    this.#clearExpiredTimer = setTimeout(async () => {
+      this.#clearExpiredTimer = null
+      await this.#clearExpired()
+      this.#scheduleExpired()
+    }, delay)
+    this.#clearExpiredTimer.unref()
+  }
```
gmaclennan, line 217: Is this necessary? If we're using `createdAt + expiryMs` as the source of truth of whether an invite has expired, rather than presence of the row in the table, can we not just clear lazily?

src/invite/invite-links-api.js
```diff
@@ -0,0 +1,406 @@
...
+          sql`${pendingInvitesTable.createdAt} < ${sql.placeholder('cutoff')}`
+        )
+        .prepare(),
+      getOldest: db
+        .select()
+        .from(pendingInvitesTable)
+        .orderBy(sql`${pendingInvitesTable.createdAt} ASC`)
+        .limit(1)
+        .prepare(),
+    }
```
gmaclennan, line 162: Right now expiry seems to be managed by "is the invite in the table", and there is a timer to periodically delete them. This seems fragile, with the possibility that an expired invite could be returned, and slightly confusing behaviour of the expiry of invites changing when the class is reinitialized. 

How about adding an explicit `expiresAt`? That would also allow a distinction between expired and canceled invites. I think it would be worth sharing the expiresAt as part of the invite link, so that an invitee can short-circuit the connection and immediately show that a link is expired?

gmaclennan (commented):
In no particular order: 
 
1. Is there a clean way to gate RPC incoming _and_ outgoing RPC on `isTrusted` by default, and then explicitly mark RPC messages which are accepted/sent to untrusted peers? I feel like the current "opt-in gating" creates a risk of forgetting to add it when adding a new RPC method and accidentally opening a security loophole.
2. It seems like an invitor would send their deviceInfo to any device that connects to them, whether trusted or untrusted. That seems like a bit too much info to leak to any device that connects, can we only send that once a device is trusted?
3. The current architecture makes it quite hard to trace the flow of "untrusted --> trusted" because you have `#markPeerAsTrusted` being passed around and the flow for how a device becomes trusted I think takes place in multiple places. Is there a way to make this cleaner / easier to read? I feel like there is a lot of mental overhead thinking through where we need to be careful about handling an untrusted peer.
4. The designs now expose an invite ID to users, as a way to help users track multiple invites that they have sent out. Right now our inviteId is random 32-bytes hex-encoded, which seems like excessive entropy (except we don't rate-limit - see below). I think this should be our first step for standardizing how IDs are exposed (see [this old thread on IDs](https://github.com/digidem/mapeo-core/issues/104)) and we should expose the inviteId as [Crockford base-32](https://en.wikipedia.org/wiki/Base32#Crockford's_Base32)-encoded (it would be easier to use zbase-32 like hypercore, but the leniency of Crockford is much better for future work around observation IDs which may be re-typed by users - and I think upper-case is easier to distinguish internationally). If we rate-limit to 1 invite per second and target P (probability of an attacker guessing an inviteId) of 2^-32 (~1 in 4 billion), and `V` concurrent invites maxing at, say, 1,000, the number of bits of entropy we need are ~ `log2(1 x 86,400 x 1,000) + log2(1/P) = log2(86,400,000) + 32 = 58` - I think 64 bits should cover us - 13 chars in base 32.
5. I think we should have some kind of rate-limiting on incoming connections, now that we are exposing comapeo to the internet. Maybe a global-rate limit and restricting concurrent connections from the same IP? Rate limit just by delaying the handshake? What do you think is best and easy? I think it's good to start from a place of "fairly secure" when we're opening up this new potential attack surface.
6. Just to document explicitly what was discussed in Slack - designs now ask for invite links to be re-usable within their validity period.
7. Maybe for naming, and consistency with front-end, can we call them "invite links" - "pending invites" seems confusing - and the structure is built around the idea that it is an actual invite "just waiting for a deviceId", rather than a "let this someone connect to me with this code and I'll decide to send an invite to them". 
8. I think it would be good to include an expiry (`e=123456`) in the URL so someone clicking the link can avoid the work trying to connect and getting denied. We want to keep the URL short to maybe seconds since unix epoc instead of milliseconds? Our own epoch? Encoded in base32/64? Don't want to over-complicate this, but also huge URLs are a bad UX.
9. For swarmPublicKey encoding in the URL, stick to hex? or Crockford base32 or base64url? What do you think? It's not designed to be human-readable, so not the same constraint as the inviteID, but maybe consistency is best.
10. How can we expose errors to the invitee? I think the failures are connection failure, invite expired, and invite cancelled. Any others I'm missing? How can an invitee know which it is? I think right now the code just disconnects a client if the inviteId can't be matched - can the invitee know from that whether they should retry (connection failure) vs. permanent failure? 
11. A few items from my previous review still pending I think, around rotating swarm identity and how we discriminate trusted/untrusted on the noise stream object we pass around.

gmaclennan (commented):
Sorry a few more comments after re-reviewing the UX designs (from https://www.notion.so/digidem/Invite-over-the-internet-2e61b08162d580e9a861c3bf93c58d70?source=copy_link):

1. Can we encode the project name and role name in the URL - the designs need that so an invitee can see what they are signing up to. Down the line users might request more info - right now the designs have connect -> accept invite as an automated flow, but we could consider in the future I guess having an additional step so we can share the extra info over the connection rather than encoded in the URL.
2. The designs require showing the invitee name on an incoming invite redeem event.
3. There doesn't seem to be a way for the user to deny an incoming invite request right now?

RangerMauve:
- Invite id b32, add at the beginning to make it easier to identify
- Need to show expiry in the listing pending invites (expiresAt)
- Rename pending invites to "invite links" (in general review the names of everything, events, review wording in notion)

src/member-api.js
RangerMauve: Logic changed after recent refactor,

src/discovery/remote-discovery.js
RangerMauve: What's a good value? Maybe 8x what our current expected chunk size is? In case there's more keys to add in the future for some reason?

src/invite/invite-links-api.js
RangerMauve: Thinking for active clearing is so we can notify the ui to update when it expires without needing to re-navigate. Also avoids needing to do the extra clear every time we list or update.

gmaclennan: Yes, and, the UI will be re-rendering anyway to show time remaining (without re querying the backend) and won't need to attach event listeners for expiring invites. But could be useful for some front end devs maybe? Also remember that SQLite queries are so cheap - I think in most cases almost insignificant. But also fine to keep the active clearing!

src/discovery/remote-discovery.js
gmaclennan: I think anything that is not a big memory risk is fine. Maybe 64Kb? We could enforce this by using 2-bytes to encode the length. I had been wondering about whether we should follow some consistency for length-prefixing - most of our messages are via protomux, which uses a varint (via compact-encoding uint), and the noise handshake (in @hyperswarm/secret-stream) uses uint24le. So no real consistency - we could decide to use uint24le to keep handshake messages using the same length prefix, but also fine I think to just decide to use uint16le for our handshake.

gmaclennan:
Some  notes about linking:

- Let's use the link `https://a.comapeo.app/invite` (no trailing slash - no real standard, but loose convention is "no slash" = resource, "slash" = list) where "a" is for "app". This will be the domain for all future deep links.
- I've already set up a CNAME for `a.comapeo.app` pointing at `digidem.github.io`
- On mobile, once registered, clicking the link will directly open the app, without going via the browser.
- On desktop, we need to open the deep link with the `comapeo:` schema. We should do this via both a `window.location=` redirect, and an "open in the app" button that the user can click (sometimes the redirect will not work)
- The deep link should be `comapeo:///invite#i=ABC123` - note the 3 slashes `///` - this gives parsing parity with the https URL.
- Desktop will need to jump through a few hoops to implement this: on windows and linux it will open a second window/instance, so the app will need to force a single instance and forward the URL to the existing open window. On Macos it's an event, but it could fire before the app has fully started, so it needs to be cached and forwarded once everything has started up.

docs/guides/invite-over-internet.md
```diff
@@ -0,0 +1,96 @@
...
+// Show Waiting to connect screen until we get the connected event or react to the errors
+// Once resolved show joined screen
+const projectId = await manager.joinProjectFromLink(url)
+// InviteRedeemConnectionClosedError => Connection error page
+// UnknownInviteIDError => Invite has expried page
+// InviteDeniedByInviterError => Request denied page
+
+await manager.cancelJoinProjectFromLink(url)
+// InvalidInternetInviteURLError
+```
```
gmaclennan: I think this API as-is is a little difficult to work with for a front-end implementation.

As I understand it, calling `joinProjectFromLink(url)` essentially creates a "join project from link task" which is a state machine that goes through these states:

1. Attempting to connect
2. Connected, awaiting acceptance
3. accepted, awaiting invite
4. Invited, awaiting initial sync
5. joined

The issue with the API as-it-is is that the final state is inferred by the promise resolving, and there is no way to list "active join project from link tasks", so the front-end needs to track that state via a long-running promise. Plus, although we don't currently expose a state for (4), with the current API it means adding a new event for each state.

I _think_ an easier API to work with would be to introduce a concept like a "joinRequest" with an associated state. I think it should also be namespaced to the `manager.invite` API, like the other invite actions right now, or maybe in its own namespace like `manager.joinRequests`, to avoid needing to change the existing API which currently is not really named to "share" with other methods (e.g. `invite.getById()`).

```js
manager.invite.on('join-request-update', ({ status, inviteId, error }) => {})

// @type {ReturnType<typeof parseInviteURL>}
const joinRequest = await manager.invite.createJoinRequest(url)

// Possibly throws with ErrorExistingJoinRequest()

const joinRequest = await manager.invite.getJoinRequestById(inviteId) // or by URL?? `inviteId` seems like the narrower target

const joinRequests = await manager.invite.getJoinRequests() // useful in case front-end process restarts - state maintained by backend

await manager.invite.cancelJoinRequest(inviteId)
```

src/discovery/remote-discovery.js
gmaclennan: still todo?

src/discovery/remote-discovery.js
```diff
@@ -0,0 +1,351 @@
...
+   * @param {AbortSignal} [opts.signal]
+   * @returns {Promise<RemoteAuthedNoiseStream>}
+   */
+  async connectPeer(publicKey, { timeout = 60_000, signal } = {}) {
+    const swarm = await this.#ensureSwarm()
+    const noisePublicKey = Buffer.from(publicKey, 'hex')
+
+    for (const existingConnection of this.#connections) {
+      if (existingConnection.remotePublicKey?.equals(noisePublicKey)) {
+        const opened = await openedNoiseSecretStream(existingConnection)
```
gmaclennan: possible race condition: By the time this has looped through all the existing connections, a new existing connection could have been established

src/discovery/remote-discovery.js
```diff
@@ -0,0 +1,351 @@
...
+        if (opened.destroyed) break
+        // @ts-ignore Some connections might not be handshaked, wait for them to be
+        if (!existingConnection.handshakePublicKey) break
```
gmaclennan: where does the wait for the handshake happen? it seems like this falls through to establish a new connection?

src/discovery/remote-discovery.js
gmaclennan: still unresolved I think?

gmaclennan: Should have a test for an oversized handshake too.

src/local-peers.js
```diff
@@ -782,52 +934,89 @@ export class LocalPeers extends TypedEmitter {
...
+    if (
+      !peer.isTrusted &&
+      !ALLOWED_UNTRUSTED_RPC.has(type) &&
+      !type.endsWith('Ack')
+    ) {
+      throw new UntrustedRPCMethodError({
+        type,
+        peerId: peer.id,
+      })
+    }
```
gmaclennan: should we disconnect here? Because we have `maxPeers: 16`, a theoretical untrusted peer could hang around indefinitely probing RPC methods. I'm happy for this to be a follow up task - just add it as an issue so we don't forget, I think it's a relatively minor security threat right now.

src/local-peers.js
```diff
@@ -782,52 +934,89 @@ export class LocalPeers extends TypedEmitter {
     const peer = this.#getPeerByProtomux(protomux)
     /* c8 ignore next */
     if (!peer) return // TODO: report error - this should not happen
+    // If the peer isn't trusted, ignore anything not allowed
+    // Allow acknowledge messages by default
+    if (
+      !peer.isTrusted &&
+      !ALLOWED_UNTRUSTED_RPC.has(type) &&
+      !type.endsWith('Ack')
```
gmaclennan: Maybe too broad? keep to only RedeemInviteOverInternetAck/DenyInviteOverInternetAck

src/discovery/remote-discovery.js
```diff
@@ -0,0 +1,351 @@
...
+    this.#deriveSwarmIdentityKeypair = deriveSwarmIdentityKeypair
+    this.#swarmOpts = swarmOpts
+  }
+
+  async #initSwarm() {
+    this.#l.log('Initializing swarm')
+
+    const swarm = new Hyperswarm({
+      keyPair: this.#deriveSwarmIdentityKeypair(),
+      maxPeers: 16,
```
gmaclennan, line 106: should we test this maxPeers? What happens if it's exceeded? Would be rare, e.g. workshop-based setup where invites are sent out to multiple invitees at the same time. I think it's fine to restrict this, just want to understand what the consequence is if it is exceeded, particularly for the "spamming" case (someone on the swarm, attempting multiple connections opportunistically - we block them doing anything, but they would be connected)

src/discovery/remote-discovery.js
```diff
@@ -0,0 +1,351 @@
...
+
+      return socket
+    } catch (e) {
+      if (e instanceof EventTimeoutError) {
+        throw new TimeoutError('Timed out waiting for peer')
+      }
+      throw e
+    } finally {
+      signal?.removeEventListener('abort', onAbort)
+      this.#shouldTrustKeys.delete(publicKey)
```
gmaclennan, line 267: I _think_ in this case the connection is left in `#connections`, but as untrusted. When a peer re-connects, isn't there a risk of the return at line #177 returning the connection that is untrusted, so it just gets locked? I think the broader issue is maybe the lifecycle - there's no swarm.leavePeer() except in the abort case. Sorry - trying to wrap this up before our team call so not as clear as I would like to be, but hopefully this is enough to go on?

src/mapeo-manager.js
```diff
@@ -329,10 +402,13 @@ export class MapeoManager extends TypedEmitter {
   }
 
   /**
-   * @param {NoiseSecretStream<any>} noiseStream
+   * @param {NoiseSecretStream<any>|RemoteAuthedNoiseStream} noiseStream
    */
   #replicate(noiseStream) {
-    const replicationStream = this.#localPeers.connect(noiseStream)
+    const isTrusted = `isTrusted` in noiseStream ? noiseStream.isTrusted : true
```
gmaclennan: We should use our new type `RemoteAuthedNoiseStream` - right now this is fragile for future maintenance - `isTrusted` defaults to `true`, so there is implied knowledge that you must only call #replicate with a trusted stream.

Suggestion: define the shared type in the noise strem helpers:

`/** @typedef {OpenedNoiseStream & { authenticatedPublicKey: Buffer, isTrusted: boolean }} AuthedNoiseStream */`

Then in LocalDiscovery, set `socket.authenticatedPublicKey = socket.remotePublicKey`, and `isTrusted=true`. 

then you can simplify the checks in `local-peers.js:1248-1257`, `mapeo-manager.js:408` and `remote-discover.js:140-144`. It avoids a future maintainer accidentially passing a stream through connect() and defaulting to trusted.

gmaclennan (changes requested):
I see there are some checks that are failing that need to pass before we can merge. Also, some API feedback and some comments that were not resolved from the last review pass, and some new comments.

docs/guides/invite-over-internet.md
RangerMauve: @achou11 @ErikSin Thoughts on this?

gmaclennan: In the react code this would be wrapped in a hook eg `const joinRequest = useJoinRequest(inviteId)` it's generally easier if the main frontend state to manage is just a screen parameter (like inviteId / joinRequestId). I haven't fully thought through what the hook wrapper would look like - Andrew maybe if you did a quick prototype, the best shape will fall out of that work.

src/discovery/remote-discovery.js
RangerMauve: one issue I have is how long we are willing to wait for the connections to settle before we assume the current set is "ready"

The specific race here is in getting duplicate connections while we wait for the current connection from the peer to resolve. Since this is UDP sockets and hyperswarm already has some dedupe happening I feel like it's an unlikely scenario.

As we iterate the only async step is waiting for the stream from a specific peer to be marked as opened or destroyed. I think just rechecking the existing connections for that peer if it's destroyed should suffice.

src/local-peers.js
RangerMauve: I think we should keep the connection in case we have some innocuous rpc we want to send on initial connection in the future. A counter to hanging indefinately is to disconnect after some time if the are still not trusted. I'm thinking a few seconds? that's in the todo already

src/local-peers.js
RangerMauve: What are the risks you forsee with allowing all acks? I was thinking it'd make adding more methods easier down the line if we focus on gating just the non-ack methods

src/discovery/remote-discovery.js
RangerMauve: Actually, it seems to limit outgoing connections, not incoming.

https://github.com/holepunchto/hyperswarm/blob/48d4241f69f848bb4c9dccb65b6fce5a9f40009d/index.js#L300

src/discovery/remote-discovery.js
RangerMauve: Yeah, it waits for the onConnection handler to happen. The extra joinPeer is a noop

src/discovery/remote-discovery.js
RangerMauve: what I mean is that if we get spammed with connections, once we reach 16 we won't be able to make new outgoing connections to join a link. Pruning the untrusted connections (and maybe banning them?) should reduce the risks further

src/discovery/remote-discovery.js
RangerMauve: Now using 16 bit and with a max message that makes the entire packet fit inside MAX_UINT16

src/invite/invite-links-api.js
RangerMauve: Lets keep this as is

src/invite/invite-links-api.js
RangerMauve: Added

src/local-peers.js
RangerMauve: I think it's fine to ignore them and add them to the allow list as needed.

docs/guides/invite-over-internet.md
RangerMauve: Gonna place it under a new `.inviteLinks` key instead to keep the logic more separate. But I guess I could merge it into one if it'd make a big difference on the consumer side?

src/discovery/remote-discovery.js
```diff
@@ -0,0 +1,373 @@
...
+      }
+
+      // @ts-ignore
+      socket.handshakePublicKey = msg.publicKey
+      // @ts-ignore
+      this.emit('connection', socket)
+      this.#pendingHandshakes.delete(socket)
+      pendingDefer.resolve(true)
+    } catch (err) {
+      this.emit('error', ensureKnownError(err))
```
gmaclennan, line 320: I don't think you attach an error listener to the remoteDiscovery instance, so this will throw an unhandled 'error' and crash node.

src/discovery/remote-discovery.js
gmaclennan: Have you maybe not pushed this change? The code I see is still unbounded length.

src/discovery/remote-discovery.js
```diff
@@ -0,0 +1,373 @@
...
+  async #findExistingPeer(noisePublicKey) {
+    for (const existingConnection of this.#connections) {
+      if (!existingConnection.remotePublicKey?.equals(noisePublicKey)) continue
+      const opened = await openedNoiseSecretStream(existingConnection)
+      // If the connection closed, try again
+      if (opened.destroyed) return this.#findExistingPeer(noisePublicKey)
+      // @ts-ignore Some connections might not be handshaked, wait for them to be
+      if (!existingConnection.handshakePublicKey) {
+        const success = await this.#pendingHandshakes.get(existingConnection)
+        if (!success) return this.#findExistingPeer(noisePublicKey)
```
gmaclennan: This recurses - a failed handshake on line 279 does `pendingDefer.resolve(false)` but does not close the socket, so a subsequent connection to the same `noisePublicKey` will hit this path with `success=false` and just loop.

src/local-peers.js
```diff
@@ -518,6 +623,7 @@ export class LocalPeers extends TypedEmitter {
     if (!peer.supportsMapShare()) {
       throw new MapShareNotSupportedByPeerError()
     }
+    checkTrusted('MapShareExtension', peer)
```
gmaclennan, line 638: we should centralize the check for sending, like we do on the receive path. Right now, any future addition of send methods needs to remember to add checkTrusted() and could easily forget. Right now it's unclear that `sendDenyInviteOverInternet/sendRedeemInviteOverInternet` omit sendTrusted() intentionally. Better to have it block-by-default like the receive side so we're explicit about this.

src/local-peers.js
gmaclennan: I don't know what the risks are, but it seems better to avoid leaving this hole open because we don't know what exploits it could open up down the line.

src/discovery/remote-discovery.js
```diff
@@ -0,0 +1,373 @@
...
+    this.#identityKeypair = identityKeypair
+    this.#deriveSwarmIdentityKeypair = deriveSwarmIdentityKeypair
+    this.#swarmOpts = swarmOpts
+  }
+
+  async #initSwarm() {
+    this.#l.log('Initializing swarm')
+
+    const swarm = new Hyperswarm({
+      keyPair: this.#deriveSwarmIdentityKeypair(),
```
gmaclennan: Hyperswarm keeps this keypair for the lifespan of this class, but during that lifespan all invites could expire, and a new invite link could be created, which would advertise a different key to the one the hyperswarm is listening on, so the link would not work / not find the invitor.

src/discovery/remote-discovery.js
```diff
@@ -0,0 +1,373 @@
...
+    await swarm.listen()
+    this.#l.log('Listening')
+    return swarm
+  }
+
+  /**
+   * @returns {Promise<Hyperswarm>}
+   */
+  async #ensureSwarm() {
+    await this.#loading
```
gmaclennan: There's a small microtask race condition edge-case here: two callers to `#ensureSwarm()` in the same tick would both see `this.#swarm` as null and init a swarm, and we would onlyh keep a reference to the second, but the first would be active and listening but never destroyed.

src/local-peers.js
```diff
@@ -782,52 +934,89 @@ export class LocalPeers extends TypedEmitter {
...
         const inviteResponse = parseInviteResponse(value)
-        const peerId = keyToId(protomux.stream.remotePublicKey)
-        this.emit('invite-response', peerId, inviteResponse)
-        peer.sendInviteResponseAck(inviteResponse).catch((e) => {
-          this.#l.log(`Error sending invite response ack ${e.stack}`)
-        })
+        peer
+          .sendInviteResponseAck(inviteResponse)
+          .then(() => {
+            this.emit('invite-response', peer.id, inviteResponse)
```
gmaclennan, line 1010: Should this be in a `finally()`? (similarly to the other cases which change this) - the connection could be dropped before the ACK is drained, which result in the `catch()`, but the failure is in the sending the Ack, not receiving the response.

src/member-api.js
```diff
@@ -654,12 +909,7 @@ export class MemberApi extends TypedEmitter {
         const memberInfo = { deviceId, role }
 
         try {
-          const configCoreId = await this.#coreOwnership.getCoreId(
-            deviceId,
-            'config'
-          )
-
-          const deviceInfo = deviceInfoByConfigCoreId.get(configCoreId)
+          const deviceInfo = await this.#getDeviceInfo(deviceId)
```
gmaclennan: This changes behaviour subtly - getMany() will throw if the device has not set its own device info yet. The change in commit https://github.com/digidem/comapeo-core/pull/1254/changes/ad82754f5984967c1316ee1a4c56d23714e078e8 paper over this for the server situation, but could this be a problem in other situations? It introduces a new "hidden rule" of "you must set your device info before you call getMany()".

src/member-api.js
```diff
@@ -80,103 +97,360 @@ const ACTIVE_ROLE_IDS = [CREATOR_ROLE_ID, MEMBER_ROLE_ID, COORDINATOR_ROLE_ID]
...
+    }))
+  }
+
+  /**
+   * When a peer disconnects, remove them from the redeemed invites set.
+   * @param {PeerInfoDisconnected} peer
+   */
+  #handlePeerRemove = async (peer) => {
+    if (!peer) return
+    for (const deviceIds of this.#redeemedInvites.values()) {
```
gmaclennan, line 312: I don't think there is a way that a device could redeem two invites, so I think this is fine, but just noting that it would leave the second redeemed invite hanging, which would block a retry for redeeming the second invite.

src/member-api.js
```diff
@@ -80,103 +97,360 @@ const ACTIVE_ROLE_IDS = [CREATOR_ROLE_ID, MEMBER_ROLE_ID, COORDINATOR_ROLE_ID]
...
+   */
+  async denyInviteLinkRequest(
+    inviteId,
+    deviceId,
+    reason = DenyInviteOverInternet_DenyReason.invitor_denied
+  ) {
+    const redeemedSet = this.#redeemedInvites.get(inviteId)
+    if (!redeemedSet || !redeemedSet.has(deviceId)) {
+      throw new InviteNotYetRedeemedError()
+    }
```
gmaclennan, line 399: When denyInviteLinkRequest() is called from line 336 for an invalid invite, `redeemedSet` will be falsy, so kHandleRedeemInviteOverInternet will throw a misleading error (InviteNotYetRedeemedError) instead of the intended UnknownInviteIDRedeemAttemptError. However also the branch at 336 I think just duplicates the code in mapeo-manager.js:1365-1379 - would it make more sense to pass the invite object to kHandleRedeemInviteOverInternet() so you don't need to repeat logic?

src/mapeo-manager.js
```diff
@@ -701,6 +791,100 @@ export class MapeoManager extends TypedEmitter {
...
+   * @param {number} [options.timeout]
+   * @returns {Promise<string>}
+   */
+  async joinProjectFromLink(url, { timeout = 60_000 } = {}) {
+    const { swarmPublicKey, inviteIdString } = parseInviteURL(url)
+    const inviteId = Buffer.from(inviteIdString, 'hex')
+
+    const redeemAbortController = new AbortController()
+    const signal = redeemAbortController.signal
+    this.#outboundRedeemInvites.set(inviteIdString, redeemAbortController)
```
gmaclennan: this is cleaned up in the `finally` block below, but `connectPeer()` could throw before/outside that try block, so this would never be cleaned up. This would mean a subsequent cancelJoinProjectFromLink(url) would find the stale controller and resolve silently, rather than throwing InvalidInternetInviteURLError

src/mapeo-manager.js
```diff
@@ -701,6 +791,100 @@ export class MapeoManager extends TypedEmitter {
...
 
+  /**
+   * Attempt to join a project over the internet
+   * @param {string} url
+   * @param {object} [options]
+   * @param {number} [options.timeout]
+   * @returns {Promise<string>}
+   */
+  async joinProjectFromLink(url, { timeout = 60_000 } = {}) {
+    const { swarmPublicKey, inviteIdString } = parseInviteURL(url)
```
gmaclennan: probably on your list already: read expiresAt and fail early.

src/invite/invite-links-api.js
```diff
@@ -0,0 +1,377 @@
...
+
+  /**
+   * Check if we should be listening over the internet based on
+   * total invite count across all projects.
+   * Only calls the callback when transitioning from 0 to 1 (true)
+   * or from 1 to 0 (false).
+   * @param {boolean} direction - true if we just added, false if we just removed
+   */
+  async #checkSetShouldListenOverInternet(direction) {
+    const count = this.#sql.getAll.all().length
```
gmaclennan, line 227: two concurrent create calls would result in `count === 2` here, so this.#setShouldListenOverInternet(true) would never be set. Would `count > 0` suffice? `this.#setShouldListenOverInternet(true)` is idempotent so ok to call it twice.

src/invite/invite-links-api.js
```diff
@@ -0,0 +1,377 @@
...
+  /**
+   * Delete all invite links whose createdAt timestamp is older than 24 hours.
+   */
+  async #clearExpired() {
+    const cutoff = Date.now()
+    const expired = this.#sql.getExpired.all({ cutoff })
+    for (const row of expired) {
+      await this.#db
+        .delete(inviteLinksTable)
+        .where(eq(inviteLinksTable.inviteId, row.inviteId))
```
gmaclennan: Nit: This could be simplified to a single `DELETE ... WHERE expiresAt < cutoff` rather than a delete for every row

src/invite/invite-links-api.js
```diff
@@ -0,0 +1,377 @@
...
+    }
+    this.#scheduleExpired()
+  }
+
+  async _close() {
+    this.#cancelScheduleExpired()
+  }
+
+  /**
+   * Delete all invite links whose createdAt timestamp is older than 24 hours.
```
gmaclennan: update comment (expiredAt)

src/invite/invite-links-api.js
```diff
@@ -0,0 +1,377 @@
...
+   * Create a new invite link record
+   * @param {Omit<InviteLinkCreate, 'projectId'>} data
+   * @returns {Promise<void>}
+   */
+  async create(data) {
+    await this.#inviteLinksApi.ready()
+    return this.#inviteLinksApi.create({
+      ...data,
+      projectId: this.#projectId,
+    })
```
gmaclennan: nit: double await ready() for these wrapper methods - the inviteLinksApi methods already await in their body.

gmaclennan (changes requested):
A few edge-cases and remaining security issues to clean up

src/discovery/remote-discovery.js
RangerMauve: Right, we should do a full teardown when we stop or track when the key expirec

docs/guides/invite-over-internet.md
RangerMauve: Updated API

src/discovery/remote-discovery.js
RangerMauve: This should be fixed by the StartStopStateMachine

src/discovery/remote-discovery.js
RangerMauve: Instead I'll check if the key changed since we last launched. Should be better for perf.

src/local-peers.js
RangerMauve: Plugged BTW

src/member-api.js
RangerMauve: I don't think they could brute force it within the time it takes to expire.

src/discovery/remote-discovery.js
RangerMauve: We end the socket on fail

src/discovery/remote-discovery.js
RangerMauve: Handled via #findExistingPeer

src/discovery/remote-discovery.js
RangerMauve: Right, we shouldLeavePeer on error.

src/discovery/remote-discovery.js
RangerMauve: Shopuld be there, the max size is now 2 bytes, also added a test for it: https://github.com/digidem/comapeo-core/pull/1254/changes#diff-5c1257d2e3ee44cf984fbd6e19e2127fce3d5989f5c80cbbfb85ab93a4c46775R241

src/local-peers.js
RangerMauve: Not sure what the API should look like for this, where could the guard go if not in the method? Have all the methods call checkTrusted and do an allow list for the `type`?

RangerMauve: doing so

src/local-peers.js
RangerMauve: Hard to say because if we can't send the ack then likely the sender will see the RPC as failed so we'd have diverging state.

src/member-api.js
RangerMauve: Good point we should remove from all of them.

src/member-api.js
RangerMauve: Simplified

src/member-api.js
RangerMauve: catching the error so the behavior is unchanged

gmaclennan:
Sorry to come late with some additional architecture comments, but I've been thinking through more details about privacy and security related to connections over the internet, and how we manage future internet sync. I think it merits some changes in this PR before we roll it out because it reduces changes that are needed in the future.

**Block project replication based on trust**
Right now, when an invitee connects, it shares the discovery keys of all the projects it is a member of with the invitor, and vice-versa. This unnecessarily leaks identifiable information - it leaks to anyone who connects, before even validating their inviteId (the project creator core replicates into the stream on `peer-add`).

**Fix:** `localPeers.on('peer-add', onPeerAdd)` shouldn't replicate to non-trusted (e.g. non-local) peers, and the `discovery-key` event should be ignored. An invitee does need, however, to replicate the project they are invited to.

**Test:** during invite, with either/both invitee and invitor with multiple projects, no `hypercore/alpha` channel should open before invite accept, and after accept it should only be the joined project.

**New protomux auth channel**
Currently the auth is a hand-rolled `readHandshakeBuffer` message that runs before Protomux attaches, with no version negotiation. For future internet sync, and more private local sync, we need a channel to negotiate membership of every project to avoid revealing stable identifiers (e.g. project discovery key) to anyone who connects. If we move the identify proofs from this branch into a new protomux channel, e.g. named `comapeo/auth`, then we can roll out future changes without breaking changes to this invite-over-the-internet protocol, or without maintaining two separate protocols. 

**Fix:** Create a protomux in remote-discovery.js and store in `noiseStream.userData` like we already do for local, and open `comapeo/auth`. The first message should be a `hello` with `{ protocolVersion, features }` - `protocolVersion` should never change, but we should treat anything other than `1` as "disconnect - you need to update to connect to this peer". `features` should list the supported message types. An `IdentityProof { publicKey, signature }` message can be identical to the current waterfall message that we have. We can remove the manual implementation for reading length-prefixed messages since protomux covers that for us. 

**Defer invitor's `IndentityProof` until acceptance**
The invitor eagerly sends its identity proof on connection, without user intervention. This means that anyone that obtains a leaked invite URL can establish the identity of the invitor. It's a small risk, but significant in some scenarios.

**Fix:** only send the `IdentityProof` after the invitor calls `acceptInviteLinkRequest`. This means from the invitee side the invitor peer has no identity until after accept. I don't _think_ this matters, but we should check the code.

docs/guides/invite-over-internet.md
```diff
@@ -0,0 +1,147 @@
...
+// InvalidInternetInviteURLError
+
+// get curren list
+const [{
+  url,
+  inviteId,
+  roleId,
+  createdAt,
+  expiresAt,
+}] = await project.$member.listInviteLinks(url)
```
RangerMauve: No URL parameter!

docs/guides/invite-over-internet.md
```diff
@@ -0,0 +1,147 @@
...
+      break
+    case 'accepted':
+      // Invitor has accepted, initial sync in progress
+      // Show "joining project" screen
+      break
+    case 'completed':
+      // Join succeeded — update.projectId has the project public ID
+      break
+    case 'failed':
+      // Join failed — update.error has the reason
```
RangerMauve, line 95: List all error types that can happen here

docs/guides/invite-over-internet.md
```diff
@@ -0,0 +1,147 @@
...
+import {parseInviteURL} from '@comapeo/core/invite-urls.js'
+
+const {
+	invitorName,
+	projectName,
+	expiresAt,
+	inviteIdString,
+} = parseInviteURL(url)
+
+// Listen for join progress updates
```
RangerMauve, line 72: Add all these props

```JavaScript
/**
 * A join request initiated via an invite URL. Listen to `join-request-update`
 * events for progress.
 *
 * @typedef {object} JoinRequest
 * @property {string} inviteId Hex invite ID
 * @property {string} swarmPublicKey Invitor's swarm public key
 * @property {string} url Original invite URL
 * @property {JoinRequestStatus} status Current status
 * @property {Error|null} error Error if status is 'failed'
 * @property {string|undefined} projectId Resolved project ID on completion
 */
```

PR #1260, opened 2026-04-20
feat: Detect if peer supports map share RPC

RangerMauve:
This is needed so that frontend can avoid sending map share requests to peers that don't support it and causing the server to wait indefinitely for a timeout.

- New `supportedFeatures` flags in local peers info
- New flag in RPC to denote map share RPC support in DeviceInfo
- Fixes for cross version tests

### DIFF
--- a/proto/rpc.proto
+++ b/proto/rpc.proto
@@ -46,6 +46,7 @@ message DeviceInfo {
   enum RPCFeatures {
     features_unspecified = 0;
     ack = 1;
+    map_share = 2;
   }
   string name = 1;
   optional DeviceType deviceType = 2;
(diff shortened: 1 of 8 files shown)


gmaclennan (approved):
This looks good. Good, quick solution. My only is using an array for tracking supported features, this seems like it would be better as a map (object) of boolean props, maybe we can revisit this when/if we add another feature here. Using `Array.includes()` is always a pain to type, and it suggests ordering is important, but it isn't.

RangerMauve:
Sure, a `Record<string, boolean?>` could make it nicer. Plus we could skip the constant string needing to be exported.

`if(peer.supportedFeatures.mapShare) etc(...)`

gmaclennan (approved):
Looks good. For completeness maybe we should expose `ack` as a peerSupportedFeature at some point, but not necessary for this work.

PR #1262, opened 2026-05-05
chore: Bump comapeo-schema to 2.3.0

RangerMauve:
Adds the new `date` type to Field types.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
       "license": "MIT",
       "dependencies": {
         "@comapeo/fallback-smp": "^1.0.0",
-        "@comapeo/schema": "^2.3.0",
+        "@comapeo/schema": "2.3.0",
         "@digidem/types": "^2.3.0",
         "@fastify/error": "^3.4.1",
         "@fastify/type-provider-typebox": "^4.1.0",
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
       "license": "MIT",
       "dependencies": {
         "@comapeo/fallback-smp": "^1.0.0",
-        "@comapeo/schema": "2.2.0",
+        "@comapeo/schema": "^2.3.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -179,7 +179,7 @@
   },
   "dependencies": {
     "@comapeo/fallback-smp": "^1.0.0",
-    "@comapeo/schema": "2.2.0",
+    "@comapeo/schema": "^2.3.0",
```
achou11: Needs to be pinned to exact version

achou11 (changes requested):

package.json
RangerMauve: Shoot! Know how to force NPM to pin when I do `npm install`? Specifying the full version didn't work.

achou11: `npm i -E @comapeo/schema@latest` (can use `--exact` instead, can use actual version instead)

package-lock.json
```diff
@@ -10,7 +10,7 @@
       "license": "MIT",
       "dependencies": {
         "@comapeo/fallback-smp": "^1.0.0",
-        "@comapeo/schema": "2.2.0",
+        "@comapeo/schema": "^2.3.0",
```
achou11: if you run `npm i` this should be updated to be pinned to an exact version

achou11 (changes requested):
Need to update the lockfile too

achou11:
btw PR title should say schema, not core 😄

package-lock.json
RangerMauve: Weird my package lock isn't updating

achou11: maybe try clearing `node_modules` and then running `npm i`? otherwise, the surest-ish way is to

1. Clear `node_modules`
2. Restore the package-lock.json to its original state i.e. from main
3. Run `npm i`

achou11 (approved):
Assuming CI passes, lgtm!

PR #1267, opened 2026-05-13
fix: stop sync once a member is removed

### LINKED ISSUE #1268: Removing or blocking a project member does not stop sync
When a coordinator downgrades a peer's role by calling `$member.assignRole(deviceId, BLOCKED_ROLE_ID)` or `$member.remove(deviceId)` while the peer is still connected, the coordinator continues replicating user data to that peer until the peer's protomux session ends. New observations, blobs, and other data flow to a peer that has just been blocked, for the lifetime of the connection. See #1267 for a failing test for this issue.

### User story

Aster (coordinator) and Beta (member) are both online, syncing normally.

  1. Aster decides Beta should be blocked (e.g. they've been adding bad data) and calls `$member.remove(bob.deviceId)`.
  2. From Aster's app's POV the block "succeeded" — the member list shows Beta as blocked.
  3. Aster creates new observations. **Beta receives them.**
  4. The block only takes effect once Beta's protomux session ends (Beta disconnects, app backgrounded, network drop, etc.) and reconnects with a fresh PSC.

gmaclennan:
When a coordinator downgrades a peer's role by calling `$member.assignRole(deviceId, BLOCKED_ROLE_ID)` or `$member.remove(deviceId)` while the peer is still connected, the coordinator continues replicating user data to that peer until the peer's protomux session ends. New observations, blobs, and other data flow to a peer that has just been blocked, for the lifetime of the connection.

Root cause: `PSC#handleStateChange` only refreshes capability on an auth-namespace `not-synced → synced` transition. The writer's own `auth.want` stays at 0 throughout a self-written role doc, so that transition never fires.

**Changes**
- `PeerSyncController` subscribes directly to `roles.on('update', …)`, filtered by `this.peerId`. The event fires after the indexer has finished processing newly-indexed role docs, so the subsequent `getRole` read is guaranteed fresh.
- A new `#refreshSyncCapability` path preserves the **auth** capability only — config, blobIndex, data, and blob all cut off immediately on a downgrade. The auth-preservation works around `BLOCKED_ROLE.sync.auth = 'blocked'`: without it, a blocked peer's auth core would stop syncing on the writer's side and they'd never receive the role doc telling them they're blocked. Follow-up: revisit the BLOCKED role defaults so this preservation can be removed.
- `PeerSyncController.dispose()` removes its `coreManager`/`syncState`/`roles` listeners; called from `SyncApi#handlePeerDisconnect`. Also closes a pre-existing minor listener leak on `coreManager` per reconnect.
- Existing `#handleStateChange` auth-refresh path is retained — it still covers the implicit project-creator role inferred from `coreOwnership`, where no role doc exists in the indexer to fire an update event (matters during initial sync).

The first commit here is a regression test confirming the issue.

### DIFF
--- a/src/sync/peer-sync-controller.js
+++ b/src/sync/peer-sync-controller.js
@@ -1,7 +1,7 @@
 import mapObject from 'map-obj'
 import { NAMESPACES, PRESYNC_NAMESPACES } from '../constants.js'
 import { Logger } from '../logger.js'
-import { createMap } from '../utils.js'
+import { createMap, noop } from '../utils.js'
 import { unreplicate } from '../lib/hypercore-helpers.js'
 import { ExhaustivenessError } from '../errors.js'
 /** @import { CoreRecord } from '../core-manager/index.js' */
@@ -21,6 +21,7 @@ export class PeerSyncController {
   #coreManager
   #protomux
   #roles
+  #syncState
   /** @type {Record<Namespace, SyncCapability>} */
   #syncCapability = createNamespaceMap('unknown')
   /** @type {SyncEnabledState} */
@@ -52,14 +53,22 @@ export class PeerSyncController {
     this.#coreManager = coreManager
     this.#protomux = protomux
     this.#roles = roles
+    this.#syncState = syncState
 
     // Always need to replicate the project creator core
     this.#replicateCoreRecord(coreManager.creatorCoreRecord)
 
     coreManager.on('add-core', this.#handleAddCore)
     syncState.on('state', this.#handleStateChange)
+    roles.on('update', this.#handleRolesUpdate)
 
     this.#updateEnabledNamespaces()
+  }
+
+  dispose() {
+    this.#coreManager.off('add-core', this.#handleAddCore)
+    this.#syncState.off('state', this.#handleStateChange)
+    this.#roles.off('update', this.#handleRolesUpdate)
   }
 
   get peerKey() {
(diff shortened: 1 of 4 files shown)


gmaclennan:
@RangerMauve somehow I managed to accidentally merge this locally. Can you still take a look and we can revert or patch any issues.

src/sync/peer-sync-controller.js
```diff
@@ -52,16 +53,24 @@ export class PeerSyncController {
...
     this.#replicateCoreRecord(coreManager.creatorCoreRecord)
 
     coreManager.on('add-core', this.#handleAddCore)
     syncState.on('state', this.#handleStateChange)
+    roles.on('update', this.#handleRolesUpdate)
 
     this.#updateEnabledNamespaces()
   }
 
+  dispose() {
```
RangerMauve, line 68: Thoughts on renaming to `close`?

src/sync/peer-sync-controller.js
```diff
@@ -154,21 +163,64 @@ export class PeerSyncController {
...
+
+  /**
+   * @returns {Promise<void>}
+   */
+  async #readAndCacheSyncCapability() {
+    try {
+      this.#log('reading role for %h', this.peerId)
+      const cap = await this.#roles.getRole(this.peerId)
+      // Copy: ROLES.*.sync is a shared object reference.
+      this.#syncCapability = { ...cap.sync }
```
RangerMauve, line 216: Why clone the object now when it was passed by reference before?

src/sync/sync-api.js
```diff
@@ -568,13 +568,15 @@ export class SyncApi extends TypedEmitter {
...
-    if (!this.#peerSyncControllers.has(protomux)) {
+    const psc = this.#peerSyncControllers.get(protomux)
+    if (!psc) {
       this.#l.log(
         'Unexpected no existing peer sync controller for peer %h',
         protomux.stream.remotePublicKey
       )
       return
     }
+    psc.dispose()
```
RangerMauve, line 579: Should we also remove it from the map?

RangerMauve (commented):
LGTM overall, just some minor comments.

PR #1269, opened 2026-05-18
feat: Throw error when attempting to remove self

### LINKED ISSUE #1167: `MemberApi.remove()` should error when attempting to remove yourself?
Extracted from https://github.com/digidem/comapeo-core-react/pull/134#discussion_r2475614007. The current implementation assumes that attempts to remove yourself using this API won't happen and that `MemberApi.leaveProject()` will be used instead. If that's the case, `MemberApi.remove()` should enforce this behavior and error in this case.

Alternatively, could consider unifying the API under `MemberApi.remove()` and deprecating `MemberApi.leaveProject()`.

RangerMauve:
Closes #1167

### DIFF
--- a/src/member-api.js
+++ b/src/member-api.js
@@ -35,6 +35,7 @@ import {
   AlreadyInvitingError,
   InvalidResponseBodyError,
   RPCDisconnectBeforeAckError,
+  CannotBlockSelfError,
 } from './errors.js'
 import { wsCoreReplicator } from './lib/ws-core-replicator.js'
 import {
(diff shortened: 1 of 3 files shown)


src/member-api.js
```diff
@@ -425,6 +426,10 @@ export class MemberApi extends TypedEmitter {
       throw new AlreadyBlockedError()
     }
 
+    if (deviceId === this.#ownDeviceId) {
+      throw new CannotBlockSelfError()
+    }
```
achou11: Maybe do this as the first check to avoid potentially unnecessary work related to the check for blocked/left?

Guess it depends on what the least surprising error is in this very specific case. Initial thought is that misuse of the API (i.e. cannot remove self) should have priority.

achou11 (approved):
Non-blocking comment that may result in changes, but otherwise LGTM!

src/member-api.js
RangerMauve: Yeah good point. FWIW I don't think we'll hit either of the conditions any time soon. :P

PR #1276, opened 2026-06-06
fix: don't reject leaveProject when post-leave sync propagation stalls

gmaclennan:
leaveProject performs the local leave (assign LEFT + clear data) and then awaits a trailing waitForSync('initial', { timeoutMs: 45s }) to propagate the LEFT role to connected peers. That wait was awaited unconditionally, so when a peer is connected but presync can't complete within 45s — a slow/flaky link, or the user moving out of range right as they leave — leaveProject rejected with Error('Sync timeout') even though the device had already left locally. This is the intermittent "Sync timeout" error users have hit leaving projects.

Make the propagation best-effort: the local leave has already succeeded and the LEFT role (persisted to the auth namespace) also propagates on the next connection, so a stalled propagation must not fail the operation.

Adds a regression test (test-e2e/sync-leave-regression.js) plus a controllable transport test helper (test-e2e/controllable-wire.js) that drives two real projects over a pause-able wire to deterministically reproduce a connected-but- stalled peer.

### DIFF
--- a/src/mapeo-manager.js
+++ b/src/mapeo-manager.js
@@ -1050,10 +1050,18 @@ export class MapeoManager extends TypedEmitter {
 
     await project[kProjectLeave]()
 
-    // Sync any role changes from project leave
-    await project.$sync.waitForSync('initial', {
-      timeoutMs: INITIAL_SYNC_TIMEOUT_MS,
-    })
+    // Failure to sync the leave action (e.g. due to poor connectivity with a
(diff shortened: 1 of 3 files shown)


RangerMauve (approved):

RangerMauve:
tested locally, our cloud tests are just having trouble provisioning VMs. Also GH actions have been more janky

PR #1282, opened 2026-06-11
fix: remove DataStore add-core listener on close()

### LINKED ISSUE #1277: DataStore does not remove its 'add-core' listener on close(), throwing 'Cannot add core after closing' during teardown
## Summary

`DataStore` registers a permanent `add-core` listener on the `CoreManager` but never removes it on `close()`. If `coreManager` emits `add-core` after the `DataStore`'s `MultiCoreIndexer` has been closed (a teardown/shutdown racing with a sync session or peer connection that adds a new core), the listener calls `addCore()` on a closed indexer, which throws **`Cannot add core after closing`** (`multi-core-indexer` `#assertOpen`).

## Where

`src/datastore/index.js` constructor:

```js
coreManager.on('add-core', (coreRecord) => {
  if (coreRecord.namespace !== namespace) return
  this.#coreIndexer.addCore(coreRecord.core)   // throws if indexer already closed
})
```

`close()` only closes the indexer and leaves the listener attached:

```js
async close() {
  await this.#coreIndexer.close()
}
```

## Why it matters

In CoMapeo mobile this throw happens inside a synchronous `EventEmitter` callback in the embedded Node backend, so it is uncaught. The backend's `process.on('uncaughtException')` handler latches the server into a permanent `ERROR` state, which the frontend then re-throws on every subsequent RPC send — turning one benign teardown race into a storm of fatal/ErrorBoundary crashes.

Sentry: COMAPEO-20J / COMAPEO-20H / COMAPEO-213
- https://awana-digital.sentry.io/issues/7389717095/
- https://awana-digital.sentry.io/issues/7389717093/
- https://awana-digital.sentry.io/issues/7415554695/

## Proposed fix

Keep a reference to the bound listener and remove it in `close()`:

```js
// constructor
this.#handleAddCore = (coreRecord) => {
  if (coreRecord.namespace !== this.#namespace) return
  this.#coreIndexer.addCore(coreRecord.core)
}
coreManager.on('add-core', this.#handleAddCore)

// close()
async close() {
  this.#coreManager.off('add-core', this.#handleAddCore)
  await this.#coreIndexer.close()
}
```

Complementary hardening (separate, in `multi-core-indexer`): `addCore()` should no-op when the indexer is closing/closed rather than throw, so a late `add-core` during teardown can never throw. See digidem/multi-core-indexer.

A PR for this fix follows.

gmaclennan:
Fixes #1277.

`DataStore` registered a permanent `add-core` listener on the `CoreManager` but never removed it on `close()`. If `coreManager` emitted `add-core` after the `DataStore`'s `MultiCoreIndexer` had closed (a teardown racing with a sync session or peer connection adding a core), the listener called `addCore()` on a closed indexer, throwing **`Cannot add core after closing`** from inside a synchronous `EventEmitter` callback (uncaught).

In CoMapeo mobile this latched the embedded backend into a permanent `ERROR` state, which the frontend then re-threw on every subsequent RPC send — a storm of fatal/ErrorBoundary crashes (Sentry COMAPEO-20J / 20H / 213).

### Changes
- Store the bound `add-core` listener as a private field and `coreManager.off('add-core', …)` it in `close()` before closing the indexer.
- Regression test: `close()` removes the add-core listener (`listenerCount` returns to baseline).

### Related
- Complementary hardening: digidem/multi-core-indexer#49 makes `addCore()` a no-op when closed, so a late add-core during teardown can't throw even if a caller misses removing a listener.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/datastore/index.js
+++ b/src/datastore/index.js
@@ -50,6 +50,8 @@ export class DataStore extends TypedEmitter {
   #pendingAppends = new Set()
   /** @type {Record<MapeoDoc['schemaName'], Set<string>>} */
   #pendingEmits
+  /** @type {(coreRecord: import('../core-manager/index.js').CoreRecord) => void} */
+  #handleAddCore
 
   /**
    * @param {object} opts
@@ -75,10 +77,11 @@ export class DataStore extends TypedEmitter {
       batch: (entries) => this.#handleEntries(entries),
       reindex,
     })
-    coreManager.on('add-core', (coreRecord) => {
+    this.#handleAddCore = (coreRecord) => {
       if (coreRecord.namespace !== namespace) return
       this.#coreIndexer.addCore(coreRecord.core)
-    })
+    }
+    coreManager.on('add-core', this.#handleAddCore)
     this.#coreIndexer.on('idle', this.#handleIndexerIdle)
   }
 
@@ -214,6 +217,10 @@ export class DataStore extends TypedEmitter {
   }
 
   async close() {
+    // Remove the add-core listener before closing the indexer so a late
+    // 'add-core' event during teardown can't call addCore() on a closed indexer
+    // (which would throw 'Cannot add core after closing').
+    this.#coreManager.off('add-core', this.#handleAddCore)
     await this.#coreIndexer.close()
   }
 
--- a/test/datastore.js
+++ b/test/datastore.js
@@ -69,6 +69,29 @@ test('read and write', async (t) => {
   )
 })
 
+test('close() removes its add-core listener from the coreManager', async (t) => {
+  const cm = createCoreManager(t)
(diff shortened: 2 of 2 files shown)


RangerMauve (approved):
Looks good overall

gmaclennan:
@RangerMauve thoughts on why the windows workflow is timing out? It seems like this normally completes in about 9 mins, but I've re-run it 3 times and it keeps failing here. Not sure what in this PR would change things, but maybe it's a github infra issue and these machines are just running slow today? In which case should we increase our timeout?

RangerMauve:
I think it's just that one test that's flaky. Unsure if their infra is just slower right now and its surfacing a race condition somewhere. Defs not related to this PR though.

PR #1286, opened 2026-06-22
fix: fix display of error message for UnknownError

achou11:
The usage of `UnknownError` passes the error directly instead of as an object with an `err` field, causing the error's message to not be expanded. Example of the error not being displayed properly:

image


RangerMauve (approved):
Good catch!

RangerMauve:
Cloud tests are just kinda messed up rn. I think I need to update the version of core inside comapeo-cloud 😅

PR #1287, opened 2026-06-22
fix: fix usage of custom errors

achou11:
Similar in nature to #1286 , where there were a few custom errors that presumably don't expand template variables as expected due to incorrect usage.

### DIFF
--- a/src/blob-api.js
+++ b/src/blob-api.js
@@ -62,7 +62,7 @@ export class BlobApi {
     const writePromises = [
       pipeline(fs.createReadStream(original), hashTransform(hash), ws).catch(
         (/** @type {Error} */ e) => {
-          throw new BlobReadError(original, { cause: e })
+          throw new BlobReadError({ path: original }, { cause: e })
         }
       ),
     ]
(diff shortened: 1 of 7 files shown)


achou11:
Part of me thinks that any custom error that only has a template variable in its message (e.g. `message: '{message}'`) should be updated so that there's some generic error message (not containing a template variable), and callers provide a custom error message i.e. reverting some of the changes in this PR.

Thoughts?

RangerMauve (changes requested):
TY for the fix for the missing param wrappers.

I don't think we should replace `new CustomError(message)` with the `{\nmessage: \n}` pattern. It's more verbose and I'd rather keep the LoC lower until we actually need to update the parameters. I also don't think we'd get in a situation where we need both `message` and additional params.

achou11:
> I also don't think we'd get in a situation where we need both `message` and additional params.

The situation I was thinking of is when the message changes from `{message}` to something like `Some error for ID ${id}`. A caller like `new ExampleError('An error occurred')` will not be aware of the change to the message template, which could lead to issues similar to what #1286 fixed.

All that being said, I think reverting the relevant changes is fine. Will update the PR to do that 👍

PR #1289, opened 2026-06-24
chore: remove typedoc

RangerMauve:
- remove typedoc deps
- remove npm scripts for typedoc build
- delete unused `docs/api` folder
- remove docs workflow

## Why?

- nobody uses them (that we are aware of)
- typedoc version mismatches are causing CI errors


achou11 (approved):

gmaclennan (approved):
yay bye-bye docs 👋 who needs docs anyway...
I feel like I've always been chasing the dream of auto-generated docs from JSDoc / TSDoc comments, but it's never quite worked or been useful.
I do think long-term we need a docs site for core, but not an immediate priority, and this was not it.

RangerMauve:
maybe we can slop gen some docs that are more useful by parsing out our tests into examples and how-tos

PR #1290, opened 2026-06-27
ci: notify downstream repos on release

gmaclennan:
Adds a `notify-downstream` job that fires a `repository_dispatch` (event type `first-party-release`) to downstream repos when `@comapeo/core` publishes, so they fast-track the dependency bump instead of waiting for their daily Dependabot run.

Pairs with the receiver workflow in digidem/comapeo-core-react-native#173. The downstream destinations are a matrix (`owner`/`repo`), so adding more later is a one-line change.

Requires: `RELEASE_BOT_APP_ID` + `RELEASE_BOT_PRIVATE_KEY` available to this repo, and the Releases bot installed on each downstream repo with Contents: write (needed to send the dispatch).


.github/workflows/release.yml
```diff
@@ -61,3 +61,34 @@ jobs:
...
+    needs: release
+    if: >-
+      needs.release.result == 'success' &&
+      github.event_name == 'pull_request' &&
+      github.event.pull_request.merged == true
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        downstream:
```
RangerMauve, line 78: Lets add comapeo-cloud to this. I think Desktop and Mobile can pull it since they likely need more changes to update and pair it with feature releases
```suggestion
        downstream:
          - owner: digidem
            repo: comapeo-cloud
```

RangerMauve (approved):

.github/workflows/release.yml
gmaclennan: Good idea. I'm in the process of turning this into a shared action so there's only one place where we define the notification targets for all our deps

PR #1297, opened 2026-07-06
Feat/geojson export flattening

### LINKED ISSUE #1283: GeoJSON flatten exported properties
## Description

https://app.notion.com/p/digidem/GeoJson-export-improvements-3731b08162d5800dad8ef5d17cc34b3b

## Tasks

- [x] Refactor GeoJSON exports into own class
- [x] Remove schema specific properties and prefix remaining ones
- [x] raw doc in a new field in the GeoJSON object
- [x] Flatten tags into their own fields in properties
- [x] Flatten attachment array into individual fields, have full names of file attachments for easier referencing
- [ ] Add category name deref'd from presetRef
- [ ] Add creator name deref'd from device id

RangerMauve:
Closes #1283

### DIFF
--- a/src/data-exporter.js
+++ b/src/data-exporter.js
@@ -12,7 +12,10 @@ import { Logger } from './logger.js'
 import ensureError from 'ensure-error'
 
 /** @import { Observation, Track } from '@comapeo/schema' */
-/** @import { Attachment, BlobId, ObservationDataType, TrackDataType } from './types.js' */
+/** @import { Attachment, BlobId, DeviceInfoDataType, ObservationDataType, PresetDataType, TrackDataType } from './types.js' */
+/** @import {Feature, Point, LineString} from 'geojson'*/
+
+/** @import {DerivedDocFields} from './datatype/index.js' */
 
 /** @typedef {Map<string, Attachment>} SeenAttachments */
 
@@ -34,6 +37,8 @@ import ensureError from 'ensure-error'
  * @typedef {object} DataExporterDependencies
  * @prop {ObservationDataType} observations DataType for observations
  * @prop {TrackDataType} tracks DataType for tracks
+ * @prop {PresetDataType} presets DataType for presets (categories)
+ * @prop {DeviceInfoDataType} deviceInfo DataType for device info (author names)
  * @prop {() => Promise<string | undefined>} getProjectName Async function to get the project name
  * @prop {import('./blob-store/index.js').BlobStore} blobStore BlobStore instance for attachment access
  * @prop {Logger} logger Logger instance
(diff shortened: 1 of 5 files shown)


test-e2e/project-export.js
```diff
@@ -214,6 +214,17 @@ test('Project export tracks and observations to zip stream', async (t) => {
       'Exported GeoJSON has expected number of features'
     )
 
+    /** @type {{ $comapeo: any, properties: Record<string, any> }[]} */
+    const features = parsed.features
+    // TODO: remove $comapeo filter once tracks go through makeObservationFeature
+    const observationFeatures = features.filter((f) => f.$comapeo !== undefined)
```
gmaclennan: Not sure what this is doing? I think both tracks and observations have this property now? Note the filter below on tracks passes even though tracks are included because attachment_1 is undefined for tracks. Test as it stands doesn't really test that much

src/data-exporter.js
```diff
@@ -0,0 +1,561 @@
...
+    const dd = `${date.getDate()}`.padStart(2, '0')
+    return `CoMapeo_${name}_${type}_${yyyy}_${mm}_${dd}`
+  }
+
+  /**
+   * @param {boolean} observations
+   * @param {boolean} tracks
+   * @returns {Promise<string>}
+   */
+  async #geoJSONFileName(observations, tracks) {
```
gmaclennan: Nit: could be combined with zipfilename below - identical logic apart from the extension.

src/data-exporter.js
```diff
@@ -0,0 +1,561 @@
...
+ * @typedef {object} DataExporterDependencies
+ * @prop {ObservationDataType} observations DataType for observations
+ * @prop {TrackDataType} tracks DataType for tracks
+ * @prop {() => Promise<string | undefined>} getProjectName Async function to get the project name
+ * @prop {import('./blob-store/index.js').BlobStore} blobStore BlobStore instance for attachment access
+ * @prop {Logger} logger Logger instance
+ */
+
+/**
+ * @typedef {object} FeatureKnownFields
```
gmaclennan, line 48: Include $updatedAt and $observations?

gmaclennan (approved):
Overall looks great and good to merge, just some minor points in the comments. 

One small thing to do at some point is handle collisions: eg if a tag is named `$id`, or any other name that clashes with our `$`-prefixed properties. We should probably prohibit `$`-prefixing for tags in comapeocat, but we should still handle it gracefully. Also: should the attachment properties be `$`-prefixed do you think? Happy to go with your decision on that.

test-e2e/project-export.js
RangerMauve: Leftover from when I just had observations with the field

PR #1310, opened 2026-08-16
feat: accept a pre-derived master key in MapeoManager

gmaclennan:
`MapeoManager` uses its `rootKey` option only to construct `new KeyManager(rootKey)`, which derives a 32-byte master key via Argon2id (64 MiB memlimit) on every boot. This adds an optional `masterKey` option passed straight through to `KeyManager` (digidem/comapeo-crypto#37), letting callers that cache the derived key skip the derivation. One line of behaviour change plus a JSDoc param.

Since crypto renamed itself upstream, this also moves the dependency from the exact-pinned `@mapeo/crypto@1.0.0-alpha.10` to `@comapeo/crypto@2.0.0` (exact pin kept — this is the one dep where derivation drift means identity loss) and renames the import across all 19 source/test files.

Tests: the same rootKey yields the same deviceId with and without `masterKey` supplied (pinning identity across the alpha.10 → 2.0.0 jump), and a deliberately wrong 32-byte master key yields a different deviceId — verified by mutation that dropping the pass-through makes this second assertion fail, so the option is provably load-bearing.

Merge sequencing: blocked on a `@comapeo/crypto@2.0.0` release — the lockfile still pins the old package (2.0.0 is unpublished, so it cannot be regenerated yet) and needs one `npm install` after the release. Part of the master-key caching plan in [comapeo-core-react-native docs/master-key-cache-plan.md](https://github.com/digidem/comapeo-core-react-native/blob/claude/key-manager-master-key-cache-9k40es/docs/master-key-cache-plan.md) (§2.4).


RangerMauve (approved):
What do you think of having our createManager(s) test utils pick from a hardcoded set of master keys for e2e tests? I think we could get huge savings on runtime if the key derivation is turning out to be so expensive.

gmaclennan:
> What do you think of having our createManager(s) test utils pick from a hardcoded set of master keys for e2e tests? I think we could get huge savings on runtime if the key derivation is turning out to be so expensive.

yes that would be fine I think. We could, if we are bad developers, just pass any random key in as the master key, and a non-matching rootKey. The only thing that uses the root key is the identity key, so it wouldn't fail anything, but that would make us bad devs...

PR #1322, opened 2026-08-27
Feat/left project error data

### LINKED ISSUE #1313: feat: typed ProjectLeftError from methods prohibited on a left project
`leaveProject()` deliberately keeps the project instance alive — the auth namespace stays open so the LEFT role record can sync to peers — but guts everything else: `kClearData` closes and unlinks the config and data stores, and closes the writer cores and deletes others' data for the config, data, blob and blobIndex namespaces.

The problem is what a caller sees afterwards. Writes fail with an untyped hypercore session-closed error from the closed writer core. Reads fail with a bare `Error: Cannot await idle after closing` thrown by multi-core-indexer. Neither error says "you left this project", and neither is distinguishable from a transient close. Any consumer that wants to render a coherent left-project state — a frontend, or an RPC layer forwarding these errors (see digidem/comapeo-ipc#88) — has to sniff error strings.

Proposal: the instance should know it is left (the manager knows `hasLeftProject` when it constructs the instance, and `kClearData` is the choke point for both the leave path and the crash-recovery re-gut in `getProject`). Methods prohibited on a left project throw a typed `ProjectLeftError` with a stable code, and the state is directly readable (`project.isLeft` or similar). On the code itself: our convention in `src/errors.js` is an `_ERROR` suffix, so `PROJECT_LEFT_ERROR` matches — but the IPC layer currently ships `PROJECT_LEFT`, and since this code becomes a cross-process wire contract it should be picked once, deliberately, and kept stable.

Surfaces to convert: the data-type CRUD methods, `$setProjectSettings`, `$blobs`, `$icons`, and the export methods. Unchanged: `$getOwnRole`, `$sync`, and the `$getProjectSettings` fallback, which all work today.

One surface needs an actual decision rather than a mechanical conversion: `$member.getMany()` throws on a left project because of its top-level `this.#dataTypes.deviceInfo.getMany()` call (deviceInfo is config-namespace). The per-member join is already left-tolerant, so the fix is small — and worth making, because role changes invalidate members queries downstream, meaning the leave flow itself hits this path. `getById(ownDeviceId)` also throws (`MissingOwnDeviceInfoError`). Making member reads degrade to deviceId-only info would let a left project's membership still render.

RangerMauve:
closes https://github.com/digidem/comapeo-core/issues/1313

We should have reasonable errors after we leave a project.
Also we should be able to do a basic member list even if their names are cleared.

### DIFF
--- a/src/mapeo-manager.js
+++ b/src/mapeo-manager.js
@@ -13,7 +13,6 @@ import { IndexWriter } from './index-writer/index.js'
 import {
   MapeoProject,
   kBlobStore,
-  kClearData,
   kProjectLeave,
   kSetIsArchiveDevice,
   kSetOwnDeviceInfo,
@@ -588,7 +587,7 @@ export class MapeoManager extends TypedEmitter {
     // trying to create a project instance if we have marked the project as
     // "left".
     if (projectKeysTableResult.hasLeftProject) {
-      await project[kClearData]()
+      await project[kProjectLeave]()
     }
 
     // 3. Keep track of project instance as we know it's a properly existing project
--- a/src/mapeo-project.js
+++ b/src/mapeo-project.js
@@ -969,6 +969,7 @@ export class MapeoProject extends ReadyResource {
     })
   }
 
+  /** Leave the project and clear the data. Does not re-assign role when already left */
   async [kProjectLeave]() {
     const { roleId } = await this.$getOwnRole()
 
--- a/src/mapeo-project.js
+++ b/src/mapeo-project.js
@@ -121,19 +121,22 @@ const EMPTY_PROJECT_SETTINGS = Object.freeze({ sendStats: false })
  */
 
 /**
- * @param {any} target
- * @returns {any}
+ * @template T
+ * @param {T} target
+ * @returns {T}
  */
 function createProjectLeftMock(target) {
-  return new Proxy(target, {
-    get(obj, prop) {
-      const value = Reflect.get(obj, prop)
-      if (typeof value === 'function') {
-        return () => Promise.reject(new ProjectLeftError())
-      }
-      return value
-    },
-  })
+  return /** @type {T} */ (
+    new Proxy(/** @type {object} */ (target), {
+      get(obj, prop) {
+        const value = Reflect.get(obj, prop)
+        if (typeof value === 'function') {
+          return () => Promise.reject(new ProjectLeftError())
+        }
+        return value
+      },
+    })
+  )
 }
 
 export class MapeoProject extends ReadyResource {
--- a/src/mapeo-project.js
+++ b/src/mapeo-project.js
@@ -970,9 +970,9 @@ export class MapeoProject extends ReadyResource {
   }
 
   async [kProjectLeave]() {
-    const ownRole = await this.$getOwnRole()
+    const { roleId } = await this.$getOwnRole()
 
-    if (ownRole.roleId !== BLOCKED_ROLE_ID) {
+    if (roleId !== BLOCKED_ROLE_ID && roleId !== LEFT_ROLE_ID) {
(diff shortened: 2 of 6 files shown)


src/mapeo-project.js
```diff
@@ -664,19 +687,26 @@ export class MapeoProject extends ReadyResource {
   }
 
   get observation() {
+    if (this.#hasLeft) return createProjectLeftMock(this.#dataTypes.observation)
```
gmaclennan, line 691: minor, but this creates a new instance every time, so `project.observation !== project.observation` and I guess if we hammered the API there would be a lot of proxy objects to garbage collect, but they are low overhead so probably not a huge issue.

src/mapeo-project.js
```diff
@@ -141,6 +162,8 @@ export class MapeoProject extends ReadyResource {
   #l
   /** @type {Boolean} this avoids loading multiple configs in parallel */
   #importingCategories
+  /** @type {Boolean} */
+  #hasLeft = false
```
gmaclennan, line 167: After restart this remains false. Not an issue right now since I don't think there is code where we call `manager.getProject()` on a left project, but it would be good if the manager could call this when it reads `hasLeftProject` in the `getProject()` code.

src/member-api.js
```diff
@@ -634,7 +634,8 @@ export class MemberApi extends TypedEmitter {
   async getMany({ includeLeft = false } = {}) {
     const [allRoles, allDeviceInfo] = await Promise.all([
       this.#roles.getAll(),
-      this.#dataTypes.deviceInfo.getMany(),
+      // If we can't get info due to leave, still allow listing
+      this.#dataTypes.deviceInfo.getMany().catch(() => []),
```
gmaclennan: This swallows legitimate errors from `getMany()`. I think maybe it's better to just let this throw if we've left?

gmaclennan (changes requested):
A few paths missing the left error I think:

- `$member.getById(ownDeviceId)`
- `$setProjectSettings()`
- geoJSON and zip exports

And also `project.$blobs` - it will still act as if it's working I think.

Minor, but the translations API has a public method `project.$translations.index()` which is synchronous, and the error proxy makes it throw async. I don't think there is a use for this method on the public API, but it's a future "gotcha". We could maybe move that to being a private method, but that's beyond the scope of this PR. Maybe ok to leave and address later since it doesn'at affect anything yet?

src/mapeo-project.js
RangerMauve: I think unless we hit a hot path with this case it should be safe.

src/mapeo-project.js
RangerMauve: we clear left projects on load so I can make sure that we track this flag for that case.

https://github.com/digidem/comapeo-core/blob/c7e2ce18084f3d8a363277a69f3cd4573c5cf186/src/mapeo-manager.js#L590

src/mapeo-manager.js
```diff
@@ -588,7 +587,7 @@ export class MapeoManager extends TypedEmitter {
     // trying to create a project instance if we have marked the project as
     // "left".
     if (projectKeysTableResult.hasLeftProject) {
-      await project[kClearData]()
+      await project[kProjectLeave]()
```
gmaclennan, line 590: This writes a new role record (left role) every time we call `getProject()` - shouldn't do that.

RangerMauve: We get own role already so can add a check there

src/mapeo-project.js
```diff
@@ -970,9 +970,9 @@ export class MapeoProject extends ReadyResource {
   }
 
   async [kProjectLeave]() {
```
gmaclennan, line 973: possibly worth a short comment that this is not idempotent (is that the right word?) now.

gmaclennan (approved):
small comment, good to merge.

PR #1326, opened 2026-09-01
test: check that concurrent getProject resolves to the same project

### LINKED ISSUE #1314: fix: de-duplicate concurrent getProject() calls for the same project
`MapeoManager.getProject()` (`src/mapeo-manager.js`) checks `#activeProjects`, and if there is no hit it constructs an instance and caches it:

```js
const activeProject = this.#activeProjects.get(projectPublicId)
if (activeProject) return activeProject
// ...synchronous db lookup and key decode...
const project = await this.#createProjectInstance(projectKeys)
if (projectKeysTableResult.hasLeftProject) {
  await project[kClearData]()
}
this.#activeProjects.set(projectPublicId, project)
```

Two calls in the same tick for a not-yet-cached project both pass the check and both construct a `MapeoProject` over the same SQLite file and the same corestore directory. The second `set()` wins; the first instance is orphaned — never closed, sqlite handle open, cores open, localPeers listeners attached, still replicating.

The orphan's `once('close')` cleanup closes over `projectPublicId`, not the instance, so if the orphan is ever closed it evicts the *winning* instance from the cache.

The race could happen from `setDeviceInfo` (opens every project), the blob and icon fastify routes which open on demand, `addProject` and `leaveProject` both call `getProject`, and an RPC server handling more than one connection (which happens transiently during an app reload) issues concurrent `getProject` calls per connection.

Fix: a pending-promise map keyed by `projectPublicId`, entries cleared on settle. Related: #407 covers the opposite race (`getProject` returning an already-closing instance).

RangerMauve:
Closes #1314

### DIFF
--- a/src/mapeo-manager.js
+++ b/src/mapeo-manager.js
@@ -145,6 +145,11 @@ export class MapeoManager extends TypedEmitter {
   // Maps project public id -> project instance
   /** @type {Map<string, MapeoProject>} */
   #activeProjects
+  // Maps project public id -> in-flight getProject() promise. Prevents
+  // concurrent calls for the same not-yet-cached project from constructing
+  // multiple MapeoProject instances over the same storage.
+  /** @type {Map<string, Promise<MapeoProject>>} */
+  #pendingProjects
   /** @type {CoreStorage} */
   #coreStorage
   #dbFolder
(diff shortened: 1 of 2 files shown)


RangerMauve (commented):
Gonna self review. It;s a simple change. Will merge and release so I can test in my upcoming IPC PR

PR #1328, opened 2026-09-02
chore: adds Content-Range and Content-Length

ErikSin:
towards [2109](https://github.com/digidem/comapeo-mobile/issues/2109) in comapeo mobile

iOS uses AVPlayer (via Expo-Audio) to play audio within Comapeo-Mobile. The Audio returned by the fastify server was not playing at all.

The blob download route (GET /projects/:projectId/blob/...) never sent Content-Length, and didn't support Range requests. Without a known content length iOS's AVPlayer can't determine a video's duration and treats it as an indefinite live stream, which stalls forever instead of playing.

Adds a parseRangeHeader helper that parses a single Range: bytes=start-end header (including open-ended and suffix forms, e.g. bytes=500-, bytes=-500), the only form clients like AVPlayer send.

When a valid Range header is present, the route now reads only the requested byte span from the blob store, responds 206 Partial Content with Content-Range and Content-Length, and always advertises Accept-Ranges: bytes. Full responses now also get an explicit Content-Length.

Threads start/length through to blobStore.createReadStreamFromEntry, which already forwarded its options straight to Hyperblobs' createReadStream — just needed the JSDoc updated to declare the two new option fields.
A missing/malformed/unsatisfiable Range header falls back to serving the full resource, matching standard HTTP semantics.

### DIFF
--- a/src/blob-store/index.js
+++ b/src/blob-store/index.js
@@ -282,6 +282,8 @@ export class BlobStore extends ReadyResource {
    * @param {import('hyperdrive').HyperdriveEntry} entry Hyperdrive entry
    * @param {object} [options]
    * @param {boolean} [options.wait=false] Set to `true` to wait for a blob to download, otherwise will throw if blob is not available locally
+   * @param {number} [options.start] Byte offset to start reading from
+   * @param {number} [options.length] Number of bytes to read
    * @returns {Promise<Readable>}
    */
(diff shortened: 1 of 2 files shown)


RangerMauve (changes requested):
Lets use https://www.npmjs.com/package/range-parser instead so we don't need to maintiain this code or worry about edge cases

PR #1329, opened 2026-09-08
fix: pin relevant hypercore deps to exact versions

achou11:
Primarily addresses an issue with runtime errors caused by misaligned transitive dependencies. In my case, our migration module created a `hypercore-storage` instance using an older version (v2) that was then being used by a corestore version that needed an instance created from a newer version (v3).

Not a revelation, but I believe hypercore's looseness with defining deps versions contributes to the complexities here. Many of their modules are pluggable and my guess is that some of them should be declaring peer deps instead of just direct deps. Can't fully wrap my head around all of it...

This PR is just pinning the versions of relevant modules to exact versions, as done on the main branch. The versions are based on the existing versions in the lockfile, which is known to work and be tested. I packed this locally and it fixed the issues I was running into in my integration branch on desktop.


RangerMauve (approved):
LGTM shall we do another prerelease?

achou11:
> LGTM shall we do another prerelease?

yes please!

